PR #58, opened 2026-04-16
Offload llama.cpp to GPU (CUDA) + polish AI provider UX

KrasimirKralev:
### GPU offload for llama.cpp (the big one)
- `scripts/start-llamacpp.sh` now passes `--n-gpu-layers` (default `99`, override via `LLAMACPP_N_GPU_LAYERS`). On devices with a CUDA-built `llama-server` (e.g. Jetson Orin Nano — just rebuild with `-DGGML_CUDA=ON -DCMAKE_CUDA_ARCHITECTURES=87`), **all 36 layers offload to the iGPU**, taking Gemma 4 E2B from ~15 tok/s (CPU-pinned at 100%) to **~60 tok/s**. On CPU-only binaries the flag is a harmless no-op.
- Follow-up PR should make `install.sh` auto-build llama.cpp with CUDA when `nvcc` is present — until then, users rebuild manually, benefit is automatic.

### Polish
- **Provider icons in the AI provider list**: `AIModelsStep` radio list now shows the actual SVG `AIProviderIcon` next to each provider. Previously the list was text-only.
- **Wait copy fix**: `ai.pleaseDontClose` "takes about 30 seconds" → "takes a couple of minutes" across all 10 locales. Local AI provisioning with a 3GB model download + first-run warm-up routinely takes more than 30s; the old copy made the wait feel broken.

## Test plan

- [ ] On a Jetson with CUDA-built llama-server: check `/home/clawbox/clawbox/data/llamacpp/server.log` — should show `offloaded 36/36 layers to GPU` and tok/s ~55-65.
- [ ] On a CPU-only device: llama-server starts normally, no change in behavior.
- [ ] Setup wizard step 4 / Settings → AI Provider: each provider row shows its brand icon.
- [ ] Settings → Local AI → Enable Gemma 4 → loading screen says "takes a couple of minutes" (not 30 seconds).
- [ ] Switch UI locale to bg/de/ja/zh — localized versions of the new wait copy render.


KrasimirKralev:
Closing — opened without explicit approval. Will reopen with the full set once reviewed locally.

PR #68, opened 2026-04-18
fix(mdns): harden avahi so <hostname>.local resolves reliably on Windows

KrasimirKralev:
## Summary
fix(mdns): harden avahi so `<hostname>.local` resolves reliably on Windows

Windows 10/11 clients frequently fail to resolve `<hostname>.local` for
up to 15 minutes after a ClawBox joins their network, or after the
laptop itself reconnects / wakes from sleep. When the first mDNS query
fails (multicast dropped, avahi throttled, interface not yet settled),
Windows caches the NXDOMAIN in its dnscache service and browser-side
DNS cache, and every retry hits the cache instead of the network.

That window is miserable UX — users hit F5, close the tab, or assume
the device is broken — and it's ClawBox-fixable, not a pure-Windows
problem. We can make avahi answer fast enough, often enough, and
across enough protocol families that Windows' first query succeeds
and nothing ever lands in the negative cache.

Three changes:

1. config/avahi-daemon.conf — new hardened config installed over the
   distro default. Key non-default settings (rationale is inline in
   the file):
     - ratelimit-interval-usec=500000 / ratelimit-burst=5000. Windows
       issues bursts of parallel A/AAAA/PTR/SRV queries across many
       service types at the same instant; the upstream defaults of
       1 s / 1000 drop replies under that load.
     - publish-a-on-ipv6=yes / publish-aaaa-on-ipv4=yes. Without the
       cross-family publish, a dual-stack Windows box that asks AAAA
       first can get an empty authoritative reply and short-circuit
       the A lookup, turning a successful resolution into a failure.
     - use-iff-running=no. Announce the instant an address is
       assigned to an interface, not once IFF_RUNNING flips on. This
       matters on Wi-Fi where that delay is multi-second.
     - check-response-ttl=no. Some consumer routers rewrite mDNS
       packet TTL in transit; the default filter drops legitimate
       responses. Turning it off trades a mild anti-loop safety net
       for actually answering queries on those routers.

2. config/99-clawbox-avahi-reload — new NetworkManager dispatcher
   hook, installed to /etc/NetworkManager/dispatcher.d/ with mode
   0755. Fires on up / down / dhcp4-change / dhcp6-change /
   connectivity-change for every non-loopback interface and sends
   `avahi-daemon --reload`. The reload triggers a fresh announcement
   burst on all active interfaces, which clears stale negative caches
   on Windows within seconds rather than the default 15 minutes.
   Explicitly ignores lo/veth/docker/br-* events so the hotspot's
   own transitions don't trigger reload loops during setup.

3. install.sh apply_hostname — previously only rewrote the `host-name`
   line of the distro default avahi-daemon.conf. Now installs
   config/avahi-daemon.conf over that default (backing up the
   original once to avahi-daemon.conf.bak.orig for operator diff),

## Test plan

## Summary by CodeRabbit

* **New Features**
  * Added automatic service discovery daemon integration with network manager for seamless hostname and service publishing.
  * Enhanced installation process to configure hardened service discovery settings during system setup.

### DIFF
--- a/config/99-clawbox-avahi-reload
+++ b/config/99-clawbox-avahi-reload
@@ -1,0 +1,35 @@
+#!/bin/sh
+# NetworkManager dispatcher hook — force avahi to re-announce on every
+# interface state change so clients with stale negative mDNS caches
+# (especially Windows) can re-resolve `<hostname>.local` within seconds
+# of the device coming back on the network, not the 15 minutes Windows
+# would otherwise hold a cached NXDOMAIN.
+#
+# Dispatcher arg 1 is the interface name, arg 2 is the action.
+# We care about: up, down, dhcp4-change, dhcp6-change, connectivity-change.
+#
+# Installed by install.sh step_hostname_mdns at
+# /etc/NetworkManager/dispatcher.d/99-clawbox-avahi-reload
+# and marked executable (NetworkManager ignores non-executable hooks).
+
+IFACE="$1"
+ACTION="$2"
+
+case "$ACTION" in
+  up|down|dhcp4-change|dhcp6-change|connectivity-change)
+    # Ignore the hotspot's own transitions so we don't reload avahi
+    # mid-handshake when start-ap.sh is bringing its AP up.
+    case "$IFACE" in
+      lo|""|veth*|docker*|br-*)
+        exit 0
+        ;;
+    esac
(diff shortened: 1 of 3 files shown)


KrasimirKralev:
@coderabbitai review

PR #73, opened 2026-04-18
fix(chat): model-dropdown switches actually stick

KrasimirKralev:
## Summary
fix(chat): green confirmation pill on model switch, keep red for errors

The chat panel's `ChatMessage` type uses the same `role: 'system'`
for both error banners (WebSocket dropped, model switch failed) and
the "Switched chat to X." confirmation we emit after a successful
dropdown pick. The shared style renders both in a red pill with red
text, so a successful model switch looks like a failure in the
stream. In the screenshot the user showed, "Switched chat to
llamacpp/gemma4-e2b-it-q4_0." appeared directly above a real
`Error:` banner in identical styling — no visual distinction.

Changes in src/components/ChatPopup.tsx:

- Add an optional `variant: 'success' | 'error'` to the `ChatMessage`
  interface. Defaults to `'error'` (no behaviour change for existing
  call sites: all the error paths keep their red styling).
- Tag the `Switched chat to X.` message with `variant: 'success'`.
- Update the inline message styles in the renderer to branch on
  `msg.variant` for `system`-role bubbles: emerald (`#22c55e` fg,
  `rgba(34,197,94,0.15)` bg) for success, keep the existing red
  (`#ef4444` fg, `rgba(239,68,68,0.15)` bg) for errors.

Other system-role call sites — WS error banners, connection
failures, "provider not configured" hints — all keep the default
`'error'` variant and render red, so nothing else regresses.

fix(chat): model-dropdown switches actually stick to the active session

The ClawBox chat-panel model dropdown (`ChatPopup.tsx`) POSTs to
`/setup-api/chat/model` when the user picks a model. That route
previously did only two things:

  await exec(openclaw, ["config", "set", "agents.defaults.model.primary", target]);
  await restartGateway();

Both writes are correct, and new sessions opened after the restart
inherit the chosen model. But they left the *currently open chat
session's per-session override* untouched. That override lives in
`<agents-dir>/<agent>/sessions/sessions.json` as:

  "providerOverride":       "deepseek",
  "modelOverride":          "deepseek-chat",
  "modelOverrideSource":    "auto",
  "authProfileOverride":    "deepseek:default",
  ...

and it's what OpenClaw actually consults at chat time to decide which
provider to dial. Because the source was `"auto"`, OpenClaw's own
auto-picker happily rewrote it to whatever it preferred on every new
message — usually the local provider when reachable — so the chat

## Test plan

### DIFF
--- a/src/app/setup-api/chat/model/route.ts
+++ b/src/app/setup-api/chat/model/route.ts
@@ -1,6 +1,14 @@
 import { NextResponse } from "next/server";
 import { getAll } from "@/lib/config-store";
-import { inferConfiguredLocalModel, readConfig, restartGateway, runOpenclawConfigSet, type OpenClawConfig } from "@/lib/openclaw-config";
+import {
+  inferConfiguredLocalModel,
+  readConfig,
+  restartGateway,
+  runOpenclawConfigSet,
+  applyModelOverrideToAllAgentSessions,
+  parseFullyQualifiedModel,
+  type OpenClawConfig,
+} from "@/lib/openclaw-config";
 import { sqliteGet, sqliteSet } from "@/lib/sqlite-store";
 
 export const dynamic = "force-dynamic";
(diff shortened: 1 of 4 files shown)


KrasimirKralev:
@coderabbitai review

KrasimirKralev:
@coderabbitai review

KrasimirKralev:
@coderabbitai review

PR #74, opened 2026-04-18
fix(local-only): sweep per-session provider overrides on toggle

KrasimirKralev:
## Summary
fix(local-only): sweep per-session provider overrides on toggle

Local-only mode previously only flipped `agents.defaults.model.primary`
and cleared `agents.defaults.model.fallbacks` on the gateway. That
affected the *defaults for new sessions* but left every existing
session's per-session overrides untouched:

  "agent:main:main": {
    "providerOverride": "deepseek",
    "modelOverride": "deepseek-chat",
    "modelOverrideSource": "auto",
    "authProfileOverride": "deepseek:default",
    ...
  }

OpenClaw reads those fields from `<agents-dir>/<agent>/sessions/
sessions.json` on every chat request and routes the turn to the
pinned provider regardless of what the agent-level defaults say.
Result: the user toggles Local-only on, the toggle reports success,
the UI model pill says `llamacpp/gemma4-e2b-it-q4_0`, and every
chat still silently goes out to the cloud provider that was pinned
at session birth. Reproduced live on a Jetson with the local model
loaded and healthy — zero CPU on llama-server during the chat,
steady outbound TLS to openclawhardware.dev from the gateway, no
fallback-decision entries in the gateway log because no fallback
was actually happening: the primary was being resolved from the
session override, not the default.

Changes in src/app/setup-api/local-ai/exclusive/route.ts:

- Enumerate every `sessions.json` under /home/clawbox/.openclaw/agents
  (configurable via OPENCLAW_AGENTS_DIR for tests) and, for every
  session entry in each file, replace `providerOverride`,
  `modelOverride`, `modelOverrideSource`, `authProfileOverride`,
  `authProfileOverrideSource`, `modelProvider`, and `model` with the
  local target (e.g. provider=llamacpp, modelId=gemma4-e2b-it-q4_0,
  sources=manual so OpenClaw doesn't re-auto-select on next restart).
- Before overwriting, snapshot each field's prior value per-session
  (distinguishing "explicitly null" from "field absent") and save the
  entire per-file backup in ClawBox's config store under a new
  `local_only_saved_session_overrides` key. When the user toggles
  Local-only OFF, the backup is read back and fields are restored to
  their original values — or deleted entirely if they weren't set
  before — so the round-trip is lossless for sessions that existed
  at toggle-on time.
- Writes use a temp-file + rename atomic replace so a crash midway
  can't leave `sessions.json` corrupt (which would make OpenClaw
  refuse to resume those sessions).
- Sessions that appear *after* toggle-on are left alone on toggle-off:
  no backup exists for them, so the fresh override from when they

## Test plan

### DIFF
--- a/src/app/setup-api/local-ai/exclusive/route.ts
+++ b/src/app/setup-api/local-ai/exclusive/route.ts
@@ -1,12 +1,37 @@
 export const dynamic = "force-dynamic";
 
+import fs from "fs/promises";
+import path from "path";
 import { NextResponse } from "next/server";
 import { get, set, setMany } from "@/lib/config-store";
 import { readConfig, restartGateway, runOpenclawConfigSet } from "@/lib/openclaw-config";
 
 const SAVED_PRIMARY_KEY = "local_only_saved_primary";
 const SAVED_FALLBACKS_KEY = "local_only_saved_fallbacks";
+const SAVED_SESSION_OVERRIDES_KEY = "local_only_saved_session_overrides";
 const MODE_KEY = "local_only_mode";
+
+const AGENTS_DIR = process.env.OPENCLAW_AGENTS_DIR || "/home/clawbox/.openclaw/agents";
+
+// Fields on each entry of `<agent>/sessions/sessions.json` that OpenClaw
+// reads to pick which provider/model the ongoing session is bound to.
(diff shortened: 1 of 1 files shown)


KrasimirKralev:
@coderabbitai review

PR #75, opened 2026-04-18
fix(chat): auto-reconnect when gateway restarts to load a new skill

KrasimirKralev:
## Summary
Installing a skill from the App Store restarts the gateway to register the skill's tools. On Jetson the restart often exceeds the chat's 24 s retry budget, so the user sees a 'Could not connect to gateway' banner + a manual Try-again button — which works, but shouldn't be needed.

- Quadruple the retry budget while a skill install is in-flight (8 → 32 attempts = ~96 s)
- Hide the error banner while the 'Reloading skills…' overlay is already showing
- Safety net useEffect: auto-retry if the chat ever lands in error state while reloadingSkill is still true

## Test plan
- [x] Installed a skill on Jetson — chat reconnected automatically, no manual button click needed
- [ ] CodeRabbit review

### DIFF
--- a/src/components/ChatPopup.tsx
+++ b/src/components/ChatPopup.tsx
@@ -10,6 +10,10 @@ import * as kv from '@/lib/client-kv'
 
 const MASCOT_LINES_KEY = 'clawbox-mascot-convo-lines'
 const MAX_RETRIES = 8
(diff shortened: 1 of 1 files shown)


KrasimirKralev:
@coderabbitai review

PR #76, opened 2026-04-18
fix(desktop): wrap icons into multiple columns when arrange-icons runs

KrasimirKralev:
## Summary
Right-click → 'Arrange icons' piled every desktop icon into col: 0, stacking them vertically without end. With more icons than fit the viewport height, the bottom ones ran off-screen and the taskbar clipped the rest — user had to zoom the browser out to reach them.

Desktop path now computes rows-per-column from window.innerHeight (reserving ~72 px for the taskbar) and fills column-by-column top-to-bottom, wrapping to the next column when the current one is full. Mobile path unchanged.

## Test plan
- [x] Verified on Jetson via the ClawBox desktop with 9 installed apps — Arrange icons now produces a tidy multi-column grid within the viewport, no vertical overflow.
- [ ] CodeRabbit review

### DIFF
--- a/src/app/page.tsx
+++ b/src/app/page.tsx
@@ -615,9 +615,22 @@ function ChromeDesktopInner() {
         positions[id] = { row: Math.floor(i / cols), col: i % cols };
       });
     } else {
-      // Desktop: single column, top-left aligned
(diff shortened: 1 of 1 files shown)


KrasimirKralev:
@coderabbitai review

PR #72, opened 2026-04-18
fix(telegram): show phased overlay + refresh bot info on Settings reconfigure

KrasimirKralev:
## Summary
fix(telegram): show phased overlay + refresh bot info on Settings reconfigure

The Settings reconfigure flow showed only a spinner with 'Connecting...'
while the backend restarted the gateway and waited for it to come back —
which can take 10+ seconds. Users got no progress feedback and the bot
name/link in the configured card kept showing the old value after save.

Changes:
- Extract the phased progress overlay from TelegramStep into a shared
  TelegramConfiguringOverlay component (5 phases: token verified →
  connecting → restarting gateway → waiting for gateway health → ready).
- Use the shared overlay in both TelegramStep (setup wizard) and
  SettingsApp (desktop Settings app).
- Refactor the telegram/status fetch into a reusable refreshTelegramStatus
  callback and re-invoke it in the overlay's onDone, so the bot name,
  @username, and 'Open in Telegram' link update to reflect the newly
  configured bot without a page reload.

## Test plan
- [x] Deployed and verified live on Jetson Orin during development session (2026-04-18)
- [ ] CodeRabbit bot review (triggered automatically on draft open)

## Summary by CodeRabbit

* **New Features**
  * Added animated progress display during Telegram configuration showing real-time setup steps and status updates.

* **Refactoring**
  * Reorganized Telegram setup component structure and improved state management for the configuration flow.

PR #78, opened 2026-04-18
fix(gateway): idempotent pre-start hook — drops gateway restart from ~80s to ~10s

KrasimirKralev:
## Summary
`scripts/gateway-pre-start.sh` ran 7 sequential `openclaw config set` calls on every gateway start, one per key under `gateway.controlUi.*` / `gateway.bind` / `gateway.mode` / `gateway.auth.*`. On a Jetson Orin Nano the CLI spends ~10 s on Node.js + SDK init per call, so pre-start sat at ~70 s before the gateway itself started loading. Observable result on the desktop: OpenClaw's SPA (loaded in an iframe) polls gateway endpoints, gets nothing, renders a "Reload gateway" prompt. Clicking it "worked" only because the delay had elapsed.

## What changed
- Single `python3` heredoc block reads `/home/clawbox/.openclaw/openclaw.json`, compares against the desired controlUi / auth / bind / mode state, and writes back atomically via tempfile + `os.replace` — only when at least one field actually differs.
- `allowedOrigins` compared as a set so dynamic LAN-IP reordering doesn't force a rewrite.
- All 7 `openclaw config set` calls removed. MCP server registration kept on the CLI since it fires only on fresh installs / factory resets and uses the CLI's richer schema validation.
- Preserved: LAN IP enumeration, corrupt-JSON recovery (fall back to empty cfg rather than refuse to boot), hostname parse guard, `set -euo pipefail`.

## Why this matters beyond first boot
The gateway restarts on skill install/uninstall, Telegram reconfigure, AI-provider change, Local-only toggle, chat model switch, browser-integration enable/disable, and every crash-triggered restart. Each of those ate ~70 s. Several recent PRs added compensating defensive budgets to work around it (PR #67 bumped config-set timeout to 30 s, PR #72 gave the Telegram overlay a 60 s health-poll timeout, PR #75 quadrupled ChatPopup's reconnect budget during skill installs). Fixing it at the source keeps those safety margins but stops them from being tight.

## Measured on Jetson
- Before: pre-start ~70 s + gateway startup ~10 s = **~80 s** to ready.
- After: pre-start ~0.1 s + gateway startup ~10 s = **~10 s** to ready.
- Subsequent no-op runs log `Gateway config already correct, skipping write` and exit in <100 ms.
- Write path tested by corrupting a value via Python and rerunning — correctly restored to target, file ownership `clawbox:clawbox` preserved, permissions `600` preserved.

## Test plan
- [x] `bash -n` syntax-check passed
- [x] Idempotent no-op path: 2 consecutive runs both log "already correct, skipping write" and complete in ~0.1 s each
- [x] Write path: corrupt `controlUi.allowInsecureAuth = False`, rerun, confirm fixed back to `True`; same for `gateway.mode = "remote"` → restored to `"local"`
- [x] File ownership + perms preserved when the script runs as the `clawbox` user (matches systemd `User=clawbox` for `clawbox-gateway.service`)
- [x] Live `systemctl restart clawbox-gateway` — pre-start completes within the same second as startup, compared to the prior ~30 s gap between `Starting` and the first `[gateway] loading configuration…` log line
- [ ] CodeRabbit review

### DIFF
--- a/scripts/gateway-pre-start.sh
+++ b/scripts/gateway-pre-start.sh
@@ -1,5 +1,21 @@
 #!/usr/bin/env bash
 # Ensure gateway config is valid before OpenClaw gateway starts.
+#
+# Previous versions of this script invoked `openclaw config set` once per
+# key (7 keys × ~10 s CLI startup on Jetson = ~70 s of dead time between
+# systemd "starting" and the gateway actually listening on LAN). During
+# that window, the desktop's OpenClaw iframe polls gateway endpoints,
+# gets refused, and renders a "Reload gateway" prompt. Clicking it
+# worked because the delay had elapsed by then — user-hostile but
+# functional.
+#
+# Now we do a single read-modify-write on openclaw.json in Python.
+# Values that already match what the gateway expects don't get touched
+# (so `meta.lastTouchedAt` doesn't flap on every restart), and the
+# whole script completes in < 1 s. This shaves ~70 s off every gateway
+# restart — not just first boot, but skill install/uninstall, Telegram
+# reconfigure, AI-provider change, Local-only toggle, chat model
+# switch, and crash-triggered restart.
 set -euo pipefail
 
 OPENCLAW_BIN="/home/clawbox/.npm-global/bin/openclaw"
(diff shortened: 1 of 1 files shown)


KrasimirKralev:
@coderabbitai review

KrasimirKralev:
@coderabbitai full review

PR #80, opened 2026-04-18
feat(ui): spinner next to Local-only toggle while request is in flight

KrasimirKralev:
## Summary
Toggling Local-only mode kicks off ~10-30 s of backend work on Jetson (session override sweep + 2 `openclaw config set` calls + gateway restart). The toggle sat at 50% opacity and disabled during that window with no other affirmative feedback, so users assumed the button broke or re-clicked repeatedly.

Adds a small `progress_activity` material-symbols spinner to the left of the toggle while `localOnlyPending` is true. Also sets `aria-busy` so screen readers announce the in-flight state. Toggle itself stays disabled during the request, same as before.

No behavior change beyond the visual affordance.

## Test plan
- [x] Live on Jetson: toggle flips → spinner appears + toggle disabled (50% opacity) → backend completes → spinner disappears + toggle updates to new state
- [ ] CodeRabbit review

### DIFF
--- a/src/components/SettingsApp.tsx
+++ b/src/components/SettingsApp.tsx
@@ -1948,19 +1948,31 @@ export default function SettingsApp({ ui }: SettingsAppProps) {
                       Route everything to the local model. Disables all cloud AI providers (including fallbacks).
                     </p>
                   </div>
-                  <button
-                    type="button"
-                    role="switch"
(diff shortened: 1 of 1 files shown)


KrasimirKralev:
@coderabbitai review

PR #81, opened 2026-04-18
fix(local-ai): unlock switching between local providers — 'Switch to Gemma 4' button + radio lock

KrasimirKralev:
## Summary
Two coupled bugs in the Local AI settings panel:

### 1. No button to switch to Gemma when Ollama is active
When Gemma is installed but Ollama is the currently-active local provider, clicking the Gemma radio rendered only the green "Gemma 4 is already configured" pill (added in #71) with no way to actually switch. User was stuck on Ollama.

**Fix:** new `llamaCppIsActive` prop (computed from `normalizedCurrentProvider === "llamacpp"`). Pill now renders only when installed AND active. When installed-but-not-active, the button stays visible and reads **"Switch to Gemma 4"** (vs "Enable Gemma 4" for the not-installed case).

### 2. Radio auto-flips back to the current provider
Selecting Gemma would flip back to Ollama within a second. Settings polls `localAiStatus` → updates `currentProviderId` → the sync `useEffect` called `setSelectedProvider(normalizedCurrentProvider)` on every tick, clobbering the user's click.

**Fix:** `userSelectedProviderRef` starts false, flips true inside `selectProvider`. Sync effect skips the provider-level set when true (model-slug sync inside the active provider still runs since it doesn't race).

### `/simplify` pass
Flattened the 4-branch description ternary and 3-branch button-label ternary in `LlamaCppModelPanel` into self-documenting local variables (`canSwitchToGemma`, `showConfiguredPill`, `description`, `buttonLabel`). Semantics unchanged.

## Test plan
- [x] Live on Jetson — Ollama active, click Gemma radio → radio stays, panel shows "Switch to Gemma 4" orange button + matching description
- [x] Click Switch to Gemma 4 → Gemma becomes active, panel updates to green "Gemma 4 is enabled and running" pill
- [x] Radio no longer auto-flips back during polling
- [x] Bun build clean, service healthy
- [ ] CodeRabbit review

### DIFF
--- a/src/components/AIModelsStep.tsx
+++ b/src/components/AIModelsStep.tsx
@@ -561,6 +561,13 @@ export default function AIModelsStep({
     [genericSteps, llamaCppInstallSteps, ollamaInstallSteps],
   );
   const [selectedProvider, setSelectedProvider] = useState<string | null>(resolvedDefaultProvider);
+  // Once the user picks a radio (via selectProvider), stop auto-syncing the
+  // selection from the currently-active provider. Without this, polling on
+  // the parent (Settings refreshes localAiStatus every few seconds, which
+  // bumps currentProviderId) would flip the radio back to whatever is
+  // currently active — so a user on Ollama who clicks "Gemma 4" sees the
+  // radio flip straight back to Ollama within a second.
+  const userSelectedProviderRef = useRef(false);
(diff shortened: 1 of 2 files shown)


KrasimirKralev:
@coderabbitai review

KrasimirKralev:
@coderabbitai review

PR #84, opened 2026-04-19
fix(updater): skip openclaw re-install when already up-to-date

KrasimirKralev:
## Summary

- `step_openclaw_install` compared `openclaw --version` output (`OpenClaw X.Y.Z (hash)`) against the bare npm version tag with literal `=`. Those strings never match, so `npm install -g openclaw@$TARGET` re-ran on every update — reinstalling the same version, taking 2-3 min on Jetson, blowing through the 120 s step timeout, and surfacing as a scary `X` on 'Updating OpenClaw' even when nothing needed to change.
- Extract the bare version with `awk '{print $2}'` and compare exactly. Happy-path now returns in ~1 s instead of ~150 s.
- Bump the step timeout 120 s → 300 s via a shared `OPENCLAW_INSTALL_TIMEOUT_MS` module constant so `UPDATE_STEPS` and `OPENCLAW_UPDATE_STEPS` can't drift. 300 s matches the sibling `chromium_install` / `vnc_install` budgets.

Applies to both `install.sh` (Jetson/ARM) and `install-x64.sh` — they're deliberately parallel.

## Symptoms this fixes

- Fresh Jetson setup: 'Updating OpenClaw' flashes an X mid-run (install really did finish, the 150 s UI timeout just fired first)
- Every subsequent update: same X, because the idempotency check never triggers
- `UpdateOverlay` shows 'Update Failed' at the top even when the rebuild+reboot succeeded

## Test plan

- [ ] Fresh Jetson from `main`: trigger update to `beta` — 'Updating OpenClaw' completes without the X
- [ ] Repeat update with no version bump: step returns in ~1 s (log: `OpenClaw is already up to date`)
- [ ] Force a version bump server-side: install actually runs and has 5 min budget instead of 2

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced version detection mechanism for installed OpenClaw binaries to ensure proper identification of installation status.
  * Extended maximum installation timeout from 120 to 300 seconds across update workflows to accommodate slower systems and reduce timeout-related failures.


KrasimirKralev:
@coderabbitai review

KrasimirKralev:
Closing — branch was cut from beta, retargeting to main surfaced unrelated beta-only commits in the diff. Reopening from a clean main-based branch.

PR #86, opened 2026-04-19
fix(settings): redirect to home after full update completes

KrasimirKralev:
## Summary

- After a full update finishes, the overlay showed 'Update complete / Restarting device...' forever. The device doesn't actually reboot on that step — only `clawbox-setup.service` restarts — so the user was stuck at a terminal UI with no way back.
- Auto-dismiss effect now handles both paths:
  - Full update (has `restart` step): 5 s grace then `window.location.href = '/'` — hard-nav picks up the freshly built client bundle
  - Scoped update (OpenClaw-only): unchanged, 3 s in-place dismiss

## Before / after

Before: overlay stays on 'Update complete' with 'Рестартиране на устройството...' subtitle indefinitely.
After: overlay auto-closes and drops user on the desktop at `/`.

## Test plan

- [ ] Trigger Settings → System Update on a device already on latest: overlay closes after ~5 s, lands on desktop
- [ ] Trigger Settings → OpenClaw Update (scoped): overlay closes after ~3 s, Settings window remains open
- [ ] Trigger a failing update (e.g. kill gateway mid-run): overlay stays on failure screen (no auto-redirect)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved update completion workflow: updates requiring device restart now provide a 5-second grace period before automatic redirection to the home page. Enhanced overlay management ensures smoother transitions between different update states. Refined visual status indicators provide clearer feedback about device restart requirements and completion status throughout the update process.

### DIFF
--- a/src/components/SettingsApp.tsx
+++ b/src/components/SettingsApp.tsx
@@ -12,7 +12,7 @@ import AIModelsStep from "./AIModelsStep";
 import TelegramConfiguringOverlay from "./TelegramConfiguringOverlay";
 import { I18nProvider, useT, LANGUAGES, type Locale } from "@/lib/i18n";
 import { QRCodeSVG } from "qrcode.react";
-import type { UpdateState } from "@/lib/updater";
+import { RESTART_STEP_ID, type UpdateState } from "@/lib/updater";
 import { cleanVersion } from "@/lib/version-utils";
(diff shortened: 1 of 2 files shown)


KrasimirKralev:
@coderabbitai review

KrasimirKralev:
Pushed simplifier pass — exported `RESTART_STEP_ID` from `@/lib/updater` and replaced the two `"restart"` string literals in SettingsApp with the typed constant. @coderabbitai review

PR #88, opened 2026-04-19
fix(apps): sync installed_apps + installed_meta on MCP/CLI install

KrasimirKralev:
## Summary

Installs triggered via the MCP `app_install` tool (or CLI) landed the skill on disk but were **invisible on the desktop and in the Store's Installed tab**. The Store UI filters \"Installed\" by the `installed_apps` preference, and the desktop filters icons by both `installed_apps` AND `installed_meta`; those were only written as a client-side side-effect of the AppStore UI install flow. Off-desktop installs flew under that radar.

This PR makes the server-side routes the source of truth:
- `src/app/setup-api/apps/install/route.ts` now writes both `pref:installed_apps` (id) and `pref:installed_meta` (name + color + iconUrl)
- `src/app/setup-api/apps/uninstall/route.ts` mirrors the delete
- `src/lib/store-categories.ts` (new) holds the category→color map previously inlined in `AppStore.tsx`, so client and server stay in step
- `src/app/page.tsx` and `src/components/AppStore.tsx` import the shared `InstalledMeta` type / colour map

## Behaviour

- Fresh install (MCP or UI): skill on disk + both prefs populated → appears on desktop and Store after page refresh
- Re-install: short-circuits — no Store fetch, no config write, so MCP retries don't thrash the upstream API
- Offline / Store unreachable: falls back to a title-cased slug + neutral color + local icon path; app still renders rather than being silently dropped
- Uninstall: both prefs cleaned up

## Scope / non-goals

**Live refresh is not in this PR.** The desktop's React state is seeded on mount, so newly installed skills appear after a hard refresh (Ctrl+Shift+R). Wiring server→client push (SSE) is a deferred follow-up — kept this PR small and focused on the correctness fix.

## Test plan

- [x] MCP `app_install` via Gemma chat: skill on disk + preferences populated (verified on Jetson at 192.168.1.57)
- [x] After page refresh: skill shows in Store's Installed tab and on the desktop with correct icon + color
- [x] Re-install of already-installed app: no Store API fetch, no config write (short-circuit)
- [x] UI install still works (regression check — same meta populated via existing client path)
- [ ] Uninstall via MCP: prefs cleaned up → skill disappears after refresh

## Summary by CodeRabbit

* **New Features**
  * App installs now fetch and store app metadata (name, color, icon, optional launch URL) with deterministic fallbacks.

* **Bug Fixes**
  * Uninstall now removes associated metadata and installed-app entries to avoid orphaned preferences.

* **Improvements**
  * Centralized category color definitions for consistent styling and matched store metadata synchronization.

### DIFF
--- a/src/app/page.tsx
+++ b/src/app/page.tsx
@@ -21,6 +21,7 @@ import SetupWizard from "@/components/SetupWizard";
 import { I18nProvider, useT } from "@/lib/i18n";
 import { cleanVersion } from "@/lib/version-utils";
 import { FEATURE_FLAG_KEYS, isFeatureFlagEnabled } from "@/lib/feature-flags";
+import type { InstalledMeta } from "@/lib/store-categories";
 
 
 const Mascot = dynamic(() => import("@/components/Mascot"), { ssr: false });
@@ -216,7 +217,7 @@ function ChromeDesktopInner() {
   const [date, setDate] = useState("");
   const [installedApps, setInstalledApps] = useState<string[]>([]);
   const [recentlyInstalled, setRecentlyInstalled] = useState<string | null>(null);
-  const [installedMeta, setInstalledMeta] = useState<Record<string, { name: string; color: string; iconUrl: string }>>({});
+  const [installedMeta, setInstalledMeta] = useState<Record<string, InstalledMeta>>({});
 
   // ─── Desktop shortcuts for built-in apps ───
   const [desktopApps, setDesktopApps] = useState<string[]>(DEFAULT_DESKTOP_APPS);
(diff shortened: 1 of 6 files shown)


KrasimirKralev:
@coderabbitai review

KrasimirKralev:
Addressed all 3 findings from the second review on 7a1f869: surface preferenceSyncError to callers + flip `ok` when set, hasOwnProperty guard on CATEGORY_COLORS, don't overwrite existing meta on partial-state re-install. Tests 24/24 pass, live-tested on device. @coderabbitai review

PR #89, opened 2026-04-19
fix(browser): run Next.js under Node to unblock Playwright CDP

KrasimirKralev:
## Summary

Agents calling \`browser_launch\` / \`browser_open\` (and the ClawBox desktop BrowserApp) failed with \"Browser control is currently blocked by a CDP connection timeout (\`ws://127.0.0.1:18800\`)\" regardless of which model was driving. CDP HTTP responded correctly (\`/json/version\` returned a valid WS URL) and a raw WebSocket upgrade via curl worked, but Playwright's \`connectOverCDP\` hung until its own timeout.

Root cause: \`clawbox-setup.service\` ran \`production-server.js\` under Bun. Bun's WebSocket client has incompatibilities with Playwright's CDP handshake — the WS never completes. Running the same Next.js standalone bundle under Node.js 22 attaches in ~2 s. \`CLAUDE.md\` already notes that Node is the right runtime for this binary (\"Bun lacks HTTP upgrade events for WebSocket proxy\"); this PR makes the systemd unit match.

## Changes

- \`config/clawbox-setup.service\`: \`ExecStart\` now \`/usr/bin/node\` instead of \`bun\`; explains why inline
- \`install-x64.sh\`: matching change for the non-systemd dev path
- \`src/app/setup-api/browser/route.ts\`: caches one shared Playwright \`Browser\` per process instead of re-attaching on every \`launch\`. Sessions hold just page refs; stale cleanup closes the page, leaving CDP intact. Saves ~1 s on repeat launches and avoids piling up dead CDP clients. \`connectOverCDP\` gets an explicit 10 s timeout so failures surface cleanly. HTTP probe bumped 1 s → 3 s to ride out Jetson load spikes.
- \`src/tests/routes/browser.test.ts\`: updated mocks (\`isConnected\`, \`page.close\`) and close-action assertion to match the shared-connection model

## Verified on Jetson

```
# Under Bun (before):   10 s timeout, error surfaces to agent
# Under Node (after):   first launch 1.8 s, second launch 1.1 s (cache hit)
```

## Test plan

- [x] \`bun run test\` — 23/23 pass
- [x] Live: \`curl POST /setup-api/browser action=launch\` returns session + screenshot
- [x] Live: second launch completes faster (shared browser reuse)
- [ ] Agent chat: \"Open the Clawbox website\" now succeeds under both Gemma and Codex (user verification)

### DIFF
--- a/config/clawbox-setup.service
+++ b/config/clawbox-setup.service
@@ -7,7 +7,13 @@ Wants=network-online.target
 Type=simple
 User=clawbox
 WorkingDirectory=/home/clawbox/clawbox
-ExecStart=/home/clawbox/.bun/bin/bun run production-server.js
+# Runs production-server.js under Node, not Bun. Bun's WebSocket client
+# has incompatibilities with Playwright's `connectOverCDP` — the CDP WS
+# handshake to desktop Chromium never completes under Bun, which the agent
+# surfaces as "Browser control is currently blocked by a CDP connection
+# timeout (ws://127.0.0.1:18800)". Node.js 22 runs the same standalone
+# bundle fine. Bun is still used for build (`bun run build`).
+ExecStart=/usr/bin/node /home/clawbox/clawbox/production-server.js
 Restart=always
 RestartSec=3
 EnvironmentFile=-/home/clawbox/clawbox/data/network.env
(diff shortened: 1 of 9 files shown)


KrasimirKralev:
@coderabbitai review

KrasimirKralev:
Pushed simplifier consensus fixes (3 commits now). Key changes: connectOverCDP timeout 10s→30s (Playwright default), nullify cachedBrowserPromise on disconnect race, retry loop 10→6 iterations, sqlite-store narrower try/catch + fallback log, package.json start → node, new test for shared-CDP reuse (24/24 passing). @coderabbitai review

PR #83, opened 2026-04-18
fix(ai-provider): sweep sessions + show switch progress + green banner

KrasimirKralev:
## Summary
Changing the primary AI provider via Settings (e.g. OpenAI → ClawBox AI) had three visible problems:

1. **Chat kept responding as the OLD provider.** Configure route updated `agents.defaults.model.primary` but left existing sessions' per-session overrides pointing at the previous model. OpenClaw's resolver respected those sticky overrides on the next turn.
2. **Chat "froze" during the gateway restart** — popup only had skill-install-aware retry + overlay logic, so a provider-change restart dropped the WS silently and user had to refresh.
3. **No affirmative feedback after the switch** — dropdown already shows a green "Switched chat to X" banner when switching models from inside the chat pane (PR #73); provider changes via Settings deserved the same.

## Changes
- **`configure/route.ts`** — after setting `agents.defaults.model.primary`, sweep per-session overrides via `applyModelOverrideToAllAgentSessions` with `source: "user"` so OpenClaw's resolver keeps the new primary bound across existing sessions. Mirrors PR #73's sweep. Only runs when the call actually set a new primary (skips local-only scopes).
- **`ChatPopup.tsx`** — listens for `clawbox:primary-ai-configured` (already dispatched by SettingsApp) alongside `clawbox-skill-installed`. Reuses the overlay + progress bar + quadrupled retry budget. Banner label adapts (`"Switching AI provider..."` vs `"Reloading skills..."`) via a new `reloadReason` state + escape-hatch `reloadReasonRef` (one-time-captured WebSocket `hello` resolve closure can't see state changes). After reconnect on a provider change, auto-"my skills were updated" prompt is skipped and a green `"Switched chat to <label>"` banner is pushed using refreshed `chat/model` state.
- **`chat/model/route.ts`** — `PROVIDER_LABELS` was missing `"openai-codex"`, so `labelForProvider` fell back to the generic `"AI Provider"` placeholder, making the new banner read `"Switched chat to AI Provider"` when switching to or from openai-codex. Added `"openai-codex": "OpenAI Codex"`.

## `/simplify` pass
Converted the inline `fetch().then().then()` chain in ChatPopup's reload branch to `async/await` (matches the existing `refreshChatModelState` pattern in the same file + CLAUDE.md preference). Semantics unchanged.

## Test plan
[...]

PR #82, opened 2026-04-18
fix(agent): seed CLAWBOX.md so the agent knows device conventions

KrasimirKralev:
## Summary
The OpenClaw agent's session-start context (AGENTS.md, BOOTSTRAP.md, etc.) has no ClawBox-specific guidance. Observed failure mode earlier today: user installed `self-improving-agent` via the App Store, asked the agent to confirm — the agent `read`-checked `/home/clawbox/.npm-global/lib/node_modules/openclaw/skills` (the npm package's built-in skills dir), saw nothing, and confidently reported the skill as not installed. The skill was actually at `~/.openclaw/workspace/skills/self-improving-agent/` the whole time.

Same class of problem applies to the desktop-Chromium integration and to skill installs:

- Agent doesn't know the `browser_*` MCP tools exist → defaults to `ui_open_app("browser")` which opens the Setup panel, not a browsing session.
- When asked to install a skill the agent might copy files into the skills dir manually instead of using `app_install` — which skips the App Store registration + gateway reload the user needs to see / uninstall it from the UI.

## What changed
1. New `config/clawbox-workspace-guide.md` documents:
   - Where user-installed skills live (`~/.openclaw/workspace/skills/`) and **do NOT** live (the npm-global path)
   - Skill install/uninstall must go through `app_install` / `app_uninstall` (so the App Store UI stays consistent + gateway reload actually fires)
   - Full `browser_*` MCP tool inventory with the standard launch → screenshot → click/type workflow, plus the "do not use `ui_open_app("browser")`" caveat
   - `ui_*` / `app_*` / `webapp_*` tools inventory
   - File-system + network pointers

2. `scripts/gateway-pre-start.sh` now seeds this file as `~/.openclaw/workspace/CLAWBOX.md` when:
   - The workspace dir exists (avoids racing first-boot OpenClaw init)
   - Shipped template differs from on-disk copy (`cmp -s`) — so later agent/user edits aren't clobbered on every restart
   
3. Also appends a one-liner reference to `AGENTS.md` (when it exists and doesn't already mention CLAWBOX.md), so the agent's normal session-start bootstrap loads our guide without us having to overwrite AGENTS.md.

## Test plan
- [x] `bash -n` syntax clean
- [x] First run on Jetson: seeded CLAWBOX.md + appended AGENTS.md reference
- [x] Second run: both checks logged "skipping" (idempotent)
- [x] AGENTS.md now ends with `## ClawBox integration` section pointing at CLAWBOX.md
[...]

PR #90, opened 2026-04-19
fix(tests): unblock 14 pre-existing failures in ai-models test suite

KrasimirKralev:
## Summary

14 tests had been red on beta since the Gemma / provider-switch work in PRs #82/#83 landed. CI was useless as a signal — every PR inherited 14 failures regardless of actual correctness. This PR is test-only — no production code changes — and gets CI back to green so real regressions stop hiding in the noise.

## Root causes (three independent)

1. **Mock drift from configure route's real imports.** The route now calls `applyModelOverrideToAllAgentSessions` and `parseFullyQualifiedModel` (added in #83 for session sweeping), plus four getters from `@/lib/llamacpp` and one from `@/lib/local-ai-runtime`. The test mocked none of these. First request-time call threw, outer catch returned 500, every test asserting 200 cascaded red. Added the missing mocks with real-shape implementations.

2. **vitest-under-bun clears mock implementations along with call history.** Factory defaults set inside `vi.mock(...)` survive `vi.resetModules` but not `vi.clearAllMocks()`. Tests passed in isolation, failed in sequence. Re-apply the implementations in `beforeEach`.

3. **`ai-models-step.test.tsx` relied on `llamaCppIsActive` defaulting true.** The panel only sets it true when `currentProviderId === "llamacpp"`. Added the prop.

## Simplifier findings applied

- Mirror real `parseFullyQualifiedModel` byte-for-byte (`idx <= 0 || idx === fq.length - 1`) so trailing-slash inputs reject as expected. Fixed the same drift in `chat-model.test.ts`.
- Extract shared proxy-URL constant via `vi.hoisted` to avoid duplicating a magic string across two mock factories.

## Result

Before: 14 failed / 1055 passed (93 files, 2 red)
After: **1069 passed** (93 files, all green)

## Test plan

- [x] `bun run test` — 1069/1069 pass locally
- [ ] CI should now show `test: SUCCESS` for the first time in weeks

### DIFF
--- a/src/tests/components/ai-models-step.test.tsx
+++ b/src/tests/components/ai-models-step.test.tsx
@@ -104,6 +104,11 @@ describe("AIModelsStep variants", () => {
         description="Local models first"
         configureScope="local"
         testId="local-ai-test"
+        // Gemma is currently the active chat provider — required for the
+        // panel to render the "configured" pill rather than the orange
+        // "Switch to Gemma 4" call-to-action (which appears when Gemma is
+        // on disk but some other provider is primary).
+        currentProviderId="llamacpp"
       />,
     );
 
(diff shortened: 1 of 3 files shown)


KrasimirKralev:
@coderabbitai review

PR #93, opened 2026-04-20
fix(telegram): stop opening bot DMs to everyone by default [security]

KrasimirKralev:
## Summary

`setTelegramToken` (and the matching shell paths in `install.sh` / `install-x64.sh`) silently wrote `dmPolicy: "open"` + `allowFrom: ["*"]` the moment a user submitted a bot token through the setup wizard. Combined with the agent's default tool surface (shell, file IO, `system_power`, etc.), this meant **any Telegram user who could find the bot handle** could DM it and trigger arbitrary commands on the Jetson. OpenClaw's upstream default is `dmPolicy: "pairing"` — every new sender has to be approved by the owner via a pairing code. We now delegate to that default.

## Changes

- **`src/lib/openclaw-config.ts`** — `setTelegramToken`: drop the two overrides. Strip `dmPolicy`/`allowFrom` from any pre-existing config via destructure-and-rest so a device that already has the insecure values stored gets re-secured on the next reconfigure without a factory reset. Also added `allowFrom?: string[]` to the channel interface (previously only in the index signature).
- **`install.sh`** — same fix on the fresh-install path (was writing the insecure JSON directly via `openclaw config set`).
- **`install-x64.sh`** — same fix on the x86 dev-install path (Node inline script block).
- **`src/tests/unit/openclaw-config.test.ts`** — inverted the two tests that asserted the old behavior (now using `.not.toHaveProperty` so a future `dmPolicy: undefined` write still fails the assertion); added a new test covering the strip-on-reconfigure path.

## Severity

**High.** An attacker who discovers the bot handle (BotFather handles are publicly enumerable) can trigger `run_command` on the Jetson with no pairing step. Ships a silent bypass of an explicit OpenClaw safety default.

## What this PR does NOT fix (worth a follow-up)

Existing devices that already have `dmPolicy: "open"` + `allowFrom: ["*"]` written to `~/.openclaw/openclaw.json` **stay vulnerable** until the owner either (a) reconfigures the bot token through Settings (triggers the TS strip path), or (b) we add a one-shot migration to `scripts/gateway-pre-start.sh` that removes the insecure override at gateway startup. (b) is the safer fleet-wide path — happy to draft as a follow-up PR if you want.

Also worth flagging in release notes: **anyone whose bot was running under the open policy should rotate their bot token via BotFather**, since the handle may have been discovered in the meantime.

## Test plan

- [x] `bun run test` — 1072/1072 pass on-device
- [x] Verified on a real Jetson that had the insecure config: reconfiguring a bot token strips `dmPolicy`/`allowFrom` from `openclaw.json`
- [ ] Recommend a human security review from `@ID-Robots/id-robots-core-team` before merge, not just `--admin squash`

## Summary by CodeRabbit

* **Bug Fixes**
  * Installer and runtime config now only enable the Telegram bot and set its token; they no longer force default DM permissions.
  * Any existing dmPolicy/allowFrom entries are stripped during token updates and at pre-start to remove insecure defaults.
* **Tests**
  * Unit tests updated to verify the token is set/enabled and that dmPolicy/allowFrom are removed.

### DIFF
--- a/install-x64.sh
+++ b/install-x64.sh
@@ -371,7 +371,11 @@ try {
   const cb=JSON.parse(fs.readFileSync(process.env.CLAWBOX_CONFIG,'utf8'));
   if(cb.telegram_bot_token){
     if(!c.channels)c.channels={};
-    c.channels.telegram={...c.channels.telegram,enabled:true,botToken:cb.telegram_bot_token,dmPolicy:'open',allowFrom:['*']};
+    // dmPolicy/allowFrom intentionally NOT set — OpenClaw's "pairing"
+    // default requires owner approval before the agent responds to a new
+    // sender. See src/lib/openclaw-config.ts:setTelegramToken.
+    const {dmPolicy:_dm,allowFrom:_af,...rest}=c.channels.telegram||{};
+    c.channels.telegram={...rest,enabled:true,botToken:cb.telegram_bot_token};
     process.stderr.write('  Telegram channel registered in OpenClaw config\n');
   }
 } catch {}
--- a/install.sh
+++ b/install.sh
@@ -708,8 +708,37 @@ step_openclaw_config() {
     local TG_TOKEN
     TG_TOKEN=$(node -e "try{const c=JSON.parse(require('fs').readFileSync('$CLAWBOX_CONFIG','utf8'));if(c.telegram_bot_token)process.stdout.write(c.telegram_bot_token)}catch{}" 2>/dev/null || true)
     if [ -n "$TG_TOKEN" ]; then
-      as_clawbox "$OPENCLAW_BIN" config set channels.telegram \
-        "{\"enabled\":true,\"botToken\":\"$TG_TOKEN\",\"dmPolicy\":\"open\",\"allowFrom\":[\"*\"]}" --json
+      # Do not set dmPolicy/allowFrom here — see src/lib/openclaw-config.ts
+      # setTelegramToken for the security rationale. `openclaw config set`
+      # may deep-merge instead of replace, which would leave legacy
+      # `dmPolicy:"open"` / `allowFrom:["*"]` values alive. Do the
+      # read-modify-write ourselves with a destructure-strip so the
+      # installer can't leave insecure values behind, matching what
+      # install-x64.sh and src/lib/openclaw-config.ts:setTelegramToken
+      # do on their paths.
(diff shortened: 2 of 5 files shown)


KrasimirKralev:
@coderabbitai review — security-sensitive, flagging for extra scrutiny on the strip-on-reconfigure semantics and whether `install.sh` + `install-x64.sh` need the same migration-comment pattern as the TS path.

KrasimirKralev:
Addressed the CodeRabbit finding in an amend: added fleet-wide migration to `scripts/gateway-pre-start.sh`. Every gateway start now scrubs `dmPolicy` and `allowFrom` from `channels.telegram` if present, so existing devices re-secure themselves after pulling this fix — no bot-token reconfigure or factory reset needed.

Scrub logic verified locally against 5 cases: insecure config gets stripped, safe config is a no-op, and all three malformed paths (no channels, no telegram, non-dict telegram) no-op without crashing. The existing pre-start atomic-write path + `meta.lastTouchedAt` comparison means the write only happens when there's actually a change, so no log-flapping on already-safe devices.

@coderabbitai review

KrasimirKralev:
Addressed the install.sh finding — replaced `openclaw config set` with an inline Node read-modify-write that destructures `dmPolicy`/`allowFrom` off any pre-existing value before writing. Matches the exact pattern used in `install-x64.sh` and `src/lib/openclaw-config.ts:setTelegramToken`, so all three install paths now behave the same.

Verified end-to-end on the Jetson (Node v22.22.2):
- **Node script:** 7/7 assertions pass — botToken updated, enabled=true, dmPolicy stripped, allowFrom stripped, customField preserved, gateway.port preserved, other channels preserved.
- **gateway-pre-start.sh migration:** injected `dmPolicy:"open"` + `allowFrom:["*"]` into the live openclaw.json, ran pre-start → both keys scrubbed, `changed` flag flipped so rewrite happened, `Updated gateway config` logged. Subsequent runs on the clean config are no-ops.
- **Full test suite:** 1072/1072 pass.

@coderabbitai review

KrasimirKralev:
Applied the ENOENT-only catch: only a missing config file falls back to empty, any parse/EACCES/IO error re-throws so the installer aborts instead of silently replacing a real config with only the telegram channel.

Verified on real Jetson (Node v22.22.2):
- Happy path: 7/7 assertions pass (strip + preserve)
- Missing file: initializes cleanly, botToken written
- **Corrupted JSON** (truncated mid-object): script aborts non-zero, file left untouched — exactly the behavior we want

@coderabbitai review

PR #95, opened 2026-04-21
fix(ai): write OpenRouter provider definition so chat actually works

KrasimirKralev:
## Summary

OpenRouter has been silently broken since the configure route gained local-AI provider definitions: ClawBox writes `auth.profiles.openrouter:default` but **never writes `models.providers.openrouter`**. OpenClaw has no built-in adapter for OpenRouter (unlike openai/anthropic/google), so without that entry the runtime has no baseUrl to call — every chat turn returns `usage: { input: 0, output: 0 }`, the assistant message is empty, and the UI shows nothing.

The only trace in the logs: `[agent/embedded] incomplete turn detected … stopReason=stop payloads=0 — surfacing error to user`.

## What shipped

- **`src/app/setup-api/ai-models/configure/route.ts`** — add an `isOpenRouter` branch that writes `models.providers.openrouter` with `baseUrl: https://openrouter.ai/api/v1`, `api: openai-completions`, and the default model. Mirrors the pattern already used for llamacpp / ollama.
- **`src/tests/routes/ai-models/configure.test.ts`** — new test asserting the openrouter branch writes the provider definition with the correct baseUrl, api, and default model slug. Regression coverage for the silent-failure mode.

The `models` array is UI-only — OpenClaw routes any `openrouter/<slug>` string through the same baseUrl. Verified on-device: after injecting the same provider def this PR writes, `openrouter/anthropic/claude-haiku-4-5` works even though only `moonshotai/kimi-k2-0905` is in the list.

## Verification

- `bun run test` — 1073/1073 pass on real Jetson (was 1072, +1 new test)
- Live-tested on Jetson with the injection this PR automates:
  - Before: kimi and haiku both silent (`payloads=0`)
  - After: both reply normally, usage accounts properly

## Scope / non-goals

- Leaves `models[]` at length 1 (just the default kimi). The UI doesn't have an OpenRouter model picker yet; when it does, this list can grow or fetch live from `/api/v1/models`. Filed as follow-up — not part of this PR.
- Existing devices that set up OpenRouter before this fix need a reconfigure (re-submit the API key through Settings → AI Provider → OpenRouter) to get the provider def written. Separate one-shot migration would be a follow-up if needed; for now the configure flow itself is the migration.

## Test plan

- [x] `bun run test` — 1073/1073
- [x] Fresh OpenRouter setup on Jetson → chat replies
- [x] Switch primary to a non-default OpenRouter model (`anthropic/claude-haiku-4-5`) → chat replies
- [ ] (for reviewers) Fresh install from scratch → setup wizard → enter OpenRouter key → chat works with default model

## Summary by CodeRabbit

* **New Features**
  * Added OpenRouter as a new AI model provider option, enabling seamless integration of OpenRouter services into your setup.
  * Automatically configures OpenAI-compatible API endpoints and provider settings.
  * Includes intelligent fallback model selection and automatic provider mode optimization.

* **Tests**
  * Added comprehensive test coverage for OpenRouter provider configuration and validation.

### DIFF
--- a/src/app/setup-api/ai-models/configure/route.ts
+++ b/src/app/setup-api/ai-models/configure/route.ts
@@ -337,6 +337,7 @@ export async function POST(request: Request) {
     const isOllama = provider === "ollama";
     const isLlamaCpp = provider === "llamacpp";
     const isClawAI = provider === "clawai";
+    const isOpenRouter = provider === "openrouter";
     const isLocalScope = scope === "local";
     if (!provider || (!normalizedApiKey && !isOllama && !isLlamaCpp && !isClawAI)) {
       return NextResponse.json(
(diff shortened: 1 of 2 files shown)


KrasimirKralev:
@coderabbitai review

PR #129, opened 2026-05-07
feat(install): post-install validation step (services + WiFi AP + dashboard)

KrasimirKralev:
## Summary
- Adds `step_validate_services` as the final `[23/23]` step in `install.sh`. Polls each expected systemd unit, plus a Wi-Fi AP-mode probe (`iw dev`) and a dashboard HTTP probe (`curl http://localhost/`), for up to 30 s before failing with a per-unit table that includes a `systemctl status` snippet for each bad unit.
- Promotes the previously-local `ALL_SERVICES` array in `step_systemd_services` to two module-level constants (`EXPECTED_ACTIVE_SERVICES`, `EXPECTED_INSTALLED_SERVICES`) so install and validation share one source of truth.
- Catches the "install.sh died midway but SSH still exited 0" failure mode where a `flash.sh` run reports success while services were never installed. Failure propagates through `set -euo pipefail` → SSH → `flash.sh`'s existing per-box FAIL counter — no changes to `flash.sh` needed. Also addressable as `--step validate_services` for re-running on demand.

## Test plan
- [x] Fail-path on host (no clawbox services installed): 11/11 checks fail, exit=1, clean output.
- [x] Fail-path on a Jetson with a partial install: `clawbox-performance.service` correctly reported as `enabled=enabled active=inactive` with `systemctl status` snippet; other units flagged missing; both functional probes correctly fail; exit=1.
- [x] Pass-path on a Jetson with a complete install: `[23/23] Validating services... ✓ All 11 checks healthy (services + WiFi AP + OpenClaw dashboard)`, exit=0.
- [x] External verification: `ClawBox-Setup` SSID visible from a separate machine at full signal; `iw dev` reports `type AP` with txpower 15 dBm; `curl http://<box-ip>/` returns HTTP 307.

## Out of scope
`install.sh:238-242` node-pty rebuild bug — forces `npm_config_nodedir` at incomplete system headers and breaks the build at step 8/23 on fresh flashes. Manual unblock used during pass-path testing here; tracked as a separate follow-up.

## Summary by CodeRabbit

* **Chores**
  * Installer adds a post-install health validation that checks required services are present, enabled and running, confirms the device is broadcasting Wi‑Fi in AP mode, and verifies the local dashboard responds over HTTP. Failures are reported with service status snippets and cause the install to fail if checks don't pass within the validation window. Installation flow now runs this validation immediately after starting services.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -39,6 +39,29 @@ GATEWAY_DIST="$NPM_PREFIX/lib/node_modules/openclaw/dist"
 DNSMASQ_DIR="/etc/NetworkManager/dnsmasq-shared.d"
 AVAHI_CONF="/etc/avahi/avahi-daemon.conf"
 
+# ── Service registry ─────────────────────────────────────────────────────────
+# Authoritative list of clawbox systemd units. Used by step_systemd_services
+# (install/enable) and step_validate_services (post-install health check).
+#
+# ACTIVE: must be enabled and active after install.
+# INSTALLED: unit file must exist on disk but the unit is not required to be
+# active because it is opt-in (clawbox-tunnel, clawbox-browser), one-shot
+# driven by a timer (clawbox-heartbeat.service), or a template that is never
+# activated standalone (clawbox-root-update@.service).
+EXPECTED_ACTIVE_SERVICES=(
+  clawbox-ap.service
+  clawbox-setup.service
+  clawbox-gateway.service
+  clawbox-performance.service
+  clawbox-heartbeat.timer
+)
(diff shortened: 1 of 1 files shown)


yalexx (dismissed):

yalexx (approved):
good

yalexx (approved):

yalexx (approved):
working

yalexx (approved):
approved after latest push

GeorgiK77 (approved):
working!!

PR #140, opened 2026-05-19
Merge beta into main: v3.0.4 release

KrasimirKralev:
Promotes the v3.0.3 + v3.0.4 work from `beta` to production. All commits have been live on `beta` and validated on physical Jetson hardware (`krasi.local`). CodeRabbit + CI green on each component PR before merge.

## v3.0.4 bundle (#139)

- **fix(remote-control):** Cloudflare Quick Tunnel panel keeps polling so the URL appears without a manual Refresh.
- **feat(picker):** Model picker merges live OpenClaw catalog with the static `PROVIDER_CATALOGS` and filters retired Claude IDs via a `DEPRECATED_MODEL_IDS` denylist. Dropdown surfaces when `allowCustom` is set so users can type their own model ID against Anthropic OAuth.
- **feat(clawkeep):** Pair card redesigned to match the ClawAI subscription panel — explicit 'Open authorization page' button so the device code is readable before focus shifts, 'Get a new code' affordance for expired codes, reuse of the existing `ai.openAuthPage` translation key.
- **feat(vnc):** Bidirectional UTF-8 clipboard via a new `/setup-api/vnc/clipboard` route that shells out to `xclip` on the guest X CLIPBOARD. Replaces the basic RFB ClientCutText path that mangled Cyrillic / CJK / emoji as Latin-1 mojibake. Stdout capped at 1 MiB and Content-Length pre-checked on POST to protect the Jetson from memory spikes. Requires `xclip` installed on the device.
- **fix(clawai-tier, client):** Preserve last-known tier on transient `/setup-api/ai-models/status` failures so the 'You are on the free plan' modal stops popping every time the gateway WebSocket blips. Hook returns the same state ref when no logical change so React bails out.
- **feat(catalog):** Trigger a background catalog refresh when a provider is configured, so the picker reflects the new entitlement without waiting for the next periodic refresh.
- **chore(release):** bump version to 3.0.4.

## Post-3.0.3 hardening (#138)

- **fix(gateway-pre-start):** Re-install `@openclaw/codex` if the nested peer-dep symlink to `openclaw` is missing or dangling. Repro'd live on a customer device where the chat failed with `Cannot find package 'openclaw'`. `-e` follows symlinks, so both 'missing' and 'dangling' trigger a `--force` reinstall, which rebuilds the link without redownloading the package.
- **fix(clawai-tier, server):** Portal 401/403 no longer flips the badge to Free. The portal returns 401/403 for several distinct cases (genuine Free, revoked token, token format migration, backend cleanup), so the response code alone can't distinguish them. Fall through to the existing `unreachable` branch instead — `clawaiAccountTier` preserves `localTier`, the downgrade-celebration popup doesn't fire spuriously, and the next poll re-tries the portal immediately so a recovered token is picked up without a 120 s wait.
- **perf:** Added a separate short negative cache (`portalUnreachableCache`, 30 s TTL — matches the poll cadence) so back-to-back polls during sustained auth failure don't hammer the portal. Successful 200 clears it so recovery stays responsive.
- **tests:** Replaced the obsolete '401/403 = Free' assertion with two new tests covering the preserve-localTier path and the negative-cache hit.

## v3.0.3

- chore: bump version (already merged to beta via earlier PRs).

Note: PR #138 + PR #139 both contain tier-flicker fixes — #139 is the **client-side** preservation in `useClawboxLogin`, #138 is the **server-side** preservation in `/setup-api/ai-models/status`. They're complementary; together they fix the modal-flicker symptom from both ends.

## Test plan

- [ ] Pull `main` on a Jetson, run the standard update flow, verify version reports 3.0.4.
- [ ] Smoke-test the Remote Desktop clipboard (Cyrillic round-trip) — see PR #139 for the full Test plan checklist.
- [ ] Smoke-test ClawKeep pairing → portal hand-off.
- [ ] Confirm the tier modal does NOT pop on a brief network blip or portal 401/403; DOES pop exactly once on a real Paid → Free transition (200 + `tier: "free"`).
- [ ] Verify the codex self-heal: on a Jetson with `@openclaw/codex` installed, manually delete `~/.openclaw/npm/node_modules/@openclaw/codex/node_modules/openclaw`, restart the gateway service, confirm the peer dep is restored and chat works.


GeorgiK77 (approved):
i trust this

Issue #141, opened 2026-05-22
Installer/repair can leave conflicting user openclaw-gateway.service alongside system clawbox-gateway.service


jamesachurchill:
## Summary

ClawBox uses the system service `clawbox-gateway.service` as the real OpenClaw gateway owner, but a pre-existing or later-created user service `openclaw-gateway.service` can still coexist and target the same gateway role/port. When that happens, the two launch models can fight each other or mislead operators into debugging the wrong service.

On my host, the stable state now requires the user unit to be explicitly masked:

```
$ ls -l ~/.config/systemd/user/openclaw-gateway.service
lrwxrwxrwx 1 clawbox clawbox 9 Apr 27 09:44 /home/clawbox/.config/systemd/user/openclaw-gateway.service -> /dev/null

$ XDG_RUNTIME_DIR=/run/user/1000 DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/1000/bus systemctl --user is-enabled openclaw-gateway.service
masked
```

This looks like an install/repair/deconfliction gap rather than an isolated runtime failure.

## Steps to reproduce

1. Start on a Jetson host where a user-level OpenClaw gateway unit exists or gets reintroduced, for example:
   - the machine previously ran standalone OpenClaw, or
   - `openclaw gateway install` / repair tooling was run in the user context.
2. Install or run ClawBox, which enables and starts the system service `clawbox-gateway.service`.
3. Reboot, or restart both the system and user managers.
4. Observe that both gateway launch models now exist on the same machine:
   - system-owned `clawbox-gateway.service`
   - user-owned `openclaw-gateway.service`
5. Depending on ordering and state, one of these happens:
   - port ownership conflict / restart churn
   - misleading failed user-service status while the real listener is the system ClawBox service
   - operator repair flow focusing on the wrong unit

## Expected behavior

ClawBox should detect and deconflict a user-level `openclaw-gateway.service` during install/repair/update flows.

At minimum I would expect one of:

- automatically disable + mask the conflicting user unit
- a loud preflight warning with exact remediation steps
- health/repair tooling that clearly says ClawBox is using `clawbox-gateway.service` and that `openclaw-gateway.service` must not be active on the same host

## Actual behavior

Manual masking of the user unit is required to reach a durable steady state. Without that, the host can end up with two competing gateway service models, and OpenClaw-oriented repair commands can steer debugging toward the wrong service.

## Device info

[...]

PR #145, opened 2026-05-26
fix(install): mask conflicting user-level openclaw-gateway.service

KrasimirKralev:
Fixes #141.

Reported by @jamesachurchill — thanks for the detailed write-up.

## Problem

A user-level `openclaw-gateway.service` (created by standalone OpenClaw, or by older `openclaw gateway install` paths) can coexist with our system-level `clawbox-gateway.service`. Both target port `18789` and both want to be the gateway owner. When that happens:

- Port-ownership conflict / restart churn
- Misleading `failed` status on the user unit while the real listener is the system one
- `openclaw doctor` and other repair tooling steer operators to debug the wrong unit

ClawBox owns `clawbox-gateway.service` as the single source of truth on appliance installs; the user variant has no role and should never be running. The reporter's stable workaround was manual masking — operator knowledge customers shouldn't need.

## Why this happens — and why it's not visible on dev devices

The user-level unit predates ClawBox by years; it was OpenClaw's native deployment shape back when OpenClaw was just a dev-laptop CLI. Once a customer's Jetson has ever run standalone OpenClaw, the unit file sits in `~/.config/systemd/user/` forever — ClawBox's installer never owned that path, so nothing has been removing it. Our internal test devices were all clean-installed straight to ClawBox, so this never surfaced on our side.

## Fix

`step_gateway_setup` (right next to the IPv4-first DNS drop-in we just added in v3.0.5) now:

1. Probes `~/.config/systemd/user/openclaw-gateway.service`
2. If it exists and isn't already a symlink to `/dev/null`, replaces it with `ln -s /dev/null` — the on-disk equivalent of `systemctl --user mask`, no live user session required
3. If it's already masked, leaves it alone (idempotent)

The original file isn't deleted, just neutralised — a developer who wants standalone OpenClaw back can `unmask` it manually. Diagnostics endpoint (a separate suggestion from the issue) deferred to a follow-up.

## Verified live on `clawbox.local` (v3.0.5 Jetson)

Created a fake competing unit, then ran the patched step twice:

| Run | Pre-state | Output | Post-state |
|-----|-----------|--------|-----------|
| 1st | Real file (97 bytes) | `Masking conflicting user-level openclaw-gateway.service` | symlink → `/dev/null` ✓ |
| 2nd | symlink → `/dev/null` | `User-level openclaw-gateway.service already masked` | unchanged ✓ |

## Test plan

- [ ] Pull on a Jetson that has a leftover user-level `openclaw-gateway.service`. Run System Update or `sudo bash install.sh --step gateway_setup`. Confirm the unit is masked, system gateway still serves on `:18789`, no port conflicts.
- [ ] Pull on a clean Jetson (no leftover user unit). Confirm the step exits silently, no errors, no spurious file creation.
- [ ] Run `sudo bash install.sh --step gateway_setup` a second time. Confirm the idempotent "already masked" log line.

## Summary by CodeRabbit

* **Chores**
  * Enhanced installation setup to handle conflicting service configurations more gracefully, ensuring a more reliable setup process.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ID-Robots/clawbox/pull/145?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1438,6 +1438,35 @@ step_recover() {
 step_gateway_setup() {
   cp "$PROJECT_DIR/config/clawbox-gateway.service" /etc/systemd/system/
 
+  # Mask any leftover user-level openclaw-gateway.service. Standalone
+  # OpenClaw (and some older `openclaw gateway install` paths) dropped a
+  # unit file at ~/.config/systemd/user/openclaw-gateway.service that
+  # competes with our system unit for port 18789. When both wake up,
+  # operators get port-ownership conflicts, restart churn, and misleading
+  # failed-status messages on the wrong unit. ClawBox owns
+  # `clawbox-gateway.service` as the single source of truth; the user
+  # variant has no role on an appliance install. Replacing the unit file
+  # with a symlink to /dev/null is the on-disk equivalent of
+  # `systemctl --user mask` — works without needing a live user session.
+  # Idempotent: if it's already masked we leave it alone. Reported via
+  # ID-Robots/clawbox#141.
+  local USER_SYSTEMD_DIR="$CLAWBOX_HOME/.config/systemd/user"
+  local USER_GATEWAY_UNIT="$USER_SYSTEMD_DIR/openclaw-gateway.service"
+  if [ -e "$USER_GATEWAY_UNIT" ] || [ -L "$USER_GATEWAY_UNIT" ]; then
+    local CURRENT_TARGET
+    CURRENT_TARGET=$(readlink "$USER_GATEWAY_UNIT" 2>/dev/null || echo "")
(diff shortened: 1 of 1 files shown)


KrasimirKralev:
Reopening to re-link with the recreated branch after my CodeRabbit-fix push (`b8d62ad`). GitHub kept the PR pinned to the pre-deletion HEAD.

KrasimirKralev:
Superseded by a fresh PR — GitHub keeps this PR pinned to the pre-branch-deletion SHA `7c855c3` and won't re-attach to the recreated branch. Opening a fresh PR from the same branch (now at `b8d62ad`, which includes the CodeRabbit fix for stopping running user-level instances before masking).

PR #146, opened 2026-05-26
fix(setup): harden completion + reconnect, refresh header

KrasimirKralev:
## Summary

Post-v3.0.5 hardening for the setup wizard. Three concerns bundled because they all surface around the late-wizard steps (3/5) and share user-impact (the customer sees an unrecoverable red error in cases where setup actually succeeded or would have succeeded with a brief retry).

### 1. Credentials step (step 3) — reconnect probe
After the hotspot/hostname change, the browser's connection to the device drops while avahi re-broadcasts and the WiFi AP cycles. The previous one-shot 5 s HEAD probe was too aggressive — users saw a red \"Could not reach\" error even when the device came back ~8 s later.

- Progressive backoff: 3 + 5 + 7 + 10 + 12 + 15 s (~52 s total)
- Soft green \"Settings saved. Reconnecting to <new URL>…\" while probing
- Red error only after the full schedule exhausts
- Backoff timer is now cancellable on abort (no setState-after-unmount warning)

### 2. Setup completion (step 5) — retry + friendly error
`POST /setup-api/setup/complete` can fail transiently when the gateway restart from the AI step is still settling, or the browser's connection is briefly degraded after the step 3 hotspot reconfig. Until now this produced a dead-end \"Failed to fetch\" banner with no recovery path.

- 3-attempt retry with [0, 2 s, 4 s] backoff
- 15 s per-attempt `AbortSignal.timeout` so a hung request doesn't block the full backoff window
- Friendly i18n string (`wizard.completionUnreachable`) added for all 10 locales — replaces raw browser error text
- \"Опитай отново\" / Retry button rendered next to the error banner
- `t` hoisted into a ref so a mid-completion locale switch can't re-fire the effect and double-POST

### 3. Header refresh
- Icon swapped to the official openclawhardware.dev brand crab (192×192, square, sharp)
- Version badge restyled inline next to \"ClawBox\" title with a muted slash separator and tabular monospace number — replaces the small green text below the title

## Files

- \`src/components/CredentialsStep.tsx\` — retry loop + cancellable timer
- \`src/components/SetupWizard.tsx\` — `postCompleteWithRetry`, friendly error mapping, retry button, version badge restyle, \`t\` ref
- \`src/lib/translations.ts\` — `wizard.completionUnreachable` for en/bg/de/es/fr/it/ja/nl/sv/zh
- \`public/clawbox-icon.png\` — replaced with brand crab from openclawhardware.dev

## Test plan

[...]

PR #148, opened 2026-05-26
Merge beta into main: v3.0.6 release

### LINKED ISSUE #141: Installer/repair can leave conflicting user openclaw-gateway.service alongside system clawbox-gateway.service

KrasimirKralev:
## v3.0.6 — post-v3.0.5 hardening

Two PRs land in this release, both already merged to beta and CI-green.

### #147 — Mask conflicting user-level `openclaw-gateway.service`
Resolves [#141](https://github.com/ID-Robots/clawbox/issues/141) (@jamesachurchill).

On installs where standalone OpenClaw (or an older `openclaw gateway install`) dropped a unit file at `~/.config/systemd/user/openclaw-gateway.service`, both that user service and ClawBox's system-level `clawbox-gateway.service` tried to bind port 18789 — port-ownership conflicts, restart churn, misleading "failed" status on the wrong unit.

- `install.sh` (`step_gateway_setup`) replaces the user unit with a symlink to `/dev/null` (on-disk equivalent of `systemctl --user mask`)
- Stops the running user instance first (via `sudo -u clawbox systemctl --user stop` with `XDG_RUNTIME_DIR`) so port 18789 frees immediately — otherwise the old process keeps holding the port until next login
- Idempotent: skips if already masked
- Non-fatal on no-session / not-running

### #146 — Setup wizard hardening + visual refresh
Three concerns bundled because they all live in the setup wizard's late steps.

**Credentials step (step 3) — reconnect probe:**
- Progressive backoff (~52 s) replaces the one-shot 5 s HEAD probe — soft "Settings saved. Reconnecting…" toast instead of red error during the typical Jetson recovery window (10–20 s)
- `AbortSignal.any([controller.signal, AbortSignal.timeout(...)])` so the probe cancels on save/unmount AND per-attempt timeout, without poisoning the parent controller
- i18n keys (`credentials.reconnecting`, `credentials.reconnectFailed`) added for 10 locales with `{url}` placeholder
- Cancellable backoff timer (no setState-after-unmount)

**Setup completion (step 5) — retry + friendly error:**
- 3-attempt retry on `POST /setup-api/setup/complete` with `[0, 2 s, 4 s]` backoff — absorbs transient 5xx from a settling gateway restart
- 15 s per-attempt `AbortSignal.timeout` so a hung request doesn't block the full backoff window
- `wizard.completionUnreachable` i18n key replaces raw "Failed to fetch" (10 locales)
- "Опитай отново" / Retry button next to the error banner
- `t` hoisted into a ref so a mid-completion locale switch doesn't re-fire the effect

**Header refresh:**
- Brand crab from openclawhardware.dev replaces `clawbox-icon.png`
- Version badge restyled inline next to title with a muted slash separator

**Test robustness:**
- `e2e-install/80-chat.spec.ts` now dismisses the `TierUpgradeCelebration` modal between fill and Send (upstream ClawAI proxy started reporting Max tier on first dashboard mount in CI)

## Files

```
30 files changed, 1,188 insertions(+), 232 deletions(-)
```

Includes the full v3.0.4 → v3.0.6 changeset as the diff sees it, but the user-visible scope vs main is just #146 + #147.

## Test plan

- [x] PR #147 — landed in beta with all checks green
- [x] PR #146 — full CI green after `ef876e2` (test-robustness fix), CodeRabbit re-reviewed
- [x] End-to-end tested on a real Jetson at 192.168.1.45 — full wizard walk-through, credentials reconnect, completion retry, new header
- [ ] Tag v3.0.6 and cut GitHub release after merge


KrasimirKralev:
@yalexx ping for approval — this is the v3.0.6 promotion bundling PR #146 (setup-wizard hardening) and PR #147 (mask conflicting user-level openclaw-gateway.service). All checks green, both PRs already CodeRabbit-reviewed and merged into beta.

Also @todor943 / @GeorgiK77 if either of you sees this first.

Side-ask while we're here, @yalexx: could you bump KrasimirKralev to **Admin** on this repo (Settings → Collaborators) and add the account to the **bypass actors list** on the main-branch ruleset? Currently every PR sits waiting on a manual CODEOWNERS approval which slows the release cadence quite a bit on the days no one else is around. Happy to discuss tradeoffs if you'd rather not.

yalexx (approved):

PR #164, opened 2026-06-05
feat: show the device IP as the primary access address

KrasimirKralev:
## Summary

On many home networks the access point drops wired→Wi-Fi mDNS multicast, so `<hostname>.local` resolution is unreliable. This surfaces the box's **LAN IP** as the primary "Access this device at" address, with `.local` kept as a best-effort fallback.

## What changed

- **`lib/system-info.ts`** — new `getReachableIpv4()`: picks the IPv4 on the lowest-metric default-route interface (parsed from `/proc/net/route`), skipping loopback and the captive-portal AP subnet (`10.42.0.x`); falls back to the first usable interface. Returns `null` when only the setup AP is up.
- **`/setup-api/system/hostname`** GET now also returns `ipv4`.
- New **`useDeviceAddress()`** hook (`localUrl` + IP-primary `primaryUrl`) replacing `useLocalUrl`.
- **SettingsApp** "Access this device at" card now shows the IP as primary with a `.local` fallback note — replacing the old "use `.local` instead" advice, which was backwards given mDNS is the unreliable part here.
- **WifiStep** "reach me at" note leads with the IP. **DoneStep** got the mechanical hook rename only (it is currently unused/dead code, so its behaviour is unchanged by design).

## Verification

Verified on a Jetson box: `getReachableIpv4()` returns the box's real LAN IP; the hostname API returns `ipv4`; the Settings → Network card renders the IP as the primary address with the `.local` fallback line. Production build + typecheck pass on-device.

## Scope

Address-display only — no change to networking, the OpenClaw gateway, or how the box actually joins the network.

## Summary by CodeRabbit

* **New Features**
  * Device now prefers and shows an IPv4 address as the primary access URL, with mDNS hostname as a fallback.
  * UI copy/link actions and connection messages now use the preferred IP-based access URL.

* **Documentation**
  * On-page guidance updated to explain IPv4 reliability, mDNS caveats, and recommending reserving the device IP in your router.


KrasimirKralev:
@coderabbitai review

PR #176, opened 2026-06-08
fix: don't show the reasoning-effort picker for local Gemma (llamacpp)

KrasimirKralev:
## Problem
Switching the chat's reasoning effort to anything but **Off** while using local **Gemma** (`llamacpp/gemma…`) threw:

> Failed to change effort: thinkingLevel "low" is not supported for llamacpp/gemma4-e2b-it-q4_0 (use off)

Gemma exposes no reasoning-effort control, but `llamacpp` had no entry in the per-provider config, so it fell through to the default `off/low/medium/high` set. The chat then pushed `sessions.patch { thinkingLevel }` to the gateway, which rejected it.

## Fix
- **Extract** the per-provider reasoning config + helpers out of the 1.6k-line `ChatPopup` into a small, testable lib — `src/lib/chat-reasoning.ts` — and declare `llamacpp` as **off-only** (`{ levels: ["off"], default: "off" }`).
- **Hide** the effort dropdown when the active provider offers no real choice (`visibleThinkingLevels.length > 1` is false). A one-option dropdown is pointless, and it keeps the wire value at `off`, which the gateway accepts. This generalizes — any future off-only provider gets the right behavior for free; there's no `provider === "llamacpp"` special-case.
- Cloud providers (OpenAI / Anthropic / Google / OpenRouter / ClawBox AI) are unchanged — they keep their full effort dropdowns.

## Tests
New `src/tests/unit/chat-reasoning.test.ts` — off-only Gemma hides the picker, cloud providers keep their sets, unknown/empty providers fall back, and every config is self-consistent (default ∈ levels, every level labelled).

## Validation
Built and tested on a live device — `bun run build` clean (type-check included), 6/6 `chat-reasoning` tests pass. Verified on the box that Gemma now shows no effort picker and no longer errors.

### DIFF
--- a/src/components/ChatPopup.tsx
+++ b/src/components/ChatPopup.tsx
@@ -14,6 +14,14 @@ import {
 import { useChatToolCalls, ToolCallPills } from '@/lib/chat-tool-events'
 import { FIX_ERROR_EVENT, buildFixErrorPrompt, type FixErrorContext } from '@/lib/ui-events'
 import { isSentinel } from '@/lib/chat-sentinels'
+import {
+  type ThinkingLevel,
+  type ProviderReasoningConfig,
+  THINKING_LEVEL_LABELS,
+  getProviderReasoningConfig,
+  readPersistedThinkingLevel,
+  PERSIST_KEY_PREFIX,
+} from '@/lib/chat-reasoning'
 
 const MASCOT_LINES_KEY = 'clawbox-mascot-convo-lines'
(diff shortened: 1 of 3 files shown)


KrasimirKralev:
@coderabbitai review

KrasimirKralev:
@coderabbitai full review

KrasimirKralev:
@coderabbitai full review

PR #179, opened 2026-06-09
feat: store discovery — sort, official badges, richer skill inspect

KrasimirKralev:
## What

Makes the App Store browsable instead of a flat list:

- **Sort dropdown** — Popular / Top rated / A–Z. ClawHub ignores sort params (verified live), so we sort the *loaded* set client-side; `popular` keeps ClawHub's own order (already ~install-count desc).
- **Official badge** — `channel === "official"` is the only meaningful trust signal (`verified` is true for every listing), shown on cards and in the inspect header.
- **Richer inspect** — pulls the per-skill detail endpoint (`?slug=`) for Featured/Official pills, install count, **Updated {date}**, a **runs code** flag, and a deep-link to the canonical ClawHub page. Falls back to list data if the detail fetch fails.
- **Per-skill detail proxy** — the `/setup-api/apps/store` route now also proxies `?slug=` (validated), via a shared `proxy()` helper that both the list and detail paths reuse.
- **8 i18n keys** added across all 10 locales (parity test passes).

## Notes

- Install count: the detail endpoint's `installsAllTime` is unreliable (returns 0 even for apps with thousands of installs), so we only trust it when positive and otherwise fall back to the list's bucketed "2800+" string — one value feeds both the header and the Downloads stat so they can't disagree.
- The remaining store-API gaps (long-form description, real install count, per-skill required-keys) are tracked upstream in openclaw/clawhub#2570 and ID-Robots/clawbox-website#158 — not in scope here.

Built and verified on the device.

## Summary by CodeRabbit

* **New Features**
  * Official app badges identifying verified, first-party applications
  * Enhanced app detail modal displaying additional metadata including update dates and code execution indicators
  * New sorting options for browsing app lists by popularity, rating, or name
  * Expanded multilingual support for the app store interface


KrasimirKralev:
@coderabbitai review

PR #201, opened 2026-06-15
fix(openai): stop Codex subscription OAuth corrupting the id token

KrasimirKralev:
## The bug
A ChatGPT Plus / Codex **subscription** login fails on every message with:

> Error: failed to set external auth: invalid ID token format

…and re-running the OAuth doesn't help. After it, the OpenAI **API-key** provider also breaks with the same error (the failed OAuth credential "poisons" it), while Claude keeps working — so it's isolated to the Codex OAuth token handling. Reported on ClawBox v3.1.3, Jetson Orin NX.

## Root cause
The OpenAI subscription OAuth flows run an `id_token → openai-api-key` token exchange and store the resulting **`sk-` key** as the codex profile's `access`. `gateway-pre-start` then synthesizes `~/.codex/auth.json` with `id_token: p.id || p.access` → the API-key string, which the Codex app-server rejects as *"invalid ID token format."* Because `~/.codex/auth.json` is shared and **rebuilt on every gateway start**, the bad credential also overrides the API-key OpenAI provider and survives re-auth.

It's account-dependent: boxes where OpenAI's api-key exchange happened to *fail* kept a valid JWT `access` and worked — hence it reproduces on some accounts and not others.

## The fix
Codex (subscription) needs the raw OAuth JWTs, not an API key:
- **`oauth/device-poll`** (device-code flow) and **`oauth/exchange`** (redirect flow): drop the `id_token → api-key` exchange; return `access_token` + `id_token` + `refresh_token`. Both completion paths now behave identically, so the bug isn't left latent on either.
- **`ai-models/configure`**: persist the `id_token` as the profile's `id`, so the synthesis writes a valid JWT id_token (and `access` stays a JWT, which also fixes `account_id` resolution that splits `access` on `.`).
- **`AIModelsStep`**: forward `id_token` from the OAuth result to configure.

`gateway-pre-start`'s synthesis is unchanged — it already prefers `id` over `access`.

## Validation (on a device)
A fresh Codex device-code login now stores a JWT `access` **and** `id` (id_token), synthesizes a valid `~/.codex/auth.json` (id_token JWT ✓, access JWT ✓, account_id ✓), and **OpenAI accepts the credential** — the "invalid ID token format" is replaced by the account's normal rate-limit. Full build green on-device.

## Interim customer workaround (for affected boxes, until this ships)
Clear the corrupted `codex:*` auth profile + `~/.codex/auth.json` and restart the gateway; use Claude / ClawBox AI meanwhile.

## Summary by CodeRabbit

* **New Features**
  * OAuth authentication configuration now supports capturing and storing ID tokens alongside access and refresh tokens for enhanced token management.

* **Improvements**
  * OAuth token exchanges updated to return complete raw JWT token sets, including ID tokens, instead of performing token conversion operations.

### DIFF
--- a/src/app/setup-api/ai-models/configure/route.ts
+++ b/src/app/setup-api/ai-models/configure/route.ts
@@ -383,6 +383,7 @@ export async function POST(request: Request) {
       provider?: string;
       apiKey?: string;
       authMode?: string;
+      idToken?: string;
       refreshToken?: string;
       expiresIn?: number;
       projectId?: string;
@@ -396,7 +397,7 @@ export async function POST(request: Request) {
       return NextResponse.json({ error: "Invalid JSON" }, { status: 400 });
     }
 
-    const { provider, apiKey, authMode = "token", refreshToken, expiresIn, projectId, scope = "primary", model: bodyModel } = body;
+    const { provider, apiKey, authMode = "token", idToken, refreshToken, expiresIn, projectId, scope = "primary", model: bodyModel } = body;
     const requestedClawboxAiTier = normalizeClawboxAiTier(body.clawaiTier);
     const normalizedApiKey = typeof apiKey === "string" ? apiKey.trim() : "";
     const isOllama = provider === "ollama";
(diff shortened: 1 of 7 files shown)


KrasimirKralev:
Addressed CodeRabbit's finding in c03406f: the configure route now rejects a codex (OpenAI subscription) save with a 502 unless the `id_token` (or trimmed `access`) is JWT-shaped — closing the `id → access` synthesis fallback that could recreate the invalid id_token. Also fixed the unit tests that asserted the removed api-key-exchange behavior (they now assert the JWTs incl. `id_token` are returned with no second exchange). Local run: 84/84 green across the OAuth + configure suites.

PR #233, opened 2026-07-01
fix(clawkeep): allow '+' in snapshot names so restore works

yalexx:
## Problem
ClawKeep restore fails for **every** real snapshot with `invalid snapshot name`. Reported by customer **Peter Woodman** (screenshots: restore → 'invalid snapshot name').

## Root cause
Snapshot object names embed the ISO timestamp's **timezone offset**, e.g.
`2026-06-22T19-31-57.761+03-00-openclaw-backup.tar.gz.enc` — note the `+`.
The name validator regex only allowed `[A-Za-z0-9._-]`, so the `+` was rejected in **both** `assertSnapshotName()` and `runRestore()` before anything ran → restore/label/lock/delete all bounce.

## Fix
Add `+` to both validation regexes → `[A-Za-z0-9._+-]`. Existing `+`-named snapshots become restorable; `Z`-suffixed (UTC) names still valid.

## Follow-up (separate)
OpenClaw core `backup create` emits a local-offset timestamp (`+03-00`) instead of UTC `Z` on some devices — worth normalizing to `Z` upstream, but not required for this fix.

## Summary by CodeRabbit

* **Documentation**
  * Cleaned up multiple docs pages so titles are handled consistently, improving page structure and readability.
  * Updated the homepage walkthrough ordering and swapped one walkthrough video to English voice-over for easier access.

* **Bug Fixes**
  * Improved snapshot name handling so filenames with `+` characters are accepted, fixing restore and snapshot-related flows for certain timestamp formats.


yalexx:
Superseded by #234 — this branch was cut from docs/updating-clawbox and carried unrelated docs commits. #234 is the same fix, isolated to a single file off main.

PR #250, opened 2026-07-05
chore(ci): bring ClawReview 🦀 mascot + triage to beta

KrasimirKralev:
Brings both bots to beta so beta-targeting PRs/issues (most of our volume) trigger them — `pull_request_target` and `issues` workflows run from the **base branch**.

Carries the same code as the main-side mascot PR (#252):
- **New:** `scripts/pr-review.mjs` + `.github/workflows/pr-review.yml` — the ClawReview mascot (informative orientation, not a code reviewer)
- **Updated:** `scripts/issue-triage.mjs` + `.github/workflows/issue-triage.yml` — same crab mascot voice + OAuth subscription transport (SDK lazy-loaded)

CI-only files, no device code. `CONTRIBUTING.md` not synced (read from the default branch only).

### DIFF
--- a/.github/workflows/issue-triage.yml
+++ b/.github/workflows/issue-triage.yml
@@ -44,10 +44,35 @@ jobs:
         # downloads ≈ minutes per issue, plus flake risk on this bun-managed
         # tree). Scoped to scripts/, only the SDK installs (~5 s); ESM
         # resolution finds scripts/node_modules from the script's own path.
-        run: npm install --prefix scripts --no-save @anthropic-ai/sdk@0.106.0
+        # Keep the versions in sync with pr-review.yml. Two transports, one
+        # wins at runtime (script picks by env): CLAUDE_CODE_OAUTH_TOKEN ->
+        # Claude Code CLI (subscription/OAuth, preferred); else the SDK.
+        env:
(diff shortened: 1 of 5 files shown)


KrasimirKralev:
Thanks @coderabbitai — all 4 addressed in the latest push (identical fix on the main-side PR #252):

1. **`issue-triage.yml` app-token `if`** ✅ — real bug. `HAS_APP` hoisted to **job-level `env`** (a step can't read env it sets on itself), and the minted token is now **scoped to `issues: write`** only.
2. **`pr-review.yml` app-token `if`** ✅ — same fix; token scoped to `pull-requests: write` + `issues: write`.
3. **Duplicated backend logic** ✅ — extracted `parseModelJson` + both transports into **`scripts/lib/ai-backend.mjs`** (`callClaude`); both bots import it, so the transport stays in sync.
4. **Static `@anthropic-ai/sdk` import** ✅ *(already resolved)* — it was made lazy earlier; after the extraction the SDK's sole lazy import lives in the shared backend and `pr-review.mjs` no longer imports it at all.

Validated end-to-end via the OAuth CLI transport with `node_modules` removed — both bots produce their output through the shared backend.

PR #251, opened 2026-07-05
fix(ci): lazy-load Anthropic SDK so the bots' OAuth path needs no SDK

KrasimirKralev:
**Hotfix for the live bots on main.** The OAuth transport (#248) installs only the Claude Code CLI, not `@anthropic-ai/sdk` — but both `scripts/pr-review.mjs` and `scripts/issue-triage.mjs` had a static top-level `import Anthropic from "@anthropic-ai/sdk"`, evaluated at module load regardless of which transport runs. Result: the `review` / `triage` job crashes with `ERR_MODULE_NOT_FOUND` before doing anything.

It slipped through because my local test machine still had the SDK installed from an earlier run; CI (OAuth-only install) exposed it — caught on the #250 CI run.

**Fix:** move the import to a dynamic `await import("@anthropic-ai/sdk")` inside `reviewViaSdk` / `classifyViaSdk`, reached only on the API-key fallback path. The OAuth/CLI path never touches it.

**Verified** by running the OAuth path locally with `scripts/node_modules` removed — full review produced, no SDK present. The identical fix rides to beta in #250.

### DIFF
--- a/scripts/issue-triage.mjs
+++ b/scripts/issue-triage.mjs
@@ -4,7 +4,9 @@
 // Needs: ANTHROPIC_API_KEY (repo secret) and GH_TOKEN (the workflow's GITHUB_TOKEN).
 import fs from "node:fs";
 import { execFileSync } from "node:child_process";
-import Anthropic from "@anthropic-ai/sdk";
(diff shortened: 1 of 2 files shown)


KrasimirKralev:
Superseded by the mascot reshape PR, which includes this exact SDK lazy-load fix as its first commit.

Issue #231, opened 2026-06-26
Harden VNC UI setup API calls against expired auth redirects


jamesachurchill:
### Summary

The VNC setup UI should handle expired setup sessions cleanly when API calls return `401` or an HTML login redirect instead of the expected JSON payload.

### Problem

Several VNC UI actions expect JSON from setup API endpoints:

- VNC availability probing;
- VNC install/repair;
- reboot request after repair;
- VNC clipboard read/write.

If the setup session expires, these calls can receive a `401` or a redirect to `/login`. Without explicit handling, the UI may attempt to parse the wrong response shape, surface a confusing error, or silently fail.

### Proposed shape

- Send `Accept: application/json` on VNC setup API calls.
- Detect `401` responses.
- Detect redirected responses whose final path is `/login`.
- Show a clear expired-login message asking the user to refresh/sign in again.
- Avoid continuing repair/reboot/poll loops after auth expiry is detected.

### Acceptance criteria

- VNC availability check surfaces a clear login-expired state on `401` or login redirect.
- VNC repair action stops and shows a clear error on auth expiry.
- Reboot request after repair stops and shows a clear error on auth expiry.
- VNC clipboard read/write does not try to parse a login page as JSON.
- Unit or component tests cover at least one `401` path and one redirected-login path.

jamesachurchill:
## Review

Endorsed — the proposed shape is correct, and it's cheaper than it looks because the middleware already does the hard half. Findings below, then implementation recommendations.

### Findings

**1. Middleware already branches on `Accept` — the header is the keystone change** `[repo-grounded]`
`src/middleware.ts` step 5: requests whose `Accept` includes `application/json` get `{ error: "Authentication required" }` with status 401; everything else gets `NextResponse.redirect(/login?redirect=…)`. None of the VNC UI fetches currently set `Accept`, so they all fall into the redirect branch. Adding `Accept: application/json` flips every call site into a clean, machine-detectable 401 with no server changes.

**2. Current failure modes split by HTTP method** `[repo-grounded]`
- **GETs** (`checkVnc`, clipboard read, reboot poll): the browser follows the redirect to `/login`, which returns **200 HTML**. `res.ok` is true, so `res.json()` either throws (`checkVnc` surfaces `Unexpected token '<'…` as the error text) or is swallowed by `.catch(() => ({}))` (clipboard read fails silently — exactly the issue's fourth bullet).
[...]

PR #253, opened 2026-07-05
fix(vnc): handle expired setup sessions on VNC UI API calls

KrasimirKralev:
Closes #231.

## Problem
The VNC panel's `/setup-api/*` calls assume a JSON body. On an expired session, `src/middleware.ts` (250-258) returns **401** to JSON clients or **307-redirects to `/login`** otherwise. Since `fetch` follows the redirect, `res.json()` was handed a login **HTML page** → a confusing `Unexpected token <` error (the `checkVnc` availability probe), or — in the post-repair reboot poll — the loop spun to its 5-minute timeout instead of stopping.

## Fix
- **New `src/lib/setup-auth.ts`** — `JSON_ACCEPT_HEADERS` + `isAuthExpired(res)` (true on `401`, or a followed redirect whose final path is `/login`).
- **All six VNC setup-api calls** now send `Accept: application/json` (so middleware answers with a clean 401 instead of a redirect) and check `isAuthExpired` before parsing: availability probe, install/repair, post-repair reboot, the **reboot poll loop (now halts instead of timing out)**, and clipboard read + write.
- On expiry the error panel shows a clear **"session expired → Sign in again"** action instead of retry/repair, and the clipboard read never parses a login page as JSON.

## Tests
`src/tests/unit/setup-auth.test.ts` covers the **401 path** and the **redirected-to-`/login` path** (plus active-session and non-login-redirect negatives) — the two paths the issue's acceptance criteria call for.

## Acceptance criteria
- [x] Availability check surfaces a clear login-expired state on 401 or login redirect
- [x] Repair action stops and shows a clear error on auth expiry
- [x] Reboot request after repair stops and shows a clear error on auth expiry
- [x] Clipboard read/write does not parse a login page as JSON
- [x] Unit tests cover a 401 path and a redirected-login path


yalexx:
Closing during PR-queue cleanup (per Karchev): stale (07-20) and conflicting/dirty against beta. Branch is preserved — reopen or rebase if this work is still wanted.

Issue #193, opened 2026-06-10
install-x64.sh: unguarded openclaw config set calls can abort steps on ConfigMutationConflictError


KrasimirKralev:
Same bug class as #192 fixed for the Jetson installer, but in the x64 path:

- `install-x64.sh:282` — `as_user "$OPENCLAW_BIN" config set gateway.controlUi.allowInsecureAuth true --json` runs under `set -euo pipefail` with no retry and no `|| true`. A `ConfigMutationConflictError` (OpenClaw 2026.6.x optimistic concurrency — triggered by the live gateway or the CLI's own state migrations) aborts the step mid-update.
- Lines ~354/369/371/372 take the opposite shortcut (`2>/dev/null || true`) — conflicts there are silently swallowed and the config is simply not applied.

Note x64 has no `gateway-pre-start.sh`, so the auth/controlUi writes can't be deferred the way #192 did for Jetson — they need the retry treatment (port `oc_config_set` from install.sh), and the silently-swallowed ones should at least log.

Found during the #192 review.

KrasimirKralev:
Triage: **bug**, same class as the already-fixed #192 (Jetson installer) but in the x64 path — unguarded `openclaw config set` under `set -euo pipefail` aborts on `ConfigMutationConflictError`. Well-scoped: guard the ~5 call sites with the same retry helper #192 introduced.

KrasimirKralev:
Deferring this one for now — it's a real edge case (unguarded `openclaw config set` calls in install-x64.sh can abort a step on `ConfigMutationConflictError`), but it's lower-priority than the security and auth-hardening work in flight. Closing to keep the backlog focused; we'll pick it back up in a dedicated install-hardening pass.

PR #254, opened 2026-07-05
fix(install-x64): guard openclaw config set against ConfigMutationConflictError

KrasimirKralev:
Closes #193.

Same bug class as #192 (fixed for the Jetson installer), now in the x64 path.

## Problem
`install-x64.sh:282` ran `config set gateway.controlUi.allowInsecureAuth true --json` under `set -euo pipefail` with **no retry** — a `ConfigMutationConflictError` (OpenClaw 2026.6.x optimistic concurrency, triggered by the live gateway or the CLI's own state migrations) aborts the install step mid-update. Four other calls (`:354/369/371/372`) took the opposite shortcut (`2>/dev/null || true`) — conflicts there were **silently swallowed** and the config just wasn't applied.

## Fix
x64 has no `gateway-pre-start.sh` to defer these writes to, so the fix is the **retry treatment**: ported `install.sh`'s `oc_config_set` (adapted to x64's `as_user`), and routed all five config-set calls through it.

- `:282` now retries up to 3× and only aborts on a **genuine persistent failure** — keeps its fail-loud intent, minus the transient-conflict abort.
- The four `|| true` calls now **retry and log** on persistent failure (still non-fatal) instead of silently dropping config with no trace.

## Validation
`bash -n` clean. (No local node/bun on the dev box — installer behavior verified on device/CI per project convention.)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved installer reliability when applying configuration, reducing failures caused by temporary update conflicts.
  * Configuration changes are now retried automatically instead of being silently skipped on error.
  * Setup now more consistently applies authentication-related and access-related settings during installation.


yalexx:
Closing during PR-queue cleanup (per Karchev): stale (07-20) and conflicting/dirty against beta. Branch is preserved — reopen or rebase if this work is still wanted.

PR #275, opened 2026-07-27
fix(codex,gemma): unblock GPT-5.6 on ChatGPT auth and stop Gemma provisioning from hanging

yalexx:
Two independent field bugs, both found while validating 3.1.11 on a real box. Targets `beta` — nothing here goes to `main`.

## 1. GPT-5.6 Sol/Terra/Luna unusable on ChatGPT-subscription boxes

`gateway-pre-start.sh` rewrites `openai/<gpt>` → `codex/<gpt>` on devices with Codex OAuth and no OpenAI API key, so a stored keyless `openai/*` model doesn't 401 with *"Missing bearer or basic authentication in header"*. Its `_CODEX_SUPPORTED` tuple is documented in-line as a mirror of `CODEX_SUPPORTED_MODEL_RE` — but the regex learned `gpt-5.6-{sol,terra,luna}` in #271 and the tuple never did.

Net effect: a box whose stored model was `openai/gpt-5.6-sol` was **never migrated**, kept resolving to `api.openai.com` with no key, and 401'd. The newest models were broken on exactly the auth mode they're sold with.

- Adds the gpt-5.6 family to the tuple.
- New `gateway-pre-start-codex-models.test.ts` pins the mirror against **both** `CODEX_SUPPORTED_MODEL_RE` and the `CODEX_MODELS` picker catalog, so "offerable but not migratable" can't come back. Verified it fails against the old tuple.
- `-pro` variants stay excluded — they 400 on the ChatGPT-account path, so migrating them would swap a working keyed route for a broken one.

## 2. "Provisioning offline Gemma 4" hangs, then dies

On a cold box that step builds llama.cpp from source with CUDA on a 6-core Jetson Orin **and** downloads a multi-GB GGUF. Two problems compounded:

- `clawbox-root-update@.service` capped `TimeoutStartSec` at **1800s**, so systemd killed the build partway through. Raised to 7200s — a ceiling, not a wait; every other root-update step finishes in seconds.
- The route ran it as a single blocking `systemctl start` that emitted **nothing** for the whole install. The wizard sat on a bare spinner, and a real hang looked identical to normal progress.

Now: `systemctl start --no-block` plus our own systemd polling, streaming journal lines to the wizard and failing the moment the unit reports `failed` instead of waiting out the timeout. Lines carry a phase-stable prefix (raw cmake/hf/apt output matches no phase keyword and would bounce the wizard back to step 1); the tail is surfaced as overlay detail.

## Verification

- 1444 unit/component tests pass (120 files)
- `tsc --noEmit` clean; `eslint` clean on changed files
- Migration mapping checked standalone: `openai/gpt-5.6-sol → codex/gpt-5.6-sol`, `openai/gpt-5.4-pro → None`
- New test proven to fail without the `_CODEX_SUPPORTED` fix

## Not covered here

The `refresh_token_reused` 401 is a **separate** issue and is not fixed by this PR. Two candidate causes: the same ChatGPT account signed in on several boxes (sufficient on its own — rotating refresh tokens are single-use), and pre-start mirroring the same `refresh_token` into every agent's `codex-home/auth.json`, giving one token family multiple independent refreshers. Needs its own change.

## Summary by CodeRabbit

- **New Features**
  - Added support for additional Codex GPT-5.6 model variants during model migration.
  - Offline Gemma 4 provisioning now streams live progress updates with more informative status details.

- **Bug Fixes**
  - Increased systemd and offline provisioning time limits to prevent premature termination during long builds/downloads.
  - Improved installation failure reporting by surfacing the latest observed error context.
  - Ensures updated system service/unit files are refreshed on in-app update flows.

### DIFF
--- a/config/clawbox-root-update@.service
+++ b/config/clawbox-root-update@.service
@@ -5,4 +5,9 @@ Description=ClawBox Root Update Step (%i)
 Type=oneshot
 EnvironmentFile=-/etc/clawbox/network.env
 ExecStart=/bin/bash /home/clawbox/clawbox/install.sh --step %i
-TimeoutStartSec=1800
+# 30 min was not enough for llamacpp_install on a cold box: that step builds
+# llama.cpp from source with CUDA on a 6-core Jetson Orin AND downloads the
+# multi-GB Gemma 4 GGUF. systemd killed the unit mid-build, so "Provisioning
+# offline Gemma 4" hung until it failed. Other steps finish in seconds — this
+# is a ceiling, not a wait, so raising it costs nothing.
+TimeoutStartSec=7200
--- a/install.sh
+++ b/install.sh
@@ -1352,6 +1352,16 @@ step_post_update() {
   # unit + autocutsel package. Devices installed before the display-:99 move
   # and the clipboard-sync addition get both here without needing a reinstall.
   step_vnc_refresh || echo "  Warning: vnc_refresh step failed (non-fatal)"
+  # Reinstall the unit files from config/ + daemon-reload. Without this, unit
+  # changes only ever reached FRESH installs: the in-app update runs
+  # bootstrap_updater -> ... -> post_update and never re-copies
+  # /etc/systemd/system, so an updated box kept running whatever unit file it
+  # was born with. That silently swallowed the llamacpp_install
+  # TimeoutStartSec raise (30 min -> 2 h) that stops "Provisioning offline
+  # Gemma 4" from being killed mid-build, and would swallow any future unit or
+  # sudoers change the same way. The step is idempotent — cp, daemon-reload,
+  # enable — and is exactly what fresh installs already run.
+  step_systemd_services || echo "  Warning: systemd_services step failed (non-fatal)"
   # Refresh the device-side ClawKeep CLI from the repo. The Python package
   # has the same version string ("0.1.0") across releases, so a plain
   # `pip install` is a no-op even after restore/scheduler bug fixes land —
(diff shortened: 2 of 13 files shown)


bram673a3a1a6f1170a8c6b148ab-ctrl:
## User guidance — Gemma provisioning hotfix

This fix is present on `beta` (the repository’s current default branch). It is for **Gemma 4** provisioning — not Gemma 3 — and it applies only **after the device has completed an in-app update containing this change**.

For an affected device:

1. Complete the normal ClawBox update first.
2. Then return to Local AI and run the normal **Enable / Switch to Gemma 4** action again.
3. Keep the setup view open to follow the live build/download detail. A cold install can take a while; it is no longer limited to the former 30-minute systemd ceiling.
4. If it fails, use the final error shown in the setup view when contacting support; it now includes the latest provisioning/journal detail.

Important boundaries:

- An installation already running under the pre-fix software does **not** gain the new timeout retroactively. If it has stalled or failed, update first, then retry.
- The update itself re-installs the systemd unit via `step_post_update -> step_systemd_services`; that delivery step is what gives existing devices the new two-hour provisioning ceiling.
- Do not describe the fix as already protecting every shipped device or as a stable/main release: this PR explicitly targets `beta`.

Verified against the current `beta` source:
- `config/clawbox-root-update@.service`: `TimeoutStartSec=7200`
- `src/app/setup-api/llamacpp/install/route.ts`: detached systemd start, live journal polling, immediate failed-unit detection, two-hour UI-side ceiling
- `install.sh`: in-app `post_update` re-deploys systemd units.

PR #282, opened 2026-07-28
Merge main into beta — reconverge history after the v3.1.11 promotion

yalexx:
**Content-neutral.** `main` and `beta` already have identical trees, so beta's tree does not change — this only records the ancestry.

### Why

The v3.1.11 promotion (#281) landed on `main` as a **single-parent** commit: `MERGE_HEAD` was dropped while creating the promotion branch, so the merge was committed without beta as a second parent. My mistake, caught right after tagging.

The release itself is unaffected — `v3.1.11` points at a tree byte-identical to `beta`, and the updater has no ancestry guard, so every box resolves the tag correctly.

What it *would* break is the next promotion: with beta not an ancestor of main, `beta → main` would re-conflict on the same files, for exactly the reason #264 caused today's conflicts. This merge stops that from recurring.

### Verification

- `git diff origin/beta` after the merge: empty
- beta's tree is unchanged; only history changes

### DIFF
--- a/.github/workflows/docs-deploy.yml
+++ b/.github/workflows/docs-deploy.yml
@@ -1,7 +1,7 @@
 name: Deploy ClawBox Docs
 
 # Rebuilds the Mintlify docs in docs-site/ and publishes the static export
-# to the gh-pages branch, which GitHub Pages serves at docs.clawbox.tech.
+# to the gh-pages branch, which GitHub Pages serves at docs.clawbox.com.
 
 on:
   push:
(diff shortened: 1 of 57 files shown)


KrasimirKralev:
Closing — the main→beta reconverge was applied directly to beta (merge commit a3e88e2): main's history (the v3.1.10/v3.1.11 promote commits) is now in beta with zero file changes, so git describe origin/beta resolves to v3.1.11 and future promotions line up. This branch predated the recent beta work (#303) and is stale.

Issue #232, opened 2026-06-26
Support configurable trusted setup/control UI origins for remote access


jamesachurchill:
### Summary

ClawBox should provide a supported way to add operator-approved remote setup/control UI origins, such as VPN or MagicDNS hostnames, without hard-coding local hostnames in generated gateway config.

### Problem

The setup/control UI needs a strict allowed-origin boundary, but real deployments may legitimately access the device through more than the default LAN names/IPs. Examples include VPN overlay DNS names, stable local DNS names, or deployment-specific HTTPS origins.

Without a supported configuration hook, operators are pushed toward local patches to generated or pre-start-managed allowed-origin lists. Those local patches are fragile across updates and can drift from upstream behavior.

### Proposed shape

- Add a documented config key for extra trusted setup/control UI origins.
- Validate entries as full origins, not arbitrary strings.
- Merge configured extra origins with the existing default local/LAN origins.
- Keep defaults conservative; do not add broad wildcard origins.
- Surface invalid configured origins clearly during pre-start or setup status checks.

### Acceptance criteria

- Operators can add a specific trusted origin through supported config.
- Invalid origins are rejected or reported clearly.
- Generated gateway/control UI config includes configured extra origins.
- The implementation does not require source patches for deployment-specific hostnames.
- Tests cover valid origin, invalid origin, and default behavior.

### Security note

This should remain an explicit allowlist. The goal is to avoid brittle local patches, not to loosen cross-origin restrictions.

jamesachurchill:
## Review: configurable trusted setup/control UI origins

Endorse the direction. Repo review confirms the problem and surfaces two findings that should shape the implementation.

### Findings

**1. The clobber mechanism is the pre-start replace semantics — the config hook must land there** `[repo-grounded]`

There are already two divergent writers of `controlUi.allowedOrigins`:

- `setControlUiAllowedOrigins()` (`src/lib/openclaw-config.ts:537`) **merges** — it explicitly preserves existing entries when a hostname rename occurs.
- `scripts/gateway-pre-start.sh` (~L226) **replaces** — it compares the on-disk list as a set against its computed list (hostname.local + localhost + 127.0.0.1 + hotspot IPs + enumerated LAN IPs) and overwrites on any difference.

[...]

PR #255, opened 2026-07-05
feat(gateway): configurable extra trusted setup/control UI origins

KrasimirKralev:
Closes #232.

## Problem
`gateway-pre-start.sh` **replaces** `controlUi.allowedOrigins` with a computed list (defaults + this host's LAN IPs) on every boot. A deployment reached through a VPN overlay name, MagicDNS, or a stable HTTPS origin therefore had to **hand-patch the generated config** — fragile, and it drifts across updates.

## Fix
A supported config hook: **`controlUi.extraAllowedOrigins`** in `~/.openclaw/openclaw.json`.

- **New `scripts/gateway_origins.py`** (importable → unit-testable):
  - `normalize_origin(v)` — validates a value as a **full `http(s)://host[:port]` origin**, rejecting bare hostnames, wildcards, paths/queries/fragments, and credentials so a configured entry can never quietly widen cross-origin access.
  - `merge_extra_origins(defaults, configured)` — unions validated extras into the defaults, drops duplicates, and returns the invalid entries for reporting.
- **`gateway-pre-start.sh`** reads `controlUi.extraAllowedOrigins`, merges validated entries into the computed `allowedOrigins`, and logs `WARN: ignoring invalid …` for each rejected one. Missing/`None` key ⇒ defaults only (unchanged behavior). Import falls back to defaults-only if the module is ever unavailable — never blocks boot.

## Usage
```jsonc
// ~/.openclaw/openclaw.json
{ "gateway": { "controlUi": { "extraAllowedOrigins": ["https://box.example.ts.net"] } } }
```
On the next gateway restart, pre-start validates and merges it into the active allowlist. Invalid entries are dropped with a logged warning.

## Tests
`src/tests/unit/gateway-origins.test.ts` drives the module via `python3` (the `test` job runs on ubuntu-latest): the module self-test plus explicit **valid**, **invalid**, and **default** cases — the three the acceptance criteria call for.

## Security
Stays an **explicit allowlist** — no wildcards, defaults conservative. The goal is to kill brittle local patches, not loosen CORS.

## Acceptance criteria
- [x] Operators can add a specific trusted origin through supported config
- [x] Invalid origins are rejected / reported clearly (pre-start `WARN`)
- [x] Generated config includes configured extra origins
- [x] No source patches required for deployment-specific hostnames
- [x] Tests cover valid, invalid, and default behavior

Follow-up: a user-facing docs-site guide entry will go as a separate PR to `main` (docs deploy from the default branch, per repo convention).

PR #295, opened 2026-08-04
Support trusted control UI origins

jamesachurchill:
## Summary

- add a supported `data/control-ui-origins.json` allowlist for genuine cross-origin or custom-origin deployments
- strictly validate exact HTTP(S) origins and merge them with generated gateway defaults
- apply the same exact scheme, host, and port allowlist to setup-redirect reflection
- preserve ordinary same-origin local, private-IP, and MagicDNS behavior without requiring configuration

## Why this remains necessary

Current OpenClaw already accepts ordinary same-origin private and MagicDNS access, so that broad part of #232 no longer needs a workaround. Genuine cross-origin deployments still require an explicit allowlist, and the pre-start script still replaces generated `allowedOrigins` on every boot.

This replaces the approach in #255 rather than reviving it unchanged. The prior `controlUi.extraAllowedOrigins` key is not part of the current strict OpenClaw schema, and #255 did not update the setup proxy's independent redirect-origin boundary.

## Security and update behavior

- exact HTTP(S) origins only; no wildcards, credentials, paths, queries, or fragments
- malformed entries are ignored with pre-start/proxy warnings and never prevent boot
- configured origins match scheme, hostname, and non-default port exactly
- the operator file is under ignored `data/`, so the updater's hard reset preserves it
- missing configuration retains current defaults

## Verification

- current `beta` base: 1,567 tests passed across 128 files
- focused origin coverage: 68 tests passed
- targeted ESLint, shell syntax, Python compile, diff, and publication scans passed
- production Next.js build and TypeScript checks passed
- current `main` has the same affected source/dependency surface as `beta`; the patch was also verified against its current v3.1.11 state during preparation
- both branches pin OpenClaw 2026.7.1; no OpenClaw runtime change is required because this is a configuration-generation and setup-proxy integration

The repository-wide lint command still reports its existing unrelated beta errors; none are in the changed files.

Closes #232

## Summary by CodeRabbit

* **New Features**
  * Added configuration for trusted Control UI origins through a JSON file and environment variable.
  * Supports exact origin matching by scheme, host, and port for setup redirects.
  * Preserves configured origin order and removes duplicates.

* **Bug Fixes**
  * Invalid origin entries are safely ignored with warnings.
  * Setup redirects now preserve the original host when using trusted origins.

* **Documentation**
  * Added configuration and validation guidance to the README.

* **Tests**
  * Added comprehensive coverage for origin validation, loading, merging, and redirect behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,7 +249,7 @@ bun run test             # Unit tests (Vitest)
 | `SESSION_SECRET` | Auto-generated | Session cookie signing key |
 | `OLLAMA_HOST` | `http://127.0.0.1:11434` | Ollama server URL |
 | `CLAWBOX_ROOT` | `/home/clawbox/clawbox` | Project root directory |
-| `CLAWBOX_CONTROL_UI_ORIGINS_FILE` | `data/control-ui-origins.json` | Extra trusted control UI origins (see below) |
+| `CLAWBOX_CONTROL_UI_ORIGINS_FILE` | `/home/clawbox/clawbox/data/control-ui-origins.json` | Extra trusted control UI origins (see below) |
 
 Additional options (OAuth client IDs, ClawBox AI, llama.cpp tuning) live in `.env.example`.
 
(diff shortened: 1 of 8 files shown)


KrasimirKralev:
Closing — merged to beta as #313 (your commits, authorship preserved), run through our full flow. /code-review + CodeRabbit surfaced and we fixed 4 real issues: a LAN/SoftAP reflection regression (a configured origin overlapping a default host stripped its broad reflection), two TS-vs-Python validator divergences (IPv4 shorthand + empty userinfo), and a malformed-Host-port 500. Verified on the Jetson (97 tests). Thanks @jamesachurchill — nice feature.

Issue #284, opened 2026-07-31
clawbox-gateway.service restart limiter does not stop permanent startup failures


jamesachurchill:
### What happened?

`clawbox-gateway.service` can restart indefinitely after a permanent OpenClaw startup/config failure.

The current unit uses:

```ini
Restart=always
RestartSec=5
```

On the affected ClawBox, the effective systemd defaults were 5 starts in a 10-second interval. A failed OpenClaw startup plus the 5-second restart delay took longer than that interval, so the limiter never accumulated enough starts to trip. The journal reached at least restart counter 84 for the same deterministic error.

## Failure trigger

OpenClaw cold start rejected an exec SecretRef command because the configured executable path was a symlink. The application correctly failed closed, but the supervisor treated every identical permanent failure as restartable.

Both Telegram providers remained unavailable throughout the loop.

## Expected behaviour

ClawBox should bound repeated gateway startup failures over a window longer than a normal failed startup cycle. After the breaker trips:

- the unit remains failed and diagnosable instead of looping;
- the last error and breaker state are visible in the setup/control surface;
- one actionable alert is emitted without duplicates;
- an operator can repair config and explicitly restart/reset-failed;
- normal one-off maintenance restarts remain reliable.

## Suggested unit-level direction

Use explicit, tested rate-limit values in the canonical unit source, for example a burst of a few failures over several minutes rather than inherited 5-in-10-second defaults. The exact values should be based on worst-case legitimate pre-start duration.

Also consider whether `Restart=on-failure` better expresses intent, although it does not replace a real rate limiter.

## Acceptance criteria

- A fixture that exits with the same permanent startup error trips the breaker after the configured burst.
- The breaker survives install/update unit regeneration.
- A single approved restart succeeds normally.
- A transient first failure followed by success recovers.
- The control surface reports breaker activation and the final startup error.
- Tests use a failure duration longer than the old 10-second limit window so the historical false protection is covered.

## Related work

PRs #263, #265, and #271 improve update recovery and make optional pre-start work non-fatal. This report concerns a separate permanent OpenClaw runtime/config failure after pre-start completes.

### ClawBox version

ClawBox main at commit `8cb6a43`

### OpenClaw version

OpenClaw 2026.7.1

### Jetson model / JetPack

Ubuntu 22.04 LTS, AArch64

### Logs and screenshots

```shell

```

KrasimirKralev:
[...]

PR #297, opened 2026-08-04
fix(gateway): bound permanent startup failures

jamesachurchill:
## Summary

- give the canonical gateway unit an explicit five-start, one-hour breaker and use `Restart=on-failure` so clean operator stops stay stopped
- detect systemd's persisted `start-limit-hit` result and show one actionable alert on the authenticated gateway-offline control page
- surface the final useful startup error with control-character cleanup, length bounds, secret redaction, and HTML escaping
- cover canonical install/update regeneration, a permanent 12-second failure cycle, transient recovery, systemd parsing, privacy, and offline-page behavior

## Breaker tuning

The existing `TimeoutStartSec=600` plus `RestartSec=5` permits a worst-case failed cycle of 605 seconds. A 3,600-second interval contains five complete cycles (3,025 seconds), so the sixth start is refused even when every failure lasts much longer than the inherited 10-second window. Keeping a burst of five also leaves reasonable headroom for intentional maintenance restarts.

## ClawReview advice

This follows ClawReview's recommended direction—an explicit longer rate limit plus operator-visible breaker reporting—with two deliberate refinements:

- the suggested five-minute example is shorter than one permitted 600-second startup, so it could age out an attempt before the next worst-case failure starts; one hour reliably contains all five cycles
- detailed state and the final error are shown on the authenticated gateway-offline control page rather than adding systemd/journal work to the public, frequently polled setup-status endpoint

## Verification

- focused Vitest suite: 29 tests passed
- changed-file ESLint passed
- `systemd-analyze verify` accepted the changed gateway unit; reported warnings were from unrelated existing host units
- production `npm run build` and its TypeScript phase passed on the exact head
- `git diff --check` passed

No live service was mutated during testing. The deterministic unit fixture uses 12-second failures to cover the historical case where each failed cycle outlasted the old 10-second limit window.

Closes #284

AI-assisted; reviewed and validated before submission.

## Summary by CodeRabbit

* **New Features**
  * Added gateway health monitoring with restart status, failure details, and recovery guidance.
  * Offline gateway pages now display actionable startup errors and restart-breaker instructions.
  * Sensitive credentials and tokens are concealed in displayed diagnostic information.

* **Bug Fixes**
  * Improved gateway restart behavior by limiting repeated failures and restarting only when failures occur.

* **Tests**
  * Added coverage for gateway health reporting, error handling, security sanitization, and installation updates.

### DIFF
--- a/config/clawbox-gateway.service
+++ b/config/clawbox-gateway.service
@@ -2,6 +2,12 @@
 Description=ClawBox OpenClaw Gateway
 After=network-online.target
 Wants=network-online.target
+# A cold Jetson can legitimately spend up to TimeoutStartSec in
+# gateway-pre-start.sh. Keep the limiter window long enough to contain several
+# complete slow failures; inherited 5-in-10s defaults never tripped when one
+# failed cycle already took longer than ten seconds (issue #284).
+StartLimitIntervalSec=3600
+StartLimitBurst=5
 
 [Service]
 Type=simple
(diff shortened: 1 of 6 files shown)


KrasimirKralev:
Closing — merged to beta as #308 (your commits, authorship preserved), run through our full flow. /code-review surfaced 9 findings (incl. a LAN-visible secret leak in the offline-page diagnostics + an install.sh reset-failed gap that would strand the gateway on update after the breaker trips); all fixed or adjudicated, Jetson-verified (19 tests, install.sh valid, journal readable). The core breaker fix was live-validated on the box earlier. Thanks @jamesachurchill — solid fix for a real bug.

Issue #285, opened 2026-07-31
force-update.sh reports completion after setup-app bootstrap without handing off to the full updater


jamesachurchill:
## Summary

The documented emergency command for `force-update.sh` hard-syncs and rebuilds the ClawBox setup app, then reports `Done`, but it does not run or clearly require the normal full updater/post-update recovery path.

That distinction matters during gateway recovery: an operator can reasonably interpret the final success message as a completed ClawBox/OpenClaw update even though only the updater UI was bootstrapped.

## Observed behaviour

Running the documented command:

```bash
bash <(curl -fsSL https://raw.githubusercontent.com/id-robots/clawbox/main/scripts/force-update.sh)
```

performed a hard fetch/reset/clean of the ClawBox checkout, rebuilt the Next.js app, restarted `clawbox-setup`, and printed `Done` when that service became active.

It did not:

- run the normal full updater workflow;
- update OpenClaw;
- perform gateway verification/channel smoke checks;
- restore supported local post-update extensions after `git clean -fd`;
- explain that a second updater step was still required.

In the observed recovery, neither OpenClaw nor ClawBox advanced because the checkout was already at the target commit. The hard clean removed locally maintained post-update extensions, and follow-up smoke checks failed until they were restored.

## Expected behaviour

Keep the hard reset as an explicit recovery tool, but make its bootstrap-only contract unmistakable.

Preferred options:

1. After `clawbox-setup` is healthy, automatically hand off to the same full updater API/workflow used by the UI; or
2. Stop with a prominent message that recovery is incomplete and provide the exact next step to run the normal updater and post-update verification.

Before destructive reset/clean, the script should also state clearly that uncommitted and untracked checkout content will be discarded. If ClawBox supports local extension hooks, invoke their documented backup/restore boundary rather than silently deleting them.

## Acceptance criteria

- Final output distinguishes `updater UI restored` from `full update complete`.
- The default recovery path reaches normal post-update gateway verification.
- The script never claims OpenClaw was updated when it was not.
- Destructive checkout cleanup is disclosed before execution.
- Tests cover an already-current checkout, an updater-code recovery, and a failed gateway after setup-app recovery.

## Related

PRs #263 and #265 added post-update gateway verification to the normal updater. The issue is that the standalone bootstrap script can stop before that verified path.

### ClawBox version

ClawBox main at commit `8cb6a43`

### OpenClaw version

OpenClaw 2026.7.1

### Jetson model / JetPack

[...]

PR #298, opened 2026-08-04
fix(update): clarify force-update UI recovery boundary

jamesachurchill:
## Summary

- preserve force-update.sh as a bootstrap-only recovery tool for a broken updater UI
- warn clearly before tracked changes and non-ignored untracked checkout content are discarded
- replace the ambiguous Done message with an explicit incomplete-recovery boundary
- direct the operator to the restored UI, then Settings -> System Update -> Update everything for the normal verified update
- cover the destructive warning, already-current checkout, UI restoration, and final handoff instructions with focused tests

## ClawReview advice

The ClawReview comment on #285 suggested option 1: automatically invoke the full updater after the setup service becomes healthy.

This change deliberately implements option 2 from the issue instead. The script is documented and structured as an emergency repair for the updater itself: it hard-syncs and rebuilds the setup service, restarts that UI, and returns control to the operator. Automatically continuing into OpenClaw updates and a device reboot would change that repair-only contract and could surprise an operator who only intended to restore access to the UI.

The final output now makes the boundary unmistakable: the UI is restored, but the full updater, OpenClaw update, reboot, and post-update gateway verification have not run. It then gives the exact normal UI action that performs those steps.

## Verification

- 4 focused Vitest tests passed
- Bash syntax and ShellCheck passed
- changed-file ESLint passed
- production build and TypeScript passed
- systemd unit verification passed
- git diff --check passed

Closes #285

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery messaging to clearly explain when recovery is incomplete and how to continue through the standard System Update workflow.
  * Added a 10-second warning before destructive cleanup, including details about tracked and untracked files and preserved ignored data.
  * Clarified that recovery restores the setup interface without automatically completing the full update process.

* **Tests**
  * Added coverage for recovery warnings, cleanup timing, restoration behavior, and already-current installations.


KrasimirKralev:
Superseded by #305. This PR fixed the messaging but left the actual dead-end: after force-update.sh makes ClawBox current, `SystemUpdateApp`'s update button (gated on a ClawBox version delta) doesn't render, so there's still no way to finish the update from the UI and OpenClaw/systemd stay stale (confirmed on-device). #305 adds a 'Force full update' recovery action (Advanced options + confirm) that re-runs the full updater via the existing `{force:true}` path, plus the honest messaging. Suggest closing in favor of #305.

KrasimirKralev:
Closing — superseded by #305 (merged to beta). #305 actually closes the dead-end: a 'Force full update' recovery action (Advanced options + confirmation) that re-runs the full updater via {force:true}, plus the honest force-update.sh messaging. Verified on the Jetson and passed /code-review + CodeRabbit clean. Thanks @jamesachurchill for the original report.

PR #305, opened 2026-08-08
fix(update): add 'Force full update' recovery action for force-update.sh (#285)

KrasimirKralev:
Fixes #285. Supersedes #298 — which only reworded `force-update.sh` and still left the dead-end below.

## The bug
`force-update.sh` self-heals a device whose UI updater is broken, but it only restores the **setup app** — it does not run the full updater's `gateway_setup` + `openclaw_install`. Afterwards ClawBox sits at the current git version, so `SystemUpdateApp`'s "Update everything" button (gated on a ClawBox version delta) never renders → **there is no way to finish the update from the UI**, and OpenClaw / the systemd gateway unit stay stale. The old script message ("Done. Reload …") implied the update was complete.

## The fix
- **`SystemUpdateApp`:** a "Force full update" action under **Advanced options** (not the happy-path hero) that re-runs the full updater via the existing `{force:true}` path — which loops `UPDATE_STEPS` from 0 including `openclaw_install` and `gateway_setup`. Gated behind a **confirmation modal** (it reboots), matching the beta-channel toggle's pattern.
- **`force-update.sh`:** honest completion message — states it recovered the UI only, that being "current" does not mean the update finished, and points at Advanced options → Force full update.

## `/simplify` applied
Placement moved off the happy path into Advanced options + added a confirm (altitude); collapsed the brittle shell-string test into one assertion; trimmed a restated rationale comment. (A double-click guard was considered — after the first click `triggerUpdate` flips `updateStarted` and the up-to-date block unmounts; the confirm modal also gates it.)

## Verified on the Jetson (aarch64)
- `tsc --noEmit`: clean — no errors in changed files (matches `beta` baseline).
- Tests: `system-update-app.test.tsx` (Advanced → Force full update → confirm → asserts `{force:true}`, and that nothing fires before confirm) + `force-update-handoff.test.ts` — both pass.
- `{force:true}` re-running `openclaw_install`/`gateway_setup` verified against `updater.ts` step order.

## Summary by CodeRabbit

* **New Features**
  * Added an Advanced option to force a full system update when the device is already up to date.
  * Added a confirmation dialog with keyboard-friendly controls, including Escape-to-close and focus handling.

* **Bug Fixes**
  * Clarified recovery messaging to explain that only the UI was restored and that a full update must be started separately.

* **Tests**
  * Added coverage for force-update confirmation, cancellation, request behavior, and recovery messaging.

### DIFF
--- a/scripts/force-update.sh
+++ b/scripts/force-update.sh
@@ -78,7 +78,14 @@ echo "[force-update] Restarting clawbox-setup..."
 sudo systemctl restart clawbox-setup
 sleep 5
 if systemctl is-active --quiet clawbox-setup; then
-  echo "[force-update] Done. Reload http://clawbox.local in your browser."
+  echo "[force-update] The ClawBox interface is restored."
+  echo "[force-update] IMPORTANT: this recovered the UI only. OpenClaw and system"
+  echo "[force-update]   services may still be on the OLD version — the interface"
+  echo "[force-update]   being current does NOT mean the update finished."
+  echo "[force-update] To finish: open http://clawbox.local, launch the System Update"
+  echo "[force-update]   app (Settings -> About -> System Update), open 'Advanced"
+  echo "[force-update]   options' and click 'Force full update'. The device reboots"
+  echo "[force-update]   when it completes."
 else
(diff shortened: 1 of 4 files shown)


KrasimirKralev:
@coderabbitai review

PR #329, opened 2026-08-09
feat(hermes): dual-harness — run the ClawBox agent on OpenClaw or Hermes

KrasimirKralev:
## Dual-harness: run the ClawBox agent on OpenClaw **or** Hermes

Adds a second agent harness — **Hermes** (Nous Research) — alongside OpenClaw, with **one shared
identity** and a **user-facing picker**. The user chooses which engine backs the device; each harness keeps
its own providers/auth. Proven end-to-end on a real Jetson (fresh factory-reset → setup → chat on both
harnesses).

### What's in it
- **Harness router** (`src/lib/harness.ts`) — the single source of truth for the active harness
  (`active_harness` config, default `openclaw`), each harness's local endpoint (OpenClaw gateway `:18789`,
  Hermes `serve` `:9119`), and a health probe. Hermes health = server-up **or** the CLI is installed
  (chat uses the `hermes -z` CLI, not the serve endpoint).
- **API** — `GET /setup-api/harness/status` (active + live health of both, for the picker),
  `GET /setup-api/harness/active` (just the active harness, no probes — the chat's lightweight mount check),
  and `POST /setup-api/harness/select` (switch + run the identity sync).
- **Desktop chat routing** — `ChatPopup` resolves the active harness on mount; in Hermes mode it skips the
  OpenClaw gateway WebSocket and routes turns to `POST /setup-api/hermes/chat`. OpenClaw's path is
  unchanged.
- **Hermes chat backend** (`/setup-api/hermes/chat`) — runs `hermes -z <message>` (its persistent default
  session gives multi-turn memory for free). Guards against argv flag-smuggling (strict model charset +
  leading-dash message).
- **Settings picker** (`HarnessPicker`, Settings → System) — OpenClaw / Hermes with live health dots and a
  one-click switch; switching reloads the desktop so the chat re-mounts against the new harness. An offline
  harness can't be selected.
- **Shared-identity bridge** (`scripts/setup-shared-identity.sh` + `clawbox-identity-sync.sh`) — a canonical
  `~/.clawbox/agent-identity/{SOUL,USER,MEMORY}.md` shared by both harnesses. OpenClaw gets **real copies**
  (its workspace scanner doesn't follow symlinks); Hermes gets **symlinks** (it does). The sync self-heals
  the bridge on first run and reindexes Hermes' FTS5 on switch. Providers/OAuth are **not** shared (per
  OpenClaw's own guidance against a shared OAuth grant).

### Validated on-device (Jetson, aarch64)
Factory reset → full setup wizard → desktop → OpenClaw chat replies "OpenClaw" → Settings picker (both
healthy) → switch to Hermes (reload) → chat replies "Hermes". Backend: `harness/status` both healthy,
`hermes/chat` returns `{harness:"hermes"}`, OpenClaw agent responds. Build + e2e green.

### Notes / scope
- Hermes is an **optional** add-on the owner installs separately (Nous installer). Until then the picker
  shows it offline and un-selectable; OpenClaw remains the default.
- v1 is **non-streaming** (Hermes replies land whole) and uses Hermes' single default session. Token
  streaming (via the Hermes serve WS), per-chat sessions, a Hermes-native model picker, and MEMORY (not
  just identity) sync for OpenClaw are follow-ups. `hermes serve`'s API is a FastAPI JSON-RPC/WS backend for
  Hermes' own desktop app (not OpenAI-compatible headless), which is why the `-z` CLI is the v1 path.
- No new dependencies. `setup-shared-identity.sh` should be wired into `install.sh` in a follow-up so the
  bridge exists before the first switch (the sync self-heals it in the meantime).

## Summary by CodeRabbit

* **New Features**
  * Added support for switching between OpenClaw and Hermes chat harnesses.
  * Added harness health indicators and selection controls in System settings.
  * Hermes chat now supports complete responses, error reporting, cancellation, and text-only input.
  * Added shared identity setup and synchronization across supported assistants.
  * Identity files are synchronized automatically when changing the active harness.
  * Added active harness and status reporting for setup and configuration workflows.
  * Added automatic fallback handling when a selected harness is unavailable.

### DIFF
--- a/scripts/clawbox-identity-sync.sh
+++ b/scripts/clawbox-identity-sync.sh
@@ -38,10 +38,13 @@ if [ -d "$OC_WS" ]; then
 fi
 
 # 2. Hermes ← canonical (symlinks already live) → reindex FTS5.
-if command -v hermes >/dev/null 2>&1; then
-  # `hermes memory reindex` rebuilds the recall index from the markdown; fall
-  # back to `sync` on older builds. Non-fatal if neither exists.
-  hermes memory reindex >/dev/null 2>&1 || hermes sync >/dev/null 2>&1 || true
+# Prefer the configured Hermes CLI (HERMES_BIN, matching src/lib/harness.ts's
+# default), falling back to PATH. `hermes memory reindex` rebuilds the recall
+# index from the markdown; fall back to `sync` on older builds. Non-fatal.
+HERMES="${HERMES_BIN:-$HOME_DIR/.local/bin/hermes}"
+[ -x "$HERMES" ] || HERMES="$(command -v hermes || true)"
+if [ -n "$HERMES" ] && [ -x "$HERMES" ]; then
+  "$HERMES" memory reindex >/dev/null 2>&1 || "$HERMES" sync >/dev/null 2>&1 || true
 fi
 
 echo "[identity-sync] done"
--- a/scripts/clawbox-identity-sync.sh
+++ b/scripts/clawbox-identity-sync.sh
@@ -1,0 +1,39 @@
+#!/usr/bin/env bash
+# Propagate the canonical shared identity to both harnesses. Run on
+# harness-switch and whenever ~/.clawbox/agent-identity/* changes (a systemd
+# path unit or the harness/select route can invoke it).
+#
+#   * OpenClaw: refresh the REAL copies (its scanner ignores symlinks) and
+#     restart the gateway so the workspace-file scan re-reads the new content
+#     (OpenClaw caches injected file content at gateway start).
+#   * Hermes: its identity files are symlinks (already live), but its state.db
+#     FTS5 recall must re-index to see external markdown edits.
+#
+# Canonical is authoritative (edit there). Edits made from inside a harness are
+# NOT auto-propagated back in this version — that reverse sync is future work.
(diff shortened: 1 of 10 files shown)


KrasimirKralev:
Addressed all 11 CodeRabbit findings in `afb868d` (build + 1639 tests green on-device; scripts syntax-clean):

- **Reject offline harness** — enforced server-side in `harness/select` (409) in addition to the disabled UI button; `setActiveHarness` stays a pure setter (the health check lives at the route layer where `harnessHealthy` is available).
- **Sync honesty** — `select` now returns `identitySynced` instead of unconditional success; also rejects non-object bodies.
- **Cancel + cap the Hermes process** — the chat route wires `request.signal` to SIGKILL the child, and Stop in Hermes mode aborts the fetch; stdout is capped at 2 MB.
- **Scripts** — `clawbox-identity-sync.sh` uses `HERMES_BIN` (falling back to PATH); `setup-shared-identity.sh` backs up existing Hermes identity files to unique names, never clobbering an earlier backup.
- **Picker** — checks `res.ok` before setting state; the section label is now a heading, not an unbound `<label>`.
- **Health** — requires an *executable* Hermes binary (`X_OK`), not just existence.

PR #362, opened 2026-08-11
fix: make the setup wizard edition-aware where it still assumed the OpenClaw gateway

KrasimirKralev:
Three bugs in one family, all hit on a real Hermes-edition device: UI paths in the setup wizard that still assume the OpenClaw gateway is present.

On this SKU the gateway is not installed and its unit is masked — verified on the device: `systemctl is-enabled clawbox-gateway.service` → `masked`, `is-active` → `inactive`, nothing answering on 18789, while `clawbox-hermes-dashboard{,-proxy}.service` are both active.

## A — the completion overlay waited for a gateway that cannot exist

`SetupCompletionOverlay` hardcoded `t("openclaw.connecting")` plus the gateway restart/wait step labels, and `SetupWizard` polled `/setup-api/gateway/health` for readiness. So the copy named a service the device doesn't have, **and** the wait could only ever exhaust its attempt budget — the owner watched a spinner for something that was never coming.

Now the wizard resolves the edition through the shared `client-harness` cache — already warmed by the AI-models step, so it's a free read — and branches both copy and probe. On hermes it polls `GET /setup-api/hermes/models` and treats `stale: false` as ready: that flag is set only when the **live Hermes dashboard** answered (`source: "dashboard"`); every fallback (on-disk catalog, cold start) is flagged stale, so it cannot pass on a dashboard that is still coming up.

Probe chosen by reading the code, not invented:

- `/setup-api/harness/status` is the obvious candidate but is deliberately in `PRE_AUTH_SENSITIVE_PREFIXES` (desktop picker only) — the wizard can't use it.
- `/setup-api/hermes/oauth` returns `200 {providers: []}` on failure, so up and down are indistinguishable.
- `/setup-api/hermes/models` is already called by this same wizard step, and `/setup/complete` has issued the session cookie by the time the poll runs → no new endpoint, no new middleware carve-out.

## B — the AI-providers step did not auto-advance

On hermes, `AIModelsStep` hands the whole step to `HermesProviderConfig`, and that panel had no auto-advance: a successful save left the owner parked on a finished step with a manual "Continue" as the only way forward, while the OpenClaw branch has advanced automatically all along (`handleConfiguringDone`, 900 ms after the configure lands).

Same pattern, same delay, deliberately conservative:

- fires **once** (latched), and only after a configure round-trip has actually succeeded — the model/provider save, the ClawBox AI tier apply, a pasted portal token, or the device-login poll's terminal `complete` status;
- never when embedded in Settings, which has no next step;
- **never** for the Hermes dashboard's own OAuth, which runs out of band in another tab (`openHermesOAuth`) and still requires an explicit save. A half-finished sign-in cannot carry the wizard forward.

`onNext` is held in a ref because the wizard passes a fresh arrow every render, which would otherwise keep resetting the timer.

## C — the Telegram step waited on the wrong service

`TelegramConfiguringOverlay` showed "Restarting gateway" / "Waiting for gateway" and polled `/setup-api/gateway/health` for up to `healthTimeoutMs` (60 s default). On hermes `restartGateway()` is already a documented no-op (`src/tests/unit/gateway-restart-hermes.test.ts`), so the poll burned the full budget and then took the give-up branch: `onDone()` **without ever reaching the "ready to chat" phase**. A save that had actually succeeded looked like a minute of spinner and then nothing.

The readiness signal here is a *different* service from bug A — not the dashboard, but Hermes' own messaging gateway, the process that receives Telegram messages. The configure route already brings it up (`ensureHermesGateway`), and `/setup-api/telegram/status` already reports it: `receiving` is true only when the token is registered with Hermes **and** that gateway is running — the same "something is listening" meaning `available` carries on the OpenClaw side. The route caches its Hermes probe for 15 s, so a 2 s poll costs one CLI round-trip per window rather than one per attempt.

Both callers benefit: the wizard's Telegram step and Settings.

## Shared

`resolveEdition()` lives in `client-harness.ts`, next to the caches it reads — three screens now make the same "is there an OpenClaw gateway on this box?" decision and duplicated fallbacks drift. Seven new translation keys, all ten locales (en, de, es, fr, it, ja, nl, sv, zh, bg).

## Tests

Regression tests per fix, since edition-conditional behaviour is exactly what rots silently:

- the completion overlay must not name or poll the gateway on hermes, must poll the Hermes route instead, and must still finish when the agent never reports ready;
- the OpenClaw path must still poll the gateway and must never touch the Hermes routes;
- auto-advance fires exactly once, only after a successful configure, not on failure, not when embedded;
- the Telegram overlay reaches the ready phase on hermes without touching the gateway, still polls the gateway on OpenClaw, and on a hermes box whose messaging gateway never starts hands control back **without** claiming success.

Per CodeRabbit, the two give-up tests assert the positive as well as the negative — "did not poll the gateway" alone would also hold for a wizard that polled nothing at all, which is a different bug with the same visible result.

## Notes for review

- `SettingsApp` passes an inline arrow as the overlay's `onDone`, so the overlay's effect re-runs on every parent render and restarts its phase machine. Pre-existing (the effect already depended on `onDone`) and untouched here, but worth a separate look.

## Summary by CodeRabbit

* **New Features**
  * Setup wizard automatically advances after successful provider configuration.
  * Added edition-specific setup and completion experiences for Hermes devices.
  * Hermes setup checks model and Telegram readiness through dedicated status indicators.
  * Added localized Hermes-specific status and completion messages across supported languages.

* **Bug Fixes**
  * Prevented embedded settings panels from advancing automatically.
  * Improved setup completion when device readiness information is delayed or unavailable.
  * Ensured Telegram setup reports readiness only after configuration and service checks complete.

### DIFF
--- a/src/components/HermesProviderConfig.tsx
+++ b/src/components/HermesProviderConfig.tsx
@@ -51,6 +51,12 @@ interface Props {
   testId?: string;
 }
 
+// The OpenClaw AI-models step advances the wizard ~900 ms after a successful
+// configure (AIModelsStep's handleConfiguringDone), which is long enough for the
+// "connected" state to register and short enough not to feel stalled. This panel
+// is the SAME step on a Hermes device, so it advances the same way.
+const AUTO_ADVANCE_DELAY_MS = 900;
+
 function readStoredUiTier(): ClawaiTier {
   if (typeof window === "undefined") return "flash";
   try {
@@ -98,6 +104,32 @@ export default function HermesProviderConfig({ embedded, onNext, testId }: Props
   const [apiKey, setApiKey] = useState("");
   const [saving, setSaving] = useState(false);
   const [saveStatus, setSaveStatus] = useState<Status>(null);
+
+  // ── Auto-advance ───────────────────────────────────────────────────────────
+  //
+  // Set ONLY once a configure round-trip has actually succeeded — a provider
+  // sign-in that happens out of band (the dashboard's own OAuth page, opened in
+  // another tab by openHermesOAuth) deliberately does NOT set it: returning from
+  // that tab has not configured anything yet, the user still has to save. So a
+  // half-finished sign-in can never carry the wizard forward.
+  const [configured, setConfigured] = useState(false);
+  const advancedRef = useRef(false);
+  // `onNext` is an inline arrow from the wizard, so its identity changes on
+  // every parent render. Held in a ref so a re-render cannot restart the timer.
+  const onNextRef = useRef(onNext);
+  useEffect(() => { onNextRef.current = onNext; }, [onNext]);
+
+  useEffect(() => {
+    // Settings embeds this panel with nowhere to advance to; only the wizard
+    // passes an onNext.
+    if (!configured || embedded) return;
+    const timer = setTimeout(() => {
+      if (advancedRef.current) return;
+      advancedRef.current = true;
+      onNextRef.current?.();
+    }, AUTO_ADVANCE_DELAY_MS);
+    return () => clearTimeout(timer);
+  }, [configured, embedded]);
 
   // Tell an already-open chat popup that this device's provider/model/tier
   // changed, so its header re-seeds instead of naming the previous provider
@@ -260,9 +292,12 @@ export default function HermesProviderConfig({ embedded, onNext, testId }: Props
     onBusyChange: setLoginBusy,
     onConfiguring: () => setClawaiStatus({ kind: "ok", msg: "Finishing setup on this device…" }),
     onComplete: () => {
+      // Only reached on the poll's terminal `complete` status, i.e. after the
+      // device finished configuring — never mid-handshake.
       void reloadClawai();
       setSelectedProvider(CLAWAI_PROVIDER);
       setClawaiStatus({ kind: "ok", msg: "ClawBox AI is now your active model" });
+      setConfigured(true);
     },
     onError: (msg) => setClawaiStatus({ kind: "err", msg }),
   });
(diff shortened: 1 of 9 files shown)


KrasimirKralev:
### CI `test` failure — root-caused and fixed (`e25fb05`)

Heads-up for **#364**: this touches `src/tests/routes/ai-models/configure.test.ts`, which is in your scope. It is a **test-file-only** change — one added `vi.mock` — and no production code under `src/app/setup-api/ai-models/**` is modified. Whoever merges second should expect a touch on that file.

**What it was.** The job exited 1 while reporting all 178 test files as passing:

```
Vitest caught 1 unhandled error during the test run.
EnvironmentTeardownError: [vitest-worker]: Closing rpc while "onUserConsoleLog" was pending
Test Files  178 passed (178)
```

Step 8c of the configure route kicks off a provider catalog refresh and deliberately does not await it. `refreshInBackground` starts a fetch (or an openclaw fork) and logs on **both** branches — `[catalog] refreshed <provider>` / `[catalog] refresh failed for <provider>` — so in a unit test it settles and writes to the console *after* the test that triggered it has returned. A console write with no owning test still travels over the worker's RPC channel; arriving during teardown, it raises that rejection and fails the run despite every assertion passing.

**Confirmed, not guessed.** Replacing the mock body with a `throw` made the route surface it as `[configure] Failed to configure AI model: MOCK_HIT_PROBE` across many cases, proving the unawaited call is on the hot path of this file's tests. Reverted to a plain no-op stub afterwards.

**Deliberately not done:** silencing the console, or a global unhandled-rejection swallow — both would hide this class of defect rather than remove it. Nothing in the file asserts on the refresh; it is out-of-band work by design.

**Pre-existing, not introduced here.** The same error hit `feat/hermes-edition` at **164** test files (run `31447461238`, ~7.5 h before this branch existed), and it is intermittent — this branch's `6aa6c13` failed on attempt 1 and passed on attempt 2 with identical code. Adding two test files (177 → 178) shifted worker scheduling enough to land it more often. It blocks this PR, so it is fixed here.

PR #363, opened 2026-08-11
fix: skills browse first-run state and chat provider list refresh

KrasimirKralev:
Two UI fixes found while testing a Hermes device.

## A — Browse said the skill catalogue was empty while its index was still building

On a freshly flashed box, Skills → Browse showed "Nothing in All sources yet / Try a different term". Waiting and retrying made ~90k skills appear, so the catalogue was fine — the UI was describing a first-run state it could not recognise.

Two gaps, both fixed:

- `setup-api/hermes/skills/browse` read `isWarming()` *before* kicking the build, so the very first request — the one that starts it — reported itself as a plain CLI answer. It now kicks first and then reports, which is the only way that request can describe itself honestly.
- The store gated its "building the catalogue" panel on `loading`. That request **completes** while the index warms (it falls back to the CLI, which on a fresh device also has nothing), so `loading` is already false by the time anything renders and the generic empty state won. `useSkillCatalog` now exposes `preparing` — warming origin, zero results — which both shows the first-run panel and suppresses the "nothing here" copy. "Nothing here" is a claim about a catalogue the device does not have yet.

It also recovers on its own: while the answer stays `warming` the hook re-asks every 5 s and stops the moment the origin changes, so nobody has to reopen the window or retype a search. The panel now says so too.

## B — The chat provider dropdown did not update when a provider was added

Adding or authenticating a provider left the chat's dropdown on the old set until a full page reload.

The server side was already correct — every write drops the model-options cache, and the `hermes config get` memo keys on `config.yaml`'s mtime. The gap was that nothing told the client to re-read.

`clawbox:hermes-model-state-changed` already existed but was emitted from one path (the Save button) and listened to in one place. It is now a named export next to `useHermesModelOptions`, the hook subscribes to it, and the two remaining configure paths emit it: the ClawBox AI device login, and the return from Hermes' own sign-in page (exactly when a provider flips to authenticated).

Kept deliberately cheap, as the hook's own comment asks: the listener bumps the nonce rather than calling `refresh`, so the re-read is a plain scoped GET. `refresh` sets the explicit flag, which busts Hermes' per-provider disk cache and re-enumerates every provider's live `/v1/models` — a device-wide sweep to answer "is the list different?". It also leaves the loaded scope in place, so `loading` never flips and the model pill keeps its position (`chat-model-pill-stability` still passes unchanged).

## Tests

New: `hermes-skills-warming.test.tsx` (5) and `hermes-provider-invalidation.test.tsx` (8). The warming tests fail 4/5 against the previous code; the fifth pins that a genuinely empty result on a device that *has* a catalogue is still reported as empty.

- Unit: 2162 passed, 0 failed (+13 over the 2160 baseline).
- E2E: 41 passed / 2 skipped / 0 failed, including `chat-popup.spec.ts:204`.
- Build clean, lint clean on the changed files.

## Summary by CodeRabbit

- **New Features**
  - Skill catalogues show a clear “building automatically” status while preparing.
  - Catalogues refresh automatically and display newly available skills without user action.
  - Hermes model options update automatically after provider or device configuration changes.
- **Bug Fixes**
  - Initial skill browsing correctly reports when index construction has started.
  - Improved synchronization of model and reasoning settings across open views.
  - Prevented outdated model requests from overwriting newer settings.
  - Preserved appropriate empty-state messaging for genuinely empty catalogues.
- **Tests**
  - Added coverage for catalog warming, automatic refresh, model invalidation, and cross-view updates.

### DIFF
--- a/src/app/setup-api/hermes/skills/browse/route.ts
+++ b/src/app/setup-api/hermes/skills/browse/route.ts
@@ -118,8 +118,15 @@ export async function GET(request: Request) {
   }
 
   // ── No index: answer from the CLI once, and kick a build for next time ──
-  const warming = isWarming();
+  // Kick the build FIRST, then report. Asking `isWarming()` beforehand meant the
+  // very first browse on a fresh device — the request that starts the build —
+  // described itself as a plain CLI answer, so the UI had nothing to distinguish
+  // "still preparing" from "genuinely nothing here" and showed the
+  // empty-catalogue copy over a catalogue that was seconds from existing.
+  // Still false when the post-failure cooldown declines to start another build,
+  // and then "no index, not building" is the honest answer.
   warmIndex();
+  const warming = isWarming();
   try {
     // A catalog-only source (skills.sh spelling, claude-marketplace, unknown)
     // has no `--source` flag — drop the filter rather than send hermes a value
--- a/src/components/ChatPopup.tsx
+++ b/src/components/ChatPopup.tsx
@@ -1526,16 +1526,39 @@ function ChatPopup({ isOpen, onClose, onOpenFull, onOpenSettingsSection, onThink
   // Safe to re-run: every in-session choice is mirrored into localStorage by the
   // change* callbacks, and those prefs win here, so a re-sync preserves the
   // user's picks while picking up new credentials / a new device pairing.
+  //
+  // LATEST SEED WINS. Seeds overlap now that three call sites emit the
+  // invalidation (a device login, an OAuth return and a Save can land within a
+  // second of each other, and the mount seed can still be in flight when the
+  // first event arrives). They all hit the same route, so a slower EARLIER
+  // response can resolve last and reinstate the provider list, device model and
+  // selection the newer one just replaced — the same "a stale provider wins"
+  // failure the model pill guards against, reached through a different door.
+  // Each seed therefore takes a generation, cancels the one it supersedes, and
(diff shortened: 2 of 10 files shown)


KrasimirKralev:
Addressed the seed race in `4609ce9` — the finding is correct and it is a direct consequence of this PR: making the invalidation fire from three call sites instead of one is what made overlapping seeds reachable.

The listener shared one `AbortController` across every event, which cannot express ordering, and the mount seed could still be in flight when the first event arrived. Each seed now takes a generation, aborts the seed it supersedes, and writes only if it is still the newest when its answer lands. `aborted` alone would not be enough — a response can resolve before the abort lands, and `json()` adds a second await for a newer seed to start in. The caller's signal is forwarded, so unmount and a harness switch still cancel; the effect's controller keeps that job and nothing else.

Regression test in `chat-header-seed-race.test.tsx` resolves the **first** of two seeds **last** — in-order resolution passes with or without the guard, so it would prove nothing. Both tests fail against the previous commit. They cover what the header names and what the dropdown offers, since a stale seed winning means a just-connected provider disappears from the picker again.

Re-verified after the change: unit **2175 passed / 0 failed** (179 files, Linux), e2e **41 passed / 2 skipped / 0 failed** run serially against a production build, build and lint clean. `chat-model-pill-stability` is untouched and green — the guard never flips `loading`/`fresh`, so the pill still holds its place.

PR #371, opened 2026-08-11
fix: focus trap for the install dialog, upstream timeouts for the hermes dashboard proxy

KrasimirKralev:
Two robustness defects found during review. Independent; separated per commit.

## 1. The install confirmation could not be completed with a keyboard

Three separate defects, and the first is the one that matters most:

- **Escape had never worked.** It was bound as an `onKeyDown` on the backdrop `div`. A `div` is not focusable, so the key event was never routed to it — the handler could not fire. The existing code only *appeared* to handle Escape.
- **`role="dialog"` sat on the full-screen backdrop**, which made the accessible dialog the entire viewport and its accessible name absorb everything behind the scrim.
- **Focus never moved into the dialog**, so it stayed on the "Install" button behind the scrim and Tab walked the store's search box, sort menu and category tabs — the controls the dialog exists to block. There was no way to reach "Install anyway" by keyboard, and no way to dismiss.

`src/components/hermes-skills/ConfirmDialog.tsx` already did this correctly, so rather than add a fourth pattern its implementation is extracted into `src/hooks/useModalDialog.ts` and the hand-rolled copies are replaced by it:

- focus moves into the panel on open, restored to the trigger on close
- Tab / Shift-Tab cycle inside the panel, and focus is pulled back if it lands outside
- Escape closes and stops there, so the window behind does not also act
- the page behind the panel is `inert` + `aria-hidden` while it is open

**Callers covered**

| Dialog | Before | Now |
|---|---|---|
| App store install confirmation | no focus-in, no trap, Escape dead, role on backdrop | full |
| Hermes skills store install/uninstall | correct (extraction source) | unchanged + background inert |
| ClawBox sign-in modal | Escape + focus-in only | full |
| Tier upgrade dialog | Escape only | full |
| System update confirmation | its own duplicate trap | shared trap + restore + inert |

**Deliberately not converted:** the ClawKeep dialogs, the VNC dialog, and the inline confirmations in `SettingsApp`. They share the defect but sit in large files outside this flow, and converting them would have inflated a diff that is already touching an e2e blast zone. **`useModalDialog` is now available for them** — that is the obvious follow-up for whoever picks it up.

## 2. `scripts/hermes-dashboard-proxy.js` had no upstream timeouts

An upstream that accepted the socket and then went quiet held the browser's request open with no status and no error. Each timeout is armed only for the phase where silence means stuck, and disarmed after, so nothing that legitimately streams is cut:

- **HTTP** — request to first response byte (30s), disarmed once headers arrive
- **SSO login** — the broker's password-login POST (10s); every request awaiting the shared promise would otherwise queue behind it
- **WebSocket** — TCP connect plus the 101 handshake (15s), disarmed on the upstream's first byte

All three are env-overridable.

### The naive fix is worse than the bug, and I checked rather than assuming

Each timeout path **writes its status before** tearing the socket down. I verified both halves of that by removing my own pre-write and re-running:

- **HTTP:** letting `destroy()` speak for itself produced a **misleading 502** ("not reachable") instead of a 504 — the operator would chase a network fault that does not exist. `destroy()` is specified to emit `close`; an `error` only follows in some shapes, and once a response is in flight it does not end the client's response at all, because `pipe()` does not propagate a source error to its destination.
- **WebSocket:** a bare `kill()` gave the browser a **silent 1006** — the assertion failed on an empty body, so nothing distinguishes a wedged dashboard from a network blip.

A timeout that produces silence reintroduces the symptom it was added to remove.

To make any of this testable the script now exports a server factory and keeps `listen()` behind a `require.main` guard; it previously bound its port at import time. The request path itself is unchanged, and the systemd unit still runs it as the main module.

## Testing

- **Unit**: 190 files / 2239 tests, +13 new. Same 12 files / 29 failures as `beta` on the Windows dev box used here (platform-specific: path separators and shell tooling); failure sets diffed identical, so no regressions. CI `test` is green.
- **Focus-trap tests assert where focus lands** after Tab, Shift-Tab and close — not that an attribute is present. All seven **fail against the previous markup** (verified by stashing the component fix and re-running). An a11y test that passes on broken markup is worse than none.
- **Proxy tests prove the timed-out request returns**: a silent upstream answers 504 in ~330ms, and the same request **hangs** with the timeout disabled — that test pins the old behaviour. Also covers 502 for an absent upstream, an explicit 504 on a stalled WebSocket handshake (verified non-vacuous), and that a healthy upstream still proxies.
- **E2E**: green in CI. Locally 38 passed + 3 flaky (all green on retry) + 2 skipped, 0 failed — matches the 41/2/0 baseline. `store-flow.spec.ts`, which drives the install modal end to end, passed. The flakes are the timing-sensitive specs `playwright.config.ts` already documents as flaky on slow runs.
- **Lint** clean (only pre-existing `` warnings); **typecheck** clean for every changed file.

### What this testing does not cover

- **No hardware verification.** Everything ran locally on a Windows dev machine; no Jetson was involved. Nothing here has been exercised on a real device.
- The local e2e run needed a workaround: `package.json`'s `postbuild` is POSIX shell and fails under Windows tooling, so its copy step was replicated by hand and Playwright pointed at an already-running standalone server. That is a local-environment limitation, not a repo defect, and `package.json` was not touched. CI ran the real pipeline.
- `e2e-install` is unaffected by these changes — `60-app-store.spec.ts` drives `setup-api` directly rather than the UI.

## Summary by CodeRabbit

- **New Features**
  - Improved modal accessibility with reliable focus management, keyboard navigation, Escape-to-close behavior, focus restoration, and background content isolation.
  - Added configurable timeout settings for dashboard responses, login requests, and WebSocket handshakes.

- **Bug Fixes**
  - Dashboard proxy timeouts now return clear 504 or 502 errors instead of hanging or closing silently.
  - Established WebSocket connections are no longer interrupted by handshake timeout handling.

- **Documentation**
  - Documented optional timeout settings, including how to disable them.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -20,3 +20,10 @@ ALLOW_INSECURE_CONTROL_UI=true
 # Optional override for OpenClaw's output budget. If unset, ClawBox uses the
 # full llama.cpp context window instead of a smaller app-level cap.
 # LLAMACPP_MAX_TOKENS=131072
+
+# Optional: Hermes dashboard proxy (scripts/hermes-dashboard-proxy.js) upstream
+# timeouts, in milliseconds. Each is armed only for the phase where silence
+# means the upstream is stuck; 0 disables one. Defaults shown.
+# HERMES_DASH_UPSTREAM_TIMEOUT_MS=30000   # request sent -> first response byte
+# HERMES_DASH_LOGIN_TIMEOUT_MS=10000      # absolute deadline for the SSO login
+# HERMES_DASH_WS_TIMEOUT_MS=15000         # TCP connect + WebSocket handshake
--- a/e2e-install/80-chat.spec.ts
+++ b/e2e-install/80-chat.spec.ts
@@ -142,6 +142,23 @@ test.describe("chat round trip", () => {
 
     await page.goto("/");
 
+    // Configuring a provider above flips the account tier, which makes the
+    // tier-upgrade dialog open over the desktop on first load. It is a real
+    // modal — `aria-modal="true"`, and the page behind it is `inert`, so the
+    // chat panel is genuinely unreachable until it is dismissed, for a screen
+    // reader and for this test alike. Dismiss it the way a user would rather
+    // than querying around it; it is absent on reruns (the "seen" tier is
+    // persisted), hence the tolerant wait.
+    const celebration = page.getByRole("dialog").filter({ hasText: /plan/i });
+    const celebrationShown = await celebration
+      .waitFor({ state: "visible", timeout: 10_000 })
+      .then(() => true)
+      .catch(() => false);
+    if (celebrationShown) {
+      await celebration.getByRole("button").first().click();
+      await expect(celebration).toBeHidden({ timeout: 10_000 });
+    }
+
     // Before the gateway WS connects, the textbox shows
     // "Waiting for the Claw to wake up…" and is disabled. Once the gateway
     // acknowledges the session, the placeholder flips to "Type a message..."
--- a/scripts/hermes-dashboard-proxy.js
+++ b/scripts/hermes-dashboard-proxy.js
@@ -79,13 +79,20 @@ const DASH_USERNAME = process.env.HERMES_DASH_USERNAME || "clawbox";
 // vanish (a bare 1006, no status). Silence is the very symptom these timeouts
 // exist to remove, so every timeout path below writes its own status FIRST and
 // tears the socket down after.
+//
+// Note the LOGIN timer is an absolute deadline rather than a socket timeout —
+// `setTimeout` on a socket measures inactivity, which an upstream that dribbles
+// bytes can reset forever. See hermesLogin().
 function envMs(name, fallback) {
-  const raw = Number.parseInt(process.env[name] || "", 10);
+  const raw = Number.parseInt(process.env[name], 10);
   return Number.isFinite(raw) && raw >= 0 ? raw : fallback;
 }
 const UPSTREAM_HEADERS_TIMEOUT_MS = envMs("HERMES_DASH_UPSTREAM_TIMEOUT_MS", 30_000);
 const LOGIN_TIMEOUT_MS = envMs("HERMES_DASH_LOGIN_TIMEOUT_MS", 10_000);
 const WS_HANDSHAKE_TIMEOUT_MS = envMs("HERMES_DASH_WS_TIMEOUT_MS", 15_000);
+// End of an HTTP header block — what the WS path waits for before it will call
+// the upgrade handshake complete and stand its timeout down.
+const HEADER_TERMINATOR = "\r\n\r\n";
 
 // Rewrite the origin part of a Referer to the upstream authority, keeping path.
 function rewriteReferer(value) {
@@ -347,6 +354,18 @@ function hermesLogin() {
   if (!pw) return Promise.resolve(null);
   const body = JSON.stringify({ provider: "basic", username: DASH_USERNAME, password: pw, next: "/" });
   return new Promise((resolve) => {
+    // EVERY request needing SSO awaits this one promise (see ensureHermesCookies),
+    // so it must settle exactly once and it must always settle. `settle` gives
+    // both: later callers are ignored, and the deadline below guarantees one.
+    let settled = false;
+    let deadline = null;
+    const settle = (value) => {
+      if (settled) return;
+      settled = true;
+      if (deadline) clearTimeout(deadline);
+      resolve(value);
+    };
+
     const reqUp = http.request(
       {
         host: UPSTREAM_HOST,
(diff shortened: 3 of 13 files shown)


KrasimirKralev:
### `e2e-install` failure on the first run: not from this branch

The failing spec is `80-chat.spec.ts:114`. It ran against **`main`'s code**, not this branch — the harness installs `main` and only switches to the PR branch two specs later.

**1. The container installs `main`.** `e2e-install/docker-compose.test.yml:49`:

```yaml
# Override branch target for upgrade tests. Default: install from main.
CLAWBOX_BRANCH: ${CLAWBOX_BRANCH:-main}
```

`CLAWBOX_BRANCH` appears **zero times** in `.github/workflows/e2e-install.yml` — the workflow sets only `CLAWBOX_UPGRADE_TARGET_BRANCH`, which is the *upgrade destination*, not the install source. So the default holds and the box boots `main`.

**2. Order is strict, and chat runs before the upgrade.** `e2e-install/playwright.config.ts` has `fullyParallel: false`, `workers: 1`, `// one shared container, strict order`, so the printed indices are execution order:

```
✘ 75  80-chat.spec.ts:114        › UI chat widget receives a streamed response
✓ 82  90-upgrade-main-to-beta.spec.ts › pin .update-branch to fix/install-modal-and-proxy-timeouts
✓ 84  90-upgrade-main-to-beta.spec.ts › update completes through restart and post_update
✓ 85  90-upgrade-main-to-beta.spec.ts › git HEAD is on fix/install-modal-and-proxy-timeouts
```

Spec 90's own first test — `verify current branch is main` — passed before the upgrade. This branch's code reached the device at #82 and **every test from that point passed**, including the reboot in `99-power`.

**3. On the last green sibling run, that same test did not pass either — it was skipped.** In `fix/build-hygiene` (success):

```
-  75 [install] › 80-chat.spec.ts:114 › UI chat widget receives a streamed response
```

It skips itself when the gateway never wakes; here it failed 30s earlier, at the textbox. Same code, different upstream weather. The same run also logged `[app-store] openclaw skills install fallback: Install failed. Please try again.` — ClawHub was flaky for that run too.

**On the `inert` / `aria-hidden` hypothesis** — right thing to check, and worth pinning regardless of this run. Restoration happens in the effect cleanup, which React runs on unmount as well as on close, so it cannot outlive the dialog. That is now asserted directly: a test opens the dialog, unmounts the whole tree with it still open, and checks the background nodes carry neither attribute. Stacked dialogs were a real (separate) defect and are fixed in `8ccefb7` — each trap records and restores the prior attribute values, and only the topmost dialog reacts to keys.

A re-run of the failed job was kicked off to confirm flake-vs-deterministic; the new push supersedes it with a full fresh run.

No spec was skipped, disabled, or retry-wrapped.

KrasimirKralev:
### Correction — the `e2e-install` failure IS from this branch. Root cause found.

**My previous comment was wrong and I'm retracting it.** I claimed specs 10–80 run `main`'s code because `docker-compose.test.yml` defaults `CLAWBOX_BRANCH` to `main`. That variable only steers the *updater's* git operations. The image is built with `context: ..` and `e2e-install/Dockerfile:98`:

```dockerfile
COPY --chown=clawbox:clawbox . /opt/clawbox-src/
```

`entrypoint.sh` seeds `/home/clawbox/clawbox` from that baked-in tree. **Every spec runs this branch's code.** I should have traced the Dockerfile before drawing a conclusion; the `CLAWBOX_BRANCH` default made a tidy story and I stopped there.

### The hypothesis about closed dialogs is refuted

The suggestion was that `inert` might be applied on mount rather than on open, poisoning the tree for permanently-mounted dialogs. That is not happening — new tests in `modal-dialog-inert.test.tsx` assert it directly:

- a mounted-but-closed dialog leaves `document.querySelectorAll("[inert]")` and `[aria-hidden="true"]` both empty
- a background control is still reachable by `getByRole` while the dialog is closed
- the marking appears only after open, and the dialog's own subtree is never marked

Those tests are worth keeping regardless — it's the right thing to guard.

### What is actually happening

From the failing run's trace DOM:

```
["DIV",{"data-testid":"chat-popup","inert":"","aria-hidden":"true", ...}]
["DIV",{"data-testid":"desktop-surface","inert":"","aria-hidden":"true", ...}]
```

and the page snapshot at failure:

```yaml
- dialog "Welcome to ClawBox Max 🎉":
    - generic: Max plan
    - button "Let’s go" [active]
```

Configuring an AI provider earlier in the suite flips the account tier, so **`TierUpgradeCelebration` is genuinely open** on that desktop. It already declared `role="dialog"` + `aria-modal="true"` **before this PR** — a promise to assistive tech that everything outside it is inert. That promise used to be false; a screen-reader user could wander the desktop behind it. This PR makes it true, so the chat panel behind the modal is now correctly out of the accessibility tree.

`getByRole` resolves against the accessibility tree, so the locator stops matching — exactly the shape observed. The element is still visible; a CSS-selector query would have found it.

### The fix

The behaviour is correct and stays. The spec was reaching *through* a modal that a real user must dismiss first, so it now dismisses it the way a user would — clicking the dialog's button — before driving the chat. Tolerant of absence, because the dialog does not reappear once the tier is marked seen.

Nothing is skipped, disabled, or retry-wrapped, and no timeout was weakened.

### Worth keeping in mind

This is a good argument for role-based locators. A CSS-selector test would have sailed through while the page was, briefly, differently reachable to assistive tech — the suite noticed precisely *because* it queries the way a screen reader does. It also caught the one thing my local desktop e2e run could not: `e2e-install` boots the real installed product, where the tier transition actually fires.

PR #376, opened 2026-08-11
docs: record why three CodeQL refutations hold

KrasimirKralev:
Closes out the CodeQL triage. Comment-only change (12 lines, no behaviour).

An adversarial pass over the open alerts concluded that most are false positives. This records the reasoning for three of them at the site, so the next reader — human or tool — does not re-derive it, and pairs with per-alert dismissals in the Security tab, the mechanism `.github/codeql/codeql-config.yml` already nominates.

### The three notes

- **`src/lib/edition-source.ts`** — why the `CLAWBOX_EDITION_FILE` override is not an edition-lock bypass: no request data flows into it, and `dual` must additionally clear `verifyDualLicense()` against a key constant that is deliberately not env-overridable. A redirected path can change the edition *label*, never unlock the premium switcher.
- **`mcp/lib/edition.ts`** — same constant, but the stake here is the *tool set*, not the licence. The env supplying the path is the one that launches this server, so anyone able to set it already has the shell the smaller set withholds.
- **`src/lib/code-projects.ts`** — why the constructed `RegExp` is not a defect: it is a search filter, deciding no path, permission or sanitisation outcome.

### Triage outcome

The false-positive alerts are dismissed individually in the Security tab, each with the specific guard named.

A small number did not survive verification and are **not** dismissed. They are tracked privately and handled separately from this PR, which is comment-only.

### DIFF
--- a/mcp/lib/edition.ts
+++ b/mcp/lib/edition.ts
@@ -20,6 +20,10 @@ const HARNESS_TIMEOUT_MS = 3_000;
 
 // Same path src/lib/edition-source.ts reads; re-stated (not exported from there)
 // only to tell "the lock is absent" apart from "the lock is unreadable".
+// Not an edition-lock bypass — see the note on the same constant in
+// src/lib/edition-source.ts. What it can influence HERE is the tool set, and the
+// env supplying it is the one that launches this server, so anyone able to set it
+// already has the clawbox shell that the smaller set withholds.
 const EDITION_FILE = process.env.CLAWBOX_EDITION_FILE || "/etc/clawbox/edition.env";
 
 /**
--- a/src/lib/code-projects.ts
+++ b/src/lib/code-projects.ts
@@ -409,6 +409,10 @@ export async function searchFiles(
     // No global flag: only test() is used, and the 'g' flag makes test()
     // stateful (persisting lastIndex), which silently skips alternating matches.
     const flags = opts.caseSensitive ? "" : "i";
+    // Flagged by CodeQL js/regex-injection: the pattern is a search FILTER, not
+    // a guard. It decides no path, permission or sanitisation outcome — only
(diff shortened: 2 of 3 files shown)


KrasimirKralev:
CodeRabbit dispositions — both declined for this PR, with reasons.

**1. `code-projects.ts` — regex backtracking.** Agreed on the substance, and the PR does not claim otherwise: an earlier draft of this comment said the length cap "bounds the matching work", which is false (a 6-character pattern is still exponential in a 2000-character slice), and that wording was removed before the commit. The note now only claims what holds — the pattern is a filter that decides no path, permission or sanitisation outcome.

The residual is availability, not access control, and it is pre-existing and authenticated-only (`/setup-api/code` is in `PRE_AUTH_SENSITIVE_PREFIXES`). A caller who can reach it can already spend the event loop more directly. Out of scope for a comment-only change; recorded for separate triage rather than fixed here.

**2. `edition-source.ts` — resolve once at startup from the fixed path.** Declined: this over-applies a guideline scoped to a different module. CLAUDE.md's "resolved once at startup" and the matching header in `mcp/lib/edition.ts` are about the **MCP tool set**, which must not change under a running server. `src/lib/edition-source.ts` is the shared reader and documents the opposite contract in its own docstring — "called per request from middleware", with an mtime cache making the re-read cheap and self-correcting.

Removing `CLAWBOX_EDITION_FILE` would also break the suite by construction: `vitest.config.ts` points it at a nonexistent path as a suite-wide guarantee, and eight test files set it. The `CLAWBOX_EDITION` fallback is likewise deliberate and documented for dev boxes and CI, where `/etc/clawbox` does not exist. Neither is a lock weakness — the premium SKU is gated by `verifyDualLicense()`, which is what this PR's comment records.

PR #377, opened 2026-08-11
fix: refuse an unsupported edition transition and make validation honest

KrasimirKralev:
## Problem

Re-running the installer with a different `CLAWBOX_EDITION` on a device that is already provisioned completed all 26 steps and printed a healthy summary — for a device left in a state it cannot recover from on its own.

The installer installs an edition; it does not migrate one. There is no step that stops, disables or masks the harness a device is leaving, and each harness keeps its AI provider credentials in its own config file, so a sign-in does not move across. What you get is:

- both harnesses running at once — two agents, two dashboards, and two pollers on one Telegram bot token;
- the incoming harness with an empty provider registry, so no model resolves and every chat turn fails;
- a previously signed-in device reading as signed out;
- `setup_complete` carried over, so the wizard step that would repair all of it never runs again.

`docs-site/editions/overview.mdx` already documented edition changes as reflash-only. The installer just did not enforce it.

## What this changes

**1. The installer refuses the transition.** `install.sh` compares the requested edition against the root-owned lock — with the legacy systemd drop-in as the fallback, so boxes provisioned before `edition.env` existed are covered too — and stops before anything is changed. The message states what the device currently is, what was requested, that changing edition requires a reflash, and names the single override.

The check sits immediately after the edition resolves, during constant parsing. That is ahead of the lock and drop-in rewrite, ahead of any mask/unmask, and ahead of any unit being copied, enabled or started — on the full-install path **and** on the `--step` dispatch path the in-app updater uses, since both reach that point before any step function is even defined. A refusal that fires after the gateway has been unmasked would not be a refusal.

`CLAWBOX_ALLOW_EDITION_CHANGE=1` is the one explicit escape hatch. It is named in the refusal and left out of the customer-facing docs on purpose.

**2. The other writer of the lock carries the same refusal.** `scripts/setup-hermes-edition.sh` also rewrites `/etc/clawbox/edition.env` and the legacy drop-in, and it read `$CLAWBOX_EDITION` ahead of the lock — the one place where the environment outranked the root-owned file, and the place that then rewrote it. Guarding only `install.sh` would have left that path open, and worse, it launders the change: once the lock reads the new edition, a later plain `install.sh` sees no mismatch and provisions on top. Both writers now refuse, and this one reads the lock first like every other consumer.

**3. `dual` is a distinct edition in both directions.** `openclaw → dual` and `hermes → dual` are refused too. The additive direction fails the same way as the destructive one — the harness being *added* comes up with no provider registry and no credentials, and `setup_complete` suppresses the step that would populate them — and the lock exists precisely so the SKU cannot be selected from the environment.

**4. Validation is honest about foreign units.** `step_validate_services` only ever asserted that *this* edition's units are up; it appends the Hermes units inside `if has_hermes_harness`, so on `openclaw` a fully running Hermes stack was not merely tolerated, it was invisible. A new `FOREIGN_EDITION_UNITS` registry — built from the existing `has_hermes_harness` / `has_openclaw_harness` predicates, so `dual` accumulates nothing and needs no special case — is now checked, and any foreign unit that is active *or* merely enabled fails the run. That is what stops the healthy summary being printable on a device running two harnesses.

Edition migration is deliberately **not** implemented. Doing it properly means migrating credentials, tearing down the other harness and re-running the AI step; that is a feature, not this fix.

## Unaffected

- **Fresh installs** — no lock yet, so nothing to disagree with.
- **Same-edition re-installs** — recorded equals requested.
- **`post_update` / the in-app updater** — `clawbox-root-update@.service` exports `CLAWBOX_EDITION` from the lock, so requested and recorded are equal by construction on every step it dispatches.
- **`config/edition.txt`** is still a resolution source but is never treated as a *record*; it lives in a customer-writable tree.
- An unrecognised recorded value normalises to `openclaw` before comparing, so a typo'd lock cannot brick a device's updates.

## Tests

`src/tests/unit/install-edition-switch-refusal.test.ts` (43 tests). Following the conventions in `install-edition-lock.test.ts` and `register-mcp-hermes.test.ts`, these **execute the shipped shell text** rather than string-matching it: the edition-resolution/refusal region and `step_validate_services` are extracted from `install.sh` and run against a temp lock file and a fake `systemd`.

- Refusal fires on lock/env disagreement, and the lock file is byte-identical afterwards with no drop-in created.
- The escape hatch permits it; only a literal `1` opens it (`yes`, `true`, `2`, `0` all still refuse).
- Fresh install, same-edition re-install, updater path (no env var), case-insensitive lock, and unrecognised lock values are all unaffected.
- All six recorded→requested transitions refuse; all three same-edition re-installs pass.
- Ordering is pinned structurally: the refusal precedes the first `step_*` definition, `DISPATCH_STEPS`, and the installer banner.
- A test asserts the refusal block contains no `systemctl stop|disable|mask|unmask` and no `rm -rf`, holding the "no migration" scope boundary.
- Validator: a clean `openclaw` box passes; the same box with the Hermes stack alive fails 3 of 15; enabled-but-inactive fails; installed-but-disabled passes; `hermes` still running the gateway fails; `dual` running both is healthy.

Full suite green — 197 files, 2458 tests. `eslint` clean on the changed files, `shellcheck -S warning` clean on both scripts, no new `tsc` errors.

## Summary by CodeRabbit

* **New Features**
  * Installation now records and validates the selected edition.
  * Reinstallations reuse the recorded edition by default and prevent incompatible edition changes.
  * Edition changes can be explicitly overridden when required.
  * Installation checks now detect active or enabled services from other editions.

* **Documentation**
  * Added guidance on first-install edition selection, reinstallations, overrides, and reflashing requirements.

* **Bug Fixes**
  * Improved handling of invalid, legacy, missing, and case-variant edition values.

### DIFF
--- a/docs-site/editions/overview.mdx
+++ b/docs-site/editions/overview.mdx
@@ -77,6 +77,29 @@ sudo CLAWBOX_EDITION=hermes bash install.sh
 Valid values are `openclaw`, `hermes` and `dual`. An unrecognised value prints a
 warning and installs as `openclaw`.
 
+That variable chooses the edition on a **first** install. On a device that has
+already been installed, the installer compares it against the recorded edition
+and stops if they differ, before it changes anything:
+
+```text
+ERROR: this device is already installed as the 'hermes' edition.
+
+         Recorded edition:  hermes   (/etc/clawbox/edition.env)
+         Requested edition: openclaw
+```
+
+Re-running the installer with the edition the device already is — or with no
+`CLAWBOX_EDITION` at all, which reads the recorded value — is the normal
+re-install and update path and is unaffected. `dual` counts as its own edition
+here, so moving to or from it is a change like any other.
+
+The installer does not migrate an edition. It has no step that removes the
+harness the device would be leaving, and each harness keeps its AI provider
+sign-in in its own configuration file, so signing in does not carry across. A
+device installed over the top of another edition ends up running two agents with
+no usable model — which is why the transition is refused rather than half
+performed. Changing a device's edition is a reflash.
+
 ## Which edition am I on?
 
 Over SSH or in the Terminal app:
--- a/install.sh
+++ b/install.sh
@@ -106,12 +106,34 @@ _read_edition_from_file() {
   printf '%s' "$v"
 }
 
+# The edition RECORDED on this device, from the root-owned records only and in
+# authority order. Deliberately excludes config/edition.txt: that is a
+# first-flash seed in a tree the customer owns, so it is a source for resolution
+# (below) but never evidence of what the device already IS.
+_read_recorded_edition() {
+  local v
+  v="$(_read_edition_from_file "$CLAWBOX_EDITION_FILE" || true)"
+  [ -n "$v" ] || v="$(_read_edition_from_file "$LEGACY_EDITION_DROPIN" || true)"
+  printf '%s' "$v"
+}
+
+# Lower-case and map anything unrecognised onto openclaw — the same rule the
+# resolution `case` below applies, minus the warning. Empty stays empty so
+# "no edition recorded" is distinguishable from "recorded as openclaw".
+_normalise_edition() {
+  local v
+  v="$(printf '%s' "${1:-}" | tr '[:upper:]' '[:lower:]' | tr -d '[:space:]')"
+  case "$v" in
+    openclaw|hermes|dual|"") printf '%s' "$v" ;;
+    *) printf 'openclaw' ;;
+  esac
+}
+
+CLAWBOX_RECORDED_EDITION_RAW="$(_read_recorded_edition)"
+
 CLAWBOX_EDITION_RAW="${CLAWBOX_EDITION:-}"
 if [ -z "$CLAWBOX_EDITION_RAW" ]; then
-  CLAWBOX_EDITION_RAW="$(_read_edition_from_file "$CLAWBOX_EDITION_FILE" || true)"
-fi
-if [ -z "$CLAWBOX_EDITION_RAW" ]; then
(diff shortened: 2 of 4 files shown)


KrasimirKralev:
All three findings were valid and are fixed in `4c18a3c`.

**1. Legacy drop-in in `setup-hermes-edition.sh`** — correct, and the most important of the three. It resolved the recorded edition from `/etc/clawbox/edition.env` only, so a device provisioned before that file existed had an empty recorded value and a conflicting `CLAWBOX_EDITION` walked straight past the refusal. That is exactly the gap `install.sh` already covers, left open in the second writer. It now falls back to the drop-in using the same two-shape parser (`CLAWBOX_EDITION=x` and `Environment=CLAWBOX_EDITION=x`).

**2. Unrecognised recorded value** — correct that the two writers disagreed, and fixed, but **not** by treating the value as absent. `install.sh` maps an unrecognised recorded edition to `openclaw`, and `setup-hermes-edition.sh` now does the same. Treating it as absent is the dangerous reading: it would clear the refusal entirely and let a typo'd lock be provisioned straight over, which is the failure this PR exists to prevent. Mapping to `openclaw` keeps both writers agreeing on what the device is, and still refuses a genuine switch away from it.

**3. Whitespace alignment in `install.sh`** — correct. The recorded value was whitespace-stripped and the resolution was not, so a padded `CLAWBOX_EDITION=" hermes "` resolved to `openclaw` (unrecognised → warn → fall back) while the lock still read `hermes`. That refused on a device nobody was trying to change, and the message reported `Requested edition: openclaw` at an operator who had typed `hermes`. The two rules are now identical.

New tests cover each: the drop-in path, the typo'd lock in both directions, a padded value that matches, and a padded value that genuinely differs. 47 tests in the file, full suite 2462 green.

PR #386, opened 2026-08-12
fix: make hermes dashboard-auth failures honest and propagate them

KrasimirKralev:
## Why

A Hermes provision could report itself broken and still be indistinguishable from a healthy one. Two real provisions on different boards printed `dashboard auth setup returned non-zero` and `Hermes is not fully provisioned`, then `All 20 checks healthy`, then the flash host printed `Setup: 1/1 succeeded`. The credentials written were always correct — the *check* failed at install time.

The cause is a lost update on `~/.hermes/config.yaml`, which has several writers, each doing an unlocked read-modify-write. `install.sh`'s `step_start_services` restarts `clawbox-setup`; `production-server.js` fire-and-forgets `register-mcp.sh`; thirteen lines later `step_hermes_edition` runs the dashboard-auth script. The auth script's critical section is *write config → write password file → re-read to verify*. A writer that snapshotted before the write and lands its `os.replace` inside that window erases the dashboard block, and the verify then reports a credential fault for credentials that are provably correct.

## What

**Serialise the writers.** One `flock` over the config, defined once in `scripts/lib/hermes-config-lock.sh` and sourced by both shell writers. The lock path is derived from the config path *after* canonicalising it, so two spellings of one file (a symlinked home, a `..`) cannot produce two lock files and quietly stop excluding each other.

A lock rather than ordering the install steps, because ordering fixes install time only: the same two writers race again on every boot from two independent units (`clawbox-setup` and `clawbox-hermes-dashboard`), where there is no single caller to order. The Hermes CLI is a writer we do not control and cannot ask to cooperate — it is covered by never being allowed to run outside a critical section: `register-mcp.sh` holds the lock across its invocation, and the Settings routes' `hermes config set`/`unset` now run under the same lock via `flock(1)`, applied at the one chokepoint (`runHermesCli`) so all five call sites are covered. Read-only calls stay unlocked; a lock conflict is a bounded 30s wait and a distinct exit code, so "the device is busy" is not reported as "the command failed".

**Both phases of the race are closed.** Phase one — a writer landing between the write and the verify — makes the verify fail loudly, and is retried. Phase two is the dangerous half: a writer landing *after* the verify left the script printing `done` and exiting 0 over a config with no dashboard block at all, a clean success reported over a box with no auth provider. Holding the lock until exit closes it for cooperating writers; a run that could not take the lock at all now exits 8 rather than reporting success. The work is still done first, so the dashboard keeps its auth provider and only the certification is withheld.

**Honest failure classes** (`scripts/setup-hermes-dashboard-auth.sh`). The old check returned a bare exit 1 for a genuine mismatch, a config another writer had rewritten, an unreadable file, and an interpreter without `scrypt` alike — and the message blamed the credentials. Each is now its own outcome with its own message and exit code: "could not run the check", "the block is missing", "the hash does not match the password", "verified but not serialised". Local write failures are reported as local write failures, not blamed on a competing writer. The classifier is exposed as a read-only `--check` reused by validation, so the validator and the provisioner agree on what healthy means.

**Failures propagate** (`install.sh`). A non-fatal `step_hermes_edition` failure never reached the summary. The step stays non-fatal (a half-provisioned box should still come up reachable), but the failure is recorded, the summary prints `PROVISIONING INCOMPLETE`, a `provision-status` marker is written, and the script exits non-zero — so the flash host cannot report success over it. `step_validate_services` now verifies the auth provider directly instead of accepting a probe that answers the same whether or not the provider works. The exit code, the `[provision-status]` line and the marker file are kept in agreement: a marker that cannot be rewritten is removed rather than left behind asserting a stale `STATUS=ok`.

## Tests

Full unit suite green: 2437 passing across 176 files.

Phase two was reproduced deterministically before the fix — shim a `flock` that cannot acquire, and fire one competing read-modify-write the instant the verify returns: 8/8 trials exited 0 with no dashboard block on disk. After the fix: 0/8, all honest non-zero.

New regression coverage: the lock is defined once and both writers source it; one lock file from a symlinked or dotted path; a held lock is waited on rather than raced through; an unserialised run does not exit 0; the block being erased right after the verify never yields success; a compute failure is not reported as a mismatch; a genuine mismatch is still caught; a vanished block is classified as not-configured rather than a credential fault; an unwritable password file is not blamed on another writer; a failed provisioning step does not produce a success summary; a marker that cannot be rewritten is not left stale; config writes are locked and reads are not.

Two bash behaviours were verified against the shell rather than assumed, and both changed the implementation: a redirection failure on a compound command does not propagate through `if ! { ...; } > file` (so that guard would have been inert), and `flock -w -E` returns the chosen code only on timeout, the command's own status otherwise.

## Follow-up

On-device confirmation on the Hermes box is still owed (`sudo bash ~/clawbox/install.sh --step hermes_edition` plus the `--check` probe); that box is disconnected at the moment.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -61,6 +61,37 @@ PROJECT_DIR="/home/clawbox/clawbox"
 CLAWBOX_USER="clawbox"
 CLAWBOX_HOME="/home/clawbox"
 
+# ── Provisioning-status signal (read by the flash host) ──────────────────────
+# A full install keeps some steps NON-FATAL on purpose — a half-provisioned box
+# should still finish and come up reachable rather than abort mid-run. But
+# "non-fatal" must never become "invisible": a step that reported errors has to
+# reach the operator's summary AND the caller's exit status, or a flash host
+# prints "Setup: 1/1 succeeded" over an install that told itself it was broken.
+# So every non-fatal failure is recorded here, the summary lists them, install.sh
+# exits non-zero, and a machine-readable marker is left for the flash host.
+PROVISION_FAILURES=()
+PROVISION_STATUS_FILE="${CLAWBOX_PROVISION_STATUS_FILE:-/etc/clawbox/provision-status}"
+
+record_provision_failure() {
+  PROVISION_FAILURES+=("$1")
+}
+
+# Persist the final provisioning verdict where the flash host (or an operator,
+# or the next update) can read it without re-parsing install.sh's stdout.
+write_provision_status() {
+  local status="$1"; shift
+  local dir
+  dir="$(dirname "$PROVISION_STATUS_FILE")"
+  mkdir -p "$dir" 2>/dev/null || true
+  {
+    echo "# Written by install.sh at the end of a full install. Machine-readable."
+    echo "STATUS=$status"
+    echo "FAILED_STEPS=$*"
+    echo "TIMESTAMP=$(date -u +%Y-%m-%dT%H:%M:%SZ 2>/dev/null || true)"
+  } > "$PROVISION_STATUS_FILE" 2>/dev/null || true
+  chmod 644 "$PROVISION_STATUS_FILE" 2>/dev/null || true
+}
+
 # ── Edition (single-harness lock) ────────────────────────────────────────────
 # openclaw | hermes | dual. "openclaw" is the native product (single, locked),
 # "hermes" is its own SKU, "dual" is premium (both harnesses + the runtime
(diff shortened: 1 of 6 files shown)


KrasimirKralev:
Worked through all seven review comments. Six applied, one declined with a reason — plus a root-cause gap the review did not raise that I think was the important one.

### The gap: the race was closed for cooperating writers only

`flock` serialised the two shell writers, but both fell back to "proceed without the lock" and still exited 0. That left the *second* phase of the lost update alive: a writer landing after the verify, where the script prints `done` and exits 0 over a config with no dashboard block at all — a clean success over a box with no auth provider.

Reproduced deterministically (shim a `flock` that cannot acquire, and fire one competing read-modify-write the instant the verify returns): **8/8 trials exited 0 with no dashboard block on disk.** After the change: **0/8**, all honest non-zero.

Three things followed:

1. **`exit 0` now has one meaning** — the invariant was verified *and* the verification was serialised. A run that could not take the lock exits `8` (`CREDS_UNSERIALISED`), distinct from a mismatch (3) and an environment failure (4). The work is still done first, so the dashboard keeps its auth provider and only the *certification* is withheld — `ExecStartPre` ignores the status via its `-` prefix, so nothing is left unable to start.
2. **The third writer participates.** `hermes config set`/`unset`, run from the Settings routes, is a read-modify-write of the same file through the CLI's own load/save. It cannot be asked to take our lock, so it now runs *under* it via `flock(1)` at the one chokepoint (`runHermesCli`), which covers all five call sites. Bounded wait (30s, so a UI action fails instead of hanging) and `-E 75` so "the device is busy" is distinguishable from "the command failed". Read-only calls stay unlocked. Verified `flock -w -E` against the real binary: 75 only on timeout, the command's own status otherwise.
3. **Why a lock and not step ordering** — ordering would fix install time only. The same two writers race again on every boot from two independent units (`clawbox-setup` and `clawbox-hermes-dashboard`), where there is no single caller to order. The Hermes CLI is covered not by cooperating but by never being allowed to run outside a critical section — `register-mcp.sh` holds the lock across its invocation.

### Applied

- **One lock helper, not two copies** — moved to `scripts/lib/hermes-config-lock.sh` and sourced by both writers. Agreed this was the sharpest of the seven: the exclusion held only while the copies stayed byte-identical, and every test would still have passed after an edit to one. Also canonicalises the config path before deriving the lock, so two spellings of one file (symlinked home, `..`) cannot yield two lock files. A missing library is refused rather than silently run unserialised.
- **`mint_credentials` write failures** — correct and important. It is always called as `mint_credentials || rc=$?`, which disables `set -e` for the whole body, so the unchecked `mkdir`/write returned 0 and the verify then reported "a concurrent process rewrote the file" for what was a full disk or a bad permission. Now returns 1 with a message that says local write failure, not concurrent writer.
- **Stale `STATUS=ok` marker** — fixed, with one correction to the suggested patch. `if ! { ...; } > "$FILE" 2>/dev/null; then` does **not** detect the failure: a redirection failure on a *compound* command does not propagate through `if !` (verified on bash 5.1 — the branch is simply never taken, so the guard would have been inert). The write is now a subshell used as the `if` condition, which detects it and also keeps the shell's own "Permission denied" off the terminal, since `> file` is opened before `2>/dev/null` takes effect. On failure the stale marker is removed rather than left contradicting the exit code.
- **Test robustness** — marker lookups guarded before slicing; lock timing measured from Node rather than `date`/`awk`; the `scryptHash` fixture asserts it actually built, so a python failure can no longer seed `password_hash: ""` and fail the mismatch test pointing at the classifier.

### Declined

**Setting `HERMES_CONFIG` in `clawbox-setup.service` after the `EnvironmentFile=` entries.** The proposed mechanism does not work: `EnvironmentFile=` overrides `Environment=` regardless of position — which is exactly why this repo puts `/etc/clawbox/edition.env` last among the `EnvironmentFile=` lines rather than writing it as an `Environment=` line, and the unit file documents that. An `Environment=HERMES_CONFIG=` would be overridden by any `EnvironmentFile` that set it.

The underlying concern is real but narrower than stated. The two writers only need to agree when they are writing the *same file* — if `HERMES_CONFIG` genuinely points elsewhere, they are not racing, they are editing different files, and there is no shared resource to protect. The actual hazard is two *different spellings* of one file producing two lock files, which is what the canonicalisation in the shared library fixes, with a test covering the symlink and `..` cases.

### Tests

Full unit suite green: **2437 passing across 176 files** (one worker-startup timeout under WSL, confirmed passing standalone). New coverage: the lock is defined once and both writers source it; one lock file from a symlinked/dotted path; an unserialised run does not exit 0; the block being erased right after the verify never yields success; an unwritable password file is not blamed on another writer; a marker that cannot be rewritten is removed rather than left stale; config writes are locked and reads are not.

On-device confirmation on the Hermes box is still owed — that box is disconnected at the moment.

KrasimirKralev:
Heads-up on timing: this merged at `66f7d5a` while the follow-up above was still in progress, so **the work described in my previous comment is not in beta.**

`fix/hermes-dashboard-auth-trust` now carries one commit that the merge did not include:

```
158667e  fix: close the config-write race and refuse to certify an unserialised run
```

It is a clean descendant of the merged head and merges into current beta (`96d7957`) without conflicts. It is the part that closes the exit-0-over-a-broken-box phase of the race, plus the six applied review findings.

I have not opened a new PR for it and have not pushed to beta — say which you'd prefer and I'll land it.

KrasimirKralev:
One more commit on the branch — a regression I introduced and caught before it could ship:

```
26e70c7  fix: keep the read-only auth check free of filesystem side effects
```

Deriving the lock path called `mkdir -p` on the config directory. `--check` is read-only and install.sh's validator runs it **as root**, so on a box where `~/.hermes` did not exist yet, root would create it and the clawbox user could no longer write the 0600 config it owns — the exact failure the auth script's own header warns about. Now canonicalised with `readlink -m` (canonicalise-missing), which creates nothing; the directory is created in `acquire_config_lock`, which only writers reach and which runs as clawbox. Pinned by a test asserting `--check` leaves its root empty.

Branch is now 2 commits ahead of beta, still conflict-free. Full unit suite: **2448 passing across 177 files.**

PR #409, opened 2026-08-20
A ClawBox can finally see the image you attach to it

yalexx:
Fixes TASK-417: attaching an image in device chat was accepted, and then the assistant answered that it could not see it.

Two independent defects, both reproduced on test box `192.168.50.65` before any code was written, and both of which have to go before the customer-visible behaviour changes.

### 1. No vision model was registered

Both ClawBox AI chat tiers are `input: ["text"]`, so OpenClaw does not inline image parts — it hands the turn a media path and expects the `image` tool to describe it. That tool resolves its model from `agents.defaults.imageModel`, a key ClawBox provisioning never wrote, so `runWithImageModelFallback` throws *"No image model configured"* (`dist/model-fallback-CvSRhgYr.js`, 2026.7.1).

A vision entry (`gpt-5.6-luna`, `input: ["text","image"]`) now rides on the ClawBox AI provider, and `agents.defaults.imageModel` points at it.

**Why the `deepseek` provider and not `openai`:** that entry *is* the ClawBox AI proxy — it already carries `api: "openai-completions"`, the proxy `baseUrl` and the `claw_` token, which is exactly the transport a vision request needs. OpenClaw's `openai` provider defaults to `openai-responses`, which the proxy does not speak. It cannot leak into the chat model picker: the device catalogue for `clawai` is the hardcoded two-entry `CLAWAI_STATIC_MODELS`, not a read of `models.providers.deepseek.models`.

Written in **both** places, because a box already in the field never re-runs the configure route: the provisioning route for new boxes, `scripts/gateway-pre-start.sh` for the ones already shipped. Same don't-clobber rule as image generation (an owner-chosen slot, including a fallbacks-only one, is left alone), and non-fatal for the same reason — a chat provider that works is worth more than a vision model.

### 2. The attachment landed where OpenClaw will not read it

The composer uploaded to `/setup-api/files?dir=uploads`, which writes to `$HOME/uploads`, then named that absolute path in the message. OpenClaw reads local media only from a fixed root allowlist (`buildMediaLocalRoots`: the openclaw tmp dir, `<configDir>/media`, `<stateDir>/{media,canvas,workspace,sandboxes}`, plus the agent workspace). `$HOME/uploads` is on none of them, so the `image` tool answered *"Local media path is not under an allowed directory"*.

Chat attachments now stage under `<stateDir>/media/chat-attachments`, which is on that allowlist, through `POST /setup-api/chat/attachments`. Deliberately a route of its own rather than loosening `file-guard.ts` over `~/.openclaw` — that tree also holds `openclaw.json`, the identity keys and every transcript. It inherits the existing `/setup-api/chat` middleware gating, including the pre-setup AP window, and returns the same `{ ok, name, path }` shape, so only the URL changed on the client.

### Measured, not guessed

Against the live proxy from a device on 2026-08-21: `max_tokens: 128000` is accepted for the vision model; `200000` (the generic default an entry falls through to when the field is absent, since a configured provider overrides the bundled catalog) comes back `400 "max_tokens is too large … supports at most 128000 completion tokens"`. The media-understanding path in 2026.7.1 happens not to send `max_tokens` at all — verified on a real box, the describe call succeeds with the field removed — so `maxTokens` here is a guard against a caller that starts sending it, not the thing that makes vision work today.

### Tests

32 new tests across four files, every one verified to fail without the change:

- `src/tests/unit/gateway-pre-start-clawai-vision.test.ts` (13) — runs the migration block **out of the shipped `.sh`**, mocking only the four names it reads from its surrounding scope. Covers idempotence, the token entitlement gate, the don't-clobber rule including fallbacks-only, and each individual repair.
- `src/tests/routes/ai-models/configure-vision.test.ts` (9) — the exact `openclaw config set` commands the provisioning route runs.
- `src/tests/routes/chat-attachments.test.ts` (8) — the staging route against a real temp filesystem, including traversal names, dotfiles, and no-partial-file-on-reject.
- `src/tests/components/chat-attachment-staging.test.tsx` (2) — drives `ChatPopup` through a scripted gateway socket: a pasted image goes to the staging route and never to the Files API, and the path the agent is handed is the staged one.

Two existing assertions were narrowed rather than deleted, because the invariant they stated genuinely changed: `configure.test.ts` asserted every `models[]` row is a 1M/text V4 chat tier (now scoped to the two tiers), and `configure-images.test.ts` asserted `imageModel` is never written (now asserts the two keys do not alias).

Local gate: **228 files / 3049 tests green**, baseline 224 / 3017.

### Behaviour on a real box

Before: the `image` tool call is rejected on the path, and the model flails through `read` / `clawbox__read_file` / `exec` workarounds — it happened to reach a correct answer by measuring pixels with Python, which is not a feature.

After: **one** `image` tool call, first try, `"model": "deepseek/gpt-5.6-luna"`, correct description of the actual picture.

## Summary by CodeRabbit

* **New Features**
  * Added ClawBox AI vision-model support for image understanding with text and image inputs.
  * Added secure chat attachment uploads with validation, size limits, cleanup, and media-directory storage.
  * Updated chat attachments to use the supported media location.

* **Bug Fixes**
  * Preserved existing image and vision-model selections during setup.
  * Improved recovery of incomplete vision-model settings.
  * Made vision-model setup failures non-blocking for chat setup.

* **Tests**
  * Added coverage for vision configuration, attachment uploads, validation, migration, and configuration preservation.

### DIFF
--- a/scripts/gateway-pre-start.sh
+++ b/scripts/gateway-pre-start.sh
@@ -490,7 +490,11 @@ if isinstance(deepseek_provider, dict) and deepseek_provider.get("baseUrl") in (
 # max_tokens 128000 is accepted and 200000 (the generic default an entry falls
 # through to when the field is absent) comes back 400 "supports at most 128000
 # completion tokens".
-CLAWBOX_VISION_MODEL_ID = "gpt-5.6-luna"
+# Honour the same env override CLAWBOX_AI_VISION_MODEL_ID gives the route, so a
+# box provisioned against a staging proxy with a different alias map is not
+# dragged back to the production slug at the next boot. Unset (the normal case)
+# means the shipped default, which the unit test pins to the TS constant.
+CLAWBOX_VISION_MODEL_ID = (os.environ.get("CLAWBOX_AI_VISION_MODEL_ID") or "").strip() or "gpt-5.6-luna"
 CLAWBOX_VISION_MODEL_NAME = "ClawBox AI Vision"
 CLAWBOX_VISION_MODEL_REF = "deepseek/" + CLAWBOX_VISION_MODEL_ID
 CLAWBOX_VISION_MAX_TOKENS = 128000
--- a/scripts/gateway-pre-start.sh
+++ b/scripts/gateway-pre-start.sh
@@ -460,6 +460,99 @@ if isinstance(deepseek_provider, dict) and deepseek_provider.get("baseUrl") in (
     deepseek_provider["baseUrl"] = "https://clawbox.com/api/ai"
     changed = True
 
+# Migration: ClawBox AI vision (image understanding).
+#
+# A ClawBox accepts an image attachment in chat and then cannot look at it.
+# Both ClawBox AI chat models are `input: ["text"]`, so OpenClaw does not
+# inline image parts; it hands the turn a media path and expects the `image`
+# tool to describe it. That tool resolves its model from
+# `agents.defaults.imageModel`, which ClawBox provisioning never wrote, so
+# runWithImageModelFallback throws "No image model configured"
+# (dist/model-fallback-CvSRhgYr.js on 2026.7.1). Reproduced on a real box on
+# 2026-08-21; see TASK-417.
+#
+# Boxes already in the field never re-run the configure route, so the repair
+# has to happen here. Mirrors buildClawboxAiProviderDefinition() and the
+# imageModel write in src/app/setup-api/ai-models/configure/route.ts; the two
+# must stay in step, and src/tests/unit/gateway-pre-start-clawai-vision.test.ts
+# runs these exact bytes out of the shipped .sh.
+#
+# Registered under the `deepseek` provider even though the id is an OpenAI one:
+# that entry IS the ClawBox AI proxy, already carrying api=openai-completions,
+# the proxy baseUrl and the claw_ token. OpenClaw's `openai` provider defaults
+# to openai-responses, which the proxy does not speak. It cannot show up in the
+# chat model picker — the clawai catalogue is the hardcoded CLAWAI_STATIC_MODELS
+# list, not a read of this array.
+#
+# The model id and the ceiling are duplicated from CLAWBOX_AI_VISION_* in
+# src/lib/clawbox-ai-models.ts because a shell migration cannot import them.
+# 128000 is measured, not guessed: against the live proxy on 2026-08-21,
+# max_tokens 128000 is accepted and 200000 (the generic default an entry falls
+# through to when the field is absent) comes back 400 "supports at most 128000
+# completion tokens".
+CLAWBOX_VISION_MODEL_ID = "gpt-5.6-luna"
+CLAWBOX_VISION_MODEL_NAME = "ClawBox AI Vision"
+CLAWBOX_VISION_MODEL_REF = "deepseek/" + CLAWBOX_VISION_MODEL_ID
+CLAWBOX_VISION_MAX_TOKENS = 128000
+
+# The token is the entitlement, exactly as for images: only a box that actually
+# has ClawBox AI gets a vision model pointed at the ClawBox AI proxy. Read here
+# rather than borrowing the image migration's `_clawai_token`, so this block
+# stays a self-contained slice its unit test can run out of the shipped .sh.
+_vision_models = deepseek_provider.get("models") if isinstance(deepseek_provider, dict) else None
+_vision_token = deepseek_provider.get("apiKey") if isinstance(deepseek_provider, dict) else None
+if isinstance(_vision_models, list) and isinstance(_vision_token, str) and _vision_token.startswith("claw_"):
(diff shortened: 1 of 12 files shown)


yalexx:
Round 3 — the outside-diff finding on `route.ts:164-175` was right, and both halves reproduced before the fix:

1. **Orphaned staged file after `fieldsLimit`.** Only visible with a chunked body. A single-buffer body lets the `file` event fire before the catch reads `written`, so cleanup ran by luck; delivering fields and file as separate chunks (as a real connection does) gave `400 Too many form fields` and then left `<uuid>-orphan.png` in the staging directory, under a path no response ever named.
2. **Worse than described: a source error mid-file did not slow the cleanup, it hung the route.** busboy and the file stream stayed alive, the write never settled, `await Promise.allSettled(pendingWrites)` never returned, and `POST` produced no response at all. Raced against a 4s timer in the repro, it timed out; it now answers 500.

Fixed in 1eca58e with one idempotent `fail` helper on every limit, parser, file, write and source path: unpipe + destroy the source, destroy the active file stream *with the error* (that is what settles the in-flight write so cleanup can unlink instead of waiting), destroy busboy, then reject. The bespoke teardown that only existed on the 413 path now routes through it too.

Two permanent tests added, both verified to fail without the change. 228 files / 3056 tests green; `tsc --noEmit` unchanged at 13 pre-existing errors in unrelated suites.

yalexx:
### Review rounds 4 and 5 — V4 Dev Loop (Claude Opus 5)

**Round 4 (`125bbd6`) — all four findings verified against the code first, all four valid, none dismissed.**

1. *Map busboy parser errors to 400.* Correct, and understated. Probing busboy 1.6.0 directly rather than trusting the docs showed the same client mistake arrives by two different paths: with no file part open the parser raises `Malformed part header` / `Unexpected end of form` **on the Busboy instance**, but a body truncated *while a file part is open* emits the identical error **on that file's stream**. Wrapping `busboy.on("error")` alone would have left the truncation half answering 500. The classifier now runs on both, and a dead connection mid-upload deliberately stays a 500 — with its own test, so the classifier cannot quietly grow to swallow it.
2. *Retention for the staging directory.* Correct. Nothing ever removed a staged attachment, and on a Jetson this directory shares a disk with `openclaw.json`, the identity keys and every transcript. Sweeps before staging: 7 days by age, then oldest-first under a 500 MB cap. Nothing younger than 60 s is ever touched, so a concurrent upload cannot be deleted out from under itself. Best-effort by construction, with a test proving an unreadable directory still lets a good upload through.
3. *Bounded polling instead of fixed sleeps.* Done.
4. *Cover the file and part limits.* Correct — busboy enforces both by going quiet, so an unhandled limit looks exactly like an ordinary short upload.

**Round 5 (`0fd5815`) — CodeRabbit caught a regression round 4 introduced, and was right.**

The polling helper returned as soon as one read came back empty. But the regressions those assertions exist for stage the orphan file *after* the response resolves — so the first read is empty, the helper returns at once, and the test passes while the orphan lands a moment later. It would have gone green on the exact leak it was written to catch; that is worse than the fixed sleep it replaced. The directory must now read empty continuously for a settling window, and any file appearing inside it resets the window. Verified rather than argued: against a directory that gets a file 120 ms after the first read, stop-on-empty returns clean and the settling version returns the orphan. The `EACCES` test also asserts the sweep actually ran, since a route that stopped sweeping entirely could not fail to read.

Local gate: 228 files / 3065 tests green. `tsc` unchanged at 13 pre-existing errors in unrelated suites.

**Hardware proof — real customer path on test box `192.168.50.65`, deployed through `force-update.sh`.**

Clipboard-pasted an image into device chat in the actual UI (VNC Chrome over CDP), not through the API:

- the composer POSTed to `/setup-api/chat/attachments` → `200`, path `/home/clawbox/.openclaw/media/chat-attachments/…-meridian.png` — inside OpenClaw's allowlisted media root, not `$HOME/uploads`;
- the attachment chip and preview rendered, served via `/setup-api/chat/media`;
- pasting the same filename twice produced two different UUID paths — the round-1 collision fix, live;
- the message reached the agent as `[Attached file: <allowlisted path>]`;
- the agent made exactly one `image` tool call with `details.model = deepseek/gpt-5.6-luna` and answered `MERIDIAN SEVEN` / `TANGERINE-4409` / `pink oval`. The test image carries text on purpose — those strings are only obtainable by actually looking at the picture.

Reproduced five times: fresh sessions and a long-lived one, at reasoning `off` and `medium`. Boot migration is idempotent on the device — the second gateway restart logged `Gateway config already correct, skipping write`.

One anomaly recorded rather than glossed over: the *first* attempt inside a pre-existing 65k-token chat session, the model answered "I can't view the image content in this session" without calling the tool at all. It did not reproduce — the same session then succeeded at `medium` and again at `off`, and every fresh session succeeded first try. Model behaviour, not this diff, but it is the customer-visible symptom this task is named after, so it is being carried on TASK-417 rather than dropped.

PR #432, opened 2026-08-22
fix: say "not on this edition" where Hermes has no OpenClaw feature

KrasimirKralev:
Two Settings surfaces told a Hermes customer that an OpenClaw-only feature was
broken, or fixable, when it was simply not part of the edition they bought.
Both were found by an edition audit and **both were validated live on Hermes
hardware today, against a merged build** — the Voice section renders the
unavailable card, and the Memory embeddings row reads "Not on this edition".

### Settings → Voice was a dead end

`SECTIONS` includes `voice` statically and `VoiceOutputPanel` rendered with no
edition check. Everything behind it is openclaw CLI work — `capability tts
convert` for Check, `config set messages.tts.provider` for Select — and the
Hermes SKU ships no openclaw binary, so Check failed with an empty reason after
burning its 120 s timeout and Select came back 409.

`/setup-api/tts` now reports `supportedOnEdition: false` on that edition (GET)
and refuses both actions up front (POST 409, before the body is read, so a
Check cannot wait out a process that never started). The panel renders an
unavailable card instead of pulsing its skeleton forever. This is the pattern
the codebase already uses for an edition-absent feature: ClawKeep reports the
same flag from `lib/clawkeep.ts` and renders its own `EditionUnavailableCard`.

### The Memory embeddings row read as a fault

`local-models/route.ts` swallowed the `openclaw memory status` result into
`null`, and the row rendered as `not-installed` with "No embedding model is
answering, so memory search cannot run." On Hermes that sends the customer
looking for a model to install that the SKU never had — and it is
indistinguishable from an OpenClaw box whose embedding provider really is down.

`EmbeddingProbe` gains a `supported` flag, the route sets it from
`openclawIsAbsent()` and skips the status call entirely on that edition, and
the row gets a distinct `not-on-this-edition` run state (neutral pill, not the
amber "look at this" one) with an honest sentence. It also drops
`managedBy: "clawkeep"`, which would have been a second dead end inside the
first.

## Scope change: the chat-side gaps moved to the harness stack

This PR originally carried three more fixes, all on the chat path: a Hermes
gate on Ctrl+V paste, a Hermes branch for the New Chat (+) button, and a
harness check in front of the sticky `sessions.patch`. **All three have been
removed from this PR**, because the harness abstraction / parity stack
(#434 and #439) supersedes them with capability-driven versions rather than
harness-name comparisons:

- **paste** → `caps.canAttachImages`. A blanket Hermes refusal is now the wrong
  answer: current Hermes supports `chat --image`, proven end-to-end today with
  a pasted PNG riding `--image`.
- **New Chat** → `HermesAdapter.resetSession()`, which also clears the durable
  transcript — something the version here did not do.
- **`sessions.patch`** → `shouldPatchSessionDefaults` on capabilities.

Those hunks also referenced `hermesSessionRef` / `hermesSentProviderRef` /
`hermesSentModelRef`, which the parity stack deletes, so keeping them here
would break the build the moment #439 lands. `shouldPatchSessionThinking` in
`lib/chat-reasoning.ts` and its test went with them — it existed only to serve
the dropped gate.

What is left is edition gating in Settings, which is orthogonal to the harness
abstraction and conflicts with nothing in it. The branch is rebuilt on current
`beta` and depends on neither #434 nor #439.

## Tests

| Finding | Test |
| --- | --- |
| Voice section | `components/voice-output-edition.test.tsx`, `routes/tts-edition.test.ts` |
| Embeddings row | `routes/local-models-edition.test.ts`, plus a case in `unit/local-models.test.ts` |

Each new test file was run against the unmodified `beta` source file it covers
and failed there; the openclaw-edition control case in each file passes on both,
so the tests are not merely asserting "hermes mode exists".

## Verification

Run on a Hermes device (Jetson, `CLAWBOX_EDITION=hermes`) in an isolated clone
of this branch with a hard-linked `node_modules`, not against the running
install:

- Touched suites: **4 files / 26 tests, all passing**.
- Full vitest suite: **277 files / 3811 tests, all passing**.
- `eslint` on every changed and new file: **clean, 0 errors and 0 warnings**.
- `tsc --noEmit` project-wide: **14 errors, byte-for-byte the same set as the
  `beta` baseline measured in a parallel clone** (all in unrelated test files),
  **none in any file this branch touches**.

## Not changed

Settings still lists the Voice row on Hermes; opening it now explains itself
rather than hiding, matching how ClawKeep behaves on the same edition. No new
user-facing strings needed translating — `VoiceOutputPanel` and
`LocalModelsPanel` carry literal copy today, so the 10-locale bundles are
untouched.

## Summary by CodeRabbit

* **New Features**
  * Local model status now distinguishes between unavailable and uninstalled features.
  * Embeddings support is clearly indicated, with explanatory status messaging.
  * Voice output shows a dedicated “Not available on this edition” state instead of loading indefinitely.
  * Text-to-speech requests return a consistent unavailable response when unsupported.

* **Tests**
  * Added coverage for edition-specific local model, embeddings, text-to-speech, and voice output behavior.


KrasimirKralev:
@coderabbitai review

KrasimirKralev:
@coderabbitai full review

PR #436, opened 2026-08-22
Hermes security hardening: pre-setup auth window, password oracle, root escalation, telemetry gating (TASK-443/444/445/446)

yalexx:
Task: TASK-443/444/445/446 (Hermes QA epic TASK-442) — reported by krasi@idrobots.com

Four security fixes that all sit on the same shared surface (`/setup-api/*` +
auth), so they land together. Both editions.

> **Rebased onto beta `0343573` on 2026-08-24** — see "Rebase notes" at the end.

Evidence: `findings/validate-round1.md` §TASK-443 (l.521), §444 (l.853),
§445 (l.896), §446 (l.1058).

---

## TASK-443 — the pre-setup trust window

The gate in front of `/setup-api/*` was a **deny-list** evaluated whenever
`setup_complete !== true`: any path nobody had remembered to name was served
with no credential. Over a ~100-route surface that grew holes, and it grew them
in the one window where the device is broadcasting an **open `ClawBox-Setup`
AP**. `setup/reset`, `update/run`, `system/power`, `install/run-step`,
`wifi/connect` and `ollama/pull` were all reachable from radio range — and on
2026-08-22T00:04Z an unauthenticated POST to `setup/reset` really did wipe the
QA box.

**Inverted to an allow-list** (`src/lib/setup-api-gate.ts`) holding only what
wizard steps 1–3 call. The default is 401; adding a route no longer changes the
security posture. Two further changes keep that list small enough to be honest:

- **The window is keyed on `password_configured`, not `setup_complete`.** A box
  with a password has an owner, even if its wizard never finished or its
  config.json lost a key. This is the sub-fix that actually closes the incident:
  the box that got wiped had a password and no `setup_complete`.
- **The initial password set mints the owner's session**, so steps 4–5 (AI
  models, Telegram, `setup/complete`) run authenticated and need no carve-out at
  all. The wizard **page** follows the same window, so a resumed wizard hits
  `/login?redirect=/setup` and comes back with a session.

A missing config.json is still first boot; a **corrupt** one now fails closed
instead of handing the window back.

Destructive handlers also check for themselves, via a deliberately
self-contained `@/lib/route-auth` (it reads config.json and `.session-secret`
directly, so a route test that mocks `config-store` can't mock the guard away).
`setup/reset` used to be `export async function POST()` — a handler that could
not have read a cookie if it wanted to.

| route | gated |
|---|---|
| `setup/reset`, `system/power`, `install/run-step`, `ollama/pull` | always — no bootstrap carve-out |
| `update/run`, `wifi/connect` | bootstrap only (steps 1–2 predate any password) |
| `system/credentials` | initial set open; **change** requires a session |

## TASK-444 — password oracle + lockout bypass

- `credentials/verify` answered in ~64 ms next to a login padded to 300 ms, and
  was throttled only by an in-memory per-`X-Forwarded-For` bucket the caller
  could reset by changing a header it controls. It now shares `/login-api`'s
  machinery outright — same `MIN_RESPONSE_MS` pad **on every exit path**
  (including the 400s), same persisted escalating lockout — and requires a
  session. It has no first-boot role.
- `system/credentials` consults **/etc/shadow** as well as the config flag, so
  drift between the two can't re-open the initial-set path (444a).
  `hasSystemPassword()` returns `null` when it can't tell, which is deliberately
  never read as "no password".
- **Login lockout:** every attempt is now counted against a capped `global`
  bucket *as well as* `cf:<ip>`. On a box serving plain HTTP with no proxy in
  front, `CF-Connecting-IP` is just a header the client picks, so keying only on
  it meant each new value minted a fresh, empty, un-capped bucket — the
  escalating lockout was one `-H 'CF-Connecting-IP: <random>'` away from
  irrelevant (confirmed live). The global bucket stays capped at
  `SHARED_BUCKET_MAX_LOCK_MS`, so **the owner still cannot be locked out**, and a
  correct password clears both. `X-Forwarded-For` is still never consulted.

## TASK-445 — root escalation

Every part of the root hand-off resolved back into a clawbox-writable directory:

- `/etc/sudoers.d/clawbox-ollama` granted NOPASSWD root on
  `/home/clawbox/clawbox/scripts/optimize-ollama.sh` — clawbox-owned, in
  clawbox-owned directories. Anything with clawbox-level code execution could
  overwrite the file and then ask sudo to run it. No wildcard, no race.
- The root-update unit `ExecStart`'d `/home/clawbox/clawbox/install.sh`, which
  install.sh hands back with `chown -R clawbox` on every root run.

Both now run from **`/usr/local/libexec/clawbox`**, installed root:root 0755
under root-owned dirs. The unit goes through a root-owned dispatcher
(`config/clawbox-root-step.sh`) that also **validates its instance name** — the
grant is `clawbox-root-update@*.service`, so the step name was unvalidated input
on the root side of the boundary.

install.sh's bootstrap (`git fetch` + `git reset --hard origin/<branch>` +
re-exec) ran on **every** `--step`; the journal shows it firing for `chpasswd`,
`set_hostname` and `validate_services`. A password change now neither reaches
the network nor mutates the source tree — only the update family gets
`CLAWBOX_ALLOW_SELF_UPDATE`. And `reset-failed *` / `start --no-block *` no
longer hand over the whole systemd unit namespace (every real caller passes a
`clawbox-*` unit).

`update/run` requires auth — see TASK-443.

## TASK-446 — unauth telemetry + catalog poison

Per the brief, I re-checked each route on the box first. **Already closed on a
provisioned device** — all seven telemetry routes plus `hermes/models` return
401 unauthenticated today, exactly as the validation found; the leak is strictly
the pre-setup window, which the TASK-443 inversion closes. Verified just now:

```
system/info 401  system/stats 401  wifi/status 401  kv 401
ai-models/status 401  telegram/status 401  gateway/health 401
hermes/models 401  hermes/models?refresh=1 401      setup/status 200
```

**Still open, and fixed here:**

- `setup/status` is public by design (`/login` and the desktop bootstrap read it
  before a session exists) **and** is served through the cloudflared tunnel, so
  it leaked `local_ai_provider`, `local_ai_model`, `ai_model_provider` and
  `telegram_configured` to anyone holding the URL. An unauthenticated caller now
  gets only the wizard's own progress.
- **Cache downgrade.** A dashboard timeout makes `buildPayload` fall back to the
  2-provider disk manifest, and `load()` installed that over a healthy
  47-provider cache *unconditionally* — one transient 8 s timeout and the device
  had apparently lost 45 providers. Payloads now carry a source rank; a worse one
  is refused while the cached one is inside its stale window, and the caller gets
  `degraded: "dashboard-unreachable"` instead of a quietly thinner catalogue.
- `?refresh=1` needs a session (it is a cache bust that fans out into a live call
  per provider, not a read). The plain GET stays reachable. An anonymous
  `?refresh=1` degrades to a read and is told so via `refreshDenied`.
- **`credentialPresent` vs `verified`** are now separate fields. Hermes'
  `authenticated` means "a key is set", never "the key works" (it is derived from
  `not is_skeleton`), and reading one as the other is why a bogus provider looked
  healthy right up until the first turn 403'd.

`gateway/health` stays on the bootstrap allow-list — UpdateStep polls it for
readiness before a password can exist, and it returns
`{available, port}`. Flagging it rather than breaking the wizard.

---

## Tests

`npm test` — **74 failed | 3622 passed | 54 skipped**, byte-identical to the
`origin/beta` baseline (`74 failed | 3595 passed`), same 6 files:
`chat-attachments`, `chat-media`, `chat-transcribe`, `files/path`, `files/route`,
`hermes-chat-failure-message`. **Zero new failures; +27 passing.**

`npx tsc --noEmit` — 17 errors, all pre-existing, all in `src/tests/*` files this
PR doesn't touch. `npx eslint` on every changed file — clean.

New/extended coverage, each of which fails against the old code:

- `middleware.test.ts` — the six routes gated even during bootstrap; the
  telemetry family gated; the window closing on `password_configured`; fail-closed
  on corrupt config; the `/setup` page gate (and that its prefix still can't leak
  `/setup-api` past the gate).
- `credentials-verify.test.ts` — rewritten: 401 without a session, right and wrong
  padded to the same floor, the 400 paths padded too, lockout survives a correct
  password, and rotating `CF-Connecting-IP` still hits the global cap.
- `system/credentials.test.ts` — session minted on first-boot set; unauthenticated
  **change** refused with nothing written.
- `hermes-model-options-downgrade.test.ts` — a failed refresh keeps the
  47-provider catalogue and reports `degraded`; a legitimate shrink still lands;
  cold start still serves the fallback.
- `models-refresh-auth.test.ts` — `?refresh=1` honoured for a session, downgraded
  for anonymous, plain GET untouched.
- `root-steps.test.ts` — the path-allowlist unit: TS/shell/install.sh lists pinned
  together, no NOPASSWD grant resolves under `/home/clawbox`, no `systemctl *`
  wildcard, `chpasswd`/`set_hostname` can't self-update.
- `setup/status.test.ts` — the trimmed anonymous payload, asserted field-by-field.

## Verification that needs the box (not covered by vitest)

The systemd/sudoers half of TASK-445 is install-time. After deploy:

```sh
sudo -l -U clawbox                      # no /home/clawbox target; no bare `*` unit specs
ls -la /usr/local/libexec/clawbox/       # root:root 0755, both scripts
systemctl cat clawbox-root-update@       # ExecStart -> /usr/local/libexec/clawbox/clawbox-root-step.sh
sudo /usr/local/libexec/clawbox/clawbox-root-step.sh nope_step   # exit 64, "step not permitted"
# change the password from Settings, then:
sudo journalctl -u 'clawbox-root-update@chpasswd.service' | grep -i bootstrap   # must be EMPTY
```

The last one is the TASK-445 regression that matters most: today that journal
line says `[bootstrap] Refreshing install.sh from origin/beta...` on a password
change.

## Notes for review

- **Behaviour change to flag:** `/setup` now redirects to `/login` once a
  password exists and the wizard is unfinished. That is what makes the allow-list
  small — steps 4–5 go through the session gate instead of a carve-out. A wizard
  resumed in the same browser is unaffected (the cookie is minted at step 3); one
  resumed in a fresh browser logs in first. `CLAWBOX_TEST_MODE` covers `/setup`
  too, so the e2e-install harness is unaffected.
- **Pre-existing bug surfaced, deliberately not fixed:** `clawkeep_install` is in
  the UI's step list but was never added to install.sh's `DISPATCH_STEPS`, so
  `install.sh --step clawkeep_install` already exits "Unknown step" — that button
  has never worked. Adding it would *enable* a root step, which is a product
  change, not a security fix. Documented in `root-steps.test.ts`.
- **Not mine, skipped:** `crit15c` (disabling Local AI leaves a dangling
  `model.provider: "clawlocal"`) is local-ai provider wiring — TASK-448/451
  territory.
- `src/app/setup-api/ai-models/configure/route.ts` is touched for one line (the
  `optimize-ollama.sh` path) to keep it matching the sudoers grant.

---

## Rebase notes (2026-08-24)

Replayed onto beta `0343573`, after the merges of #435 (MCP/tooling), #437
(observability), #440 (BG i18n), #442 (reasoning) and #444 (docs). Only four of
this branch's 38 files were touched on both sides, and only one of them
conflicted.

- **`src/middleware.ts` — the one real conflict.** beta had added
  `/setup-api/pets` and `/setup-api/mascot-lines/regenerate` to
  `PRE_AUTH_SENSITIVE_PREFIXES` / `_EXACT`; this branch deletes that deny-list
  and replaces it with the allow-list in `src/lib/setup-api-gate.ts`. Both
  additions are **subsumed by the inversion** — under an allow-list neither path
  is named, so both are 401 by default, which is what beta wanted and stronger
  than what beta wrote. beta's assertions for them are kept verbatim and still
  pass.
  - The one thing the inversion *would* have broken is beta's other intent: the
    bare `GET /setup-api/mascot-lines` must stay reachable on the wizard. It is
    now an **exact** entry in `BOOTSTRAP_ALLOWED_EXACT` — exact, so the
    `/regenerate` leaf next to it (a ~3.8 GB cold model load, up to three
    minutes, desktop-only caller) stays gated exactly as beta gated it.
- **`src/tests/middleware/middleware.test.ts`** — auto-merged; beta's new cases
  (`pets`, `pets/select`, `mascot-lines/regenerate`, and "still allows the
  mascot GET") all pass unchanged against the allow-list.
- **`install.sh`** and **`src/app/setup-api/ai-models/configure/route.ts`** —
  auto-merged, disjoint hunks. Re-checked by hand that the
  `/usr/local/libexec/clawbox/optimize-ollama.sh` sudo call and the
  `_clawbox_may_self_update` bootstrap gate both survived beta's rewrites intact.

One new commit on top of the three original ones: `setup/status.test.ts` typed
its imported handler as taking an optional `Request`, which stopped
typechecking once TASK-446 made the handler read the session cookie off that
argument. Split into a required-arg handler type and the optional-arg wrapper
the existing assertions call.

### Verification after the rebase

Run in a clean worktree of the rebased branch (`bun install --frozen-lockfile`),
against a clean worktree of `origin/beta` as the baseline:

| | beta `0343573` | this branch |
|---|---|---|
| `npx vitest run --config vitest.config.ts` | 310 files, **4429 passed, 0 failed** | 310 files, **4482 passed, 0 failed** |
| `npx tsc --noEmit` | 19 errors | **19 errors — the identical set**, diff empty |
| `bun run build` | — | exit 0 |
| `npx eslint <changed .ts/.tsx>` | — | 0 errors, 1 warning (`_api` unused in `ai-models/configure/route.ts`, **pre-existing on beta**, byte-identical) |

The +53 tests are this PR's own. The earlier "74 failed" baseline in the
Tests section above was an artefact of the old worktree's shared
`node_modules` symlink (node builtins failed to resolve); with a real install
both sides are fully green, so that section's numbers are superseded by this
table.

`src/tests/unit/run-tunnel.test.ts` flaked once across the three full runs
("keeps the URL history across the stop") and passes in isolation and on the
other two runs; it is a shell-script test this PR does not touch.

## Summary by CodeRabbit

* **Security**
  * Added session protection to setup, installation, reset, power, Wi‑Fi, update, and Ollama actions.
  * Restricted bootstrap access and privileged system operations to approved workflows.
  * Improved password protection with global and per-address lockouts and consistent response timing.
  * Limited sensitive setup details to authenticated sessions.

* **Reliability**
  * Preserved higher-quality Hermes model data when refreshes use degraded fallbacks.
  * Added clearer credential verification and degraded-status reporting.

* **Tests**
  * Expanded coverage for authentication, authorization, lockouts, bootstrap access, and privileged operations.

### DIFF
--- a/config/clawbox-root-step.sh
+++ b/config/clawbox-root-step.sh
@@ -1,0 +1,97 @@
+#!/usr/bin/env bash
+#
+# Root entrypoint for clawbox-root-update@<step>.service.
+#
+# Installed by install.sh to /usr/local/libexec/clawbox/clawbox-root-step.sh as
+# root:root 0755, under root-owned directories. That location matters: the unit
+# used to ExecStart /home/clawbox/clawbox/install.sh directly, and every path
+# component of that — the project dir, and install.sh itself — is writable by
+# the clawbox user the web server runs as. install.sh even hands the tree back
+# with `chown -R clawbox:clawbox` on every root run. So "clawbox may start
+# clawbox-root-update@*.service" meant "clawbox may edit the file root is about
+# to execute": a scoped NOPASSWD grant that is a one-step local root. TASK-445.
+#
+# This script cannot make install.sh itself immutable — the updater has to be
+# able to replace it — so it does the two things a root-owned entrypoint can:
+#
+#   1. Validates the instance name against its own allow-list. The sudoers grant
+#      is `clawbox-root-update@*.service`, so without this the step name is
+#      unvalidated input on the root side of the boundary.
+#
+#   2. Decides whether this step may self-update. install.sh's bootstrap block
+#      does `git fetch` + `git reset --hard origin/<branch>` + re-exec, and it
+#      ran on EVERY `--step` — including `chpasswd`. A password change must not
+#      reach out to the network, and must not be a way to pull new code onto the
+#      box. Only the update family gets CLAWBOX_ALLOW_SELF_UPDATE=1; everything
+#      else is pinned to the on-disk copy.
+#
+# Keep the two lists below in step with src/lib/root-steps.ts and install.sh's
+# DISPATCH_STEPS — src/tests/unit/root-steps.test.ts fails the build otherwise.
+
+set -euo pipefail
+
+PROJECT_DIR="/home/clawbox/clawbox"
+ENTRYPOINT="$PROJECT_DIR/install.sh"
+
+step="${1:-}"
+
+if [ -z "$step" ]; then
+  echo "clawbox-root-step: no step given" >&2
+  exit 64
+fi
+
+# Reject anything that isn't a plain step identifier before it reaches a case
+# match, a path, or a git ref.
+case "$step" in
+  *[!a-z0-9_]*)
+    echo "clawbox-root-step: refusing malformed step name: $step" >&2
+    exit 64
+    ;;
+esac
+
+# Every step the root-update template is allowed to run. Mirrors install.sh's
+# DISPATCH_STEPS.
+ALLOWED_STEPS="
+bootstrap_updater apt_update nvidia_jetpack performance_mode jtop_install
+ollama_install llamacpp_install llamacpp_model chromium_install ai_tools_install
+coding_harness vnc_install vnc_refresh openclaw_setup openclaw_install
+openclaw_patch openclaw_config openclaw_models openclaw_tts edition_lock
+edition_foreign_teardown hermes_install hermes_edition network_setup
+set_hostname setup_config system_config git_pull build rebuild rebuild_reboot
+restart restart_ap recover chpasswd gateway_setup ffmpeg_install polkit_rules
+systemd_services directories_permissions captive_portal_dns desktop_theme
+fix_git_perms browser_launch cloudflared_install nm_dispatcher sysctl_linkdown
+post_update update_smoke validate_services clawkeep_install
+"
+
+# Steps allowed to run install.sh's git fetch / reset --hard self-update.
+SELF_UPDATING_STEPS="
+bootstrap_updater git_pull build rebuild rebuild_reboot post_update update_smoke
+"
+
+contains() {
+  local needle="$1" haystack="$2" item
+  for item in $haystack; do
+    [ "$item" = "$needle" ] && return 0
+  done
+  return 1
+}
+
+if ! contains "$step" "$ALLOWED_STEPS"; then
+  echo "clawbox-root-step: step not permitted: $step" >&2
+  exit 64
+fi
+
+if [ ! -f "$ENTRYPOINT" ]; then
+  echo "clawbox-root-step: $ENTRYPOINT is missing" >&2
+  exit 66
+fi
+
+if contains "$step" "$SELF_UPDATING_STEPS"; then
+  export CLAWBOX_ALLOW_SELF_UPDATE=1
+else
+  # Pin this run to the on-disk copy: no fetch, no reset --hard, no re-exec.
+  export CLAWBOX_INSTALL_BOOTSTRAPPED=1
+fi
+
+exec /bin/bash "$ENTRYPOINT" --step "$step"
--- a/config/clawbox-root-update@.service
+++ b/config/clawbox-root-update@.service
@@ -11,7 +11,14 @@ EnvironmentFile=-/etc/clawbox/network.env
 # directly (it is the authority, not this environment), but exporting it here
 # keeps anything the steps shell out to consistent with the resolved edition.
 EnvironmentFile=-/etc/clawbox/edition.env
-ExecStart=/bin/bash /home/clawbox/clawbox/install.sh --step %i
+# Root-owned entrypoint, outside every clawbox-writable directory. Pointing
+# ExecStart straight at /home/clawbox/clawbox/install.sh meant the clawbox user
+# — who the web server, the in-UI terminal and the agent's shell all run as —
+# could edit the file root was about to execute, turning the scoped
+# `clawbox-root-update@*.service` sudoers grant into a one-step local root. The
+# dispatcher also validates %i (that grant accepts ANY instance name) and only
+# lets the update family run install.sh's git-fetch/reset self-update. TASK-445.
+ExecStart=/usr/local/libexec/clawbox/clawbox-root-step.sh %i
 # 30 min was not enough for llamacpp_install on a cold box: that step builds
 # llama.cpp from source with CUDA on a 6-core Jetson Orin AND downloads the
 # multi-GB Gemma 4 GGUF. systemd killed the unit mid-build, so "Provisioning
--- a/config/clawbox-sudoers
+++ b/config/clawbox-sudoers
@@ -74,8 +74,13 @@ clawbox ALL=(root) NOPASSWD: /usr/bin/systemctl enable  --now ollama.service
 clawbox ALL=(root) NOPASSWD: /usr/bin/systemctl disable --now ollama.service
 
 # Update / power / install paths.
-clawbox ALL=(root) NOPASSWD: /usr/bin/systemctl reset-failed *
-clawbox ALL=(root) NOPASSWD: /usr/bin/systemctl start --no-block *
+#
+# `reset-failed *` and `start --no-block *` used to take ANY unit name. Both are
+# only ever called with a clawbox-* unit (src/lib/updater.ts, the llamacpp
+# installer, the credentials chpasswd path, install/run-step), so scope them to
+# that prefix rather than handing over the whole unit namespace. TASK-445.
+clawbox ALL=(root) NOPASSWD: /usr/bin/systemctl reset-failed clawbox-*
+clawbox ALL=(root) NOPASSWD: /usr/bin/systemctl start --no-block clawbox-*
 clawbox ALL=(root) NOPASSWD: /usr/bin/systemctl start clawbox-root-update@*.service
 clawbox ALL=(root) NOPASSWD: /usr/bin/systemctl reboot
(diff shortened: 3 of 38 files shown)


KrasimirKralev:
TASK-527 coordination note — no action needed on your side.

527 was raised to lock `/setup-api/hermes/oauth/{start,submit,cancel}`, the inline provider sign-in relay: `start` opens a real OAuth session against the owner's dashboard and `submit` feeds it an authorization code, and under the old `PRE_AUTH_SENSITIVE_PREFIXES` deny-list neither was ever named, so both answered anyone in radio range of the open AP.

Your allow-list inversion already closed it. The family is not on `BOOTSTRAP_ALLOWED_PREFIXES`/`_EXACT`, so it is 401 by default pre-setup, and the session gate covers it after. Confirmed on the bench box: all four leaves plus the bare path 401 with no session, and all four work end to end against the live dashboard once you log in with the box password, so the wizard's own flow is intact. `poll` is gated too — it relays only `session_id`/`status`/`error_message`/`expires_at`, but under an allow-list the exemption would have to be added rather than removed and nothing needs it, since the id it polls can only have come from `start`.

So 527 shrank to verification plus the coverage that was missing, in #477. One thing worth flagging from it, since it is a property of your list rather than of the OAuth routes: the entries are prefixes, so a single plausible addition (`/setup-api/hermes`, or the oauth path itself, to read the provider list a step earlier) reopens all four leaves at once — and the assertion beta already had on the parent path `/setup-api/hermes/oauth` still passes when that happens. #477 names the leaves individually and adds a structural assertion against the list, verified by making that exact edit on the device and watching both guards fail.

Deliberately not done there: an in-handler `requireSession` on the three write routes, the way you gave the destructive handlers a second line via `@/lib/route-auth`. The argument for it is yours — one hand-maintained list in front of a ~100-route surface — and credential-minting is arguably in the same class as destructive. It is left as an explicit call for review rather than a second gate added on the quiet. If you think it belongs, that is a one-line change on top of #477.

PR #444, opened 2026-08-22
Hermes docs vs reality: safe factory-reset + remove non-Hermes claims (TASK-454)

yalexx:
Task: TASK-454 (Hermes QA epic TASK-442) — reported by krasi@idrobots.com

Docs-only. Every claim below was checked against the source tree in this branch and, where the question was live behaviour, read-only against the Hermes box at 192.168.50.71.

## 1. The SSH factory reset (the important one)

`support/recovery.mdx` promised a wipe of "AI keys, chats, Telegram pairing, ClawKeep pairing, saved Wi-Fi, Ollama models, Documents/Downloads/Desktop, SSH authorized keys". The command underneath it did almost none of that, and destroyed two things the UI reset deliberately keeps.

**It over-deleted.** `rm -rf /home/clawbox/clawbox/data/*` removes `network.env` and the multi-GB offline model weights under `data/llamacpp/models/`. `DATA_KEEP` in `src/app/setup-api/setup/reset/route.ts:22-47` keeps both, and the comment there says why: a reset box reboots into AP mode with no internet, so a deleted model cannot be re-downloaded.

**It under-deleted.** `~/.hermes` survived untouched — 1.1 GB on the box, holding `config.yaml` (billing token, dashboard session signing secret, scrypt password hash), `.env` (provider keys), `auth.json` (OAuth tokens), `memories/`, `sessions/`, `skills/`, `pairing/`. So did `~/.ssh/authorized_keys`, shell history, the browser profile, Ollama models and every saved Wi-Fi profile. A box "reset" that way is not safe to sell on.

**It stopped the wrong services.** `clawbox-gateway` is masked on Hermes (verified: `is-enabled` → `masked`, nothing on :18789), while `clawbox-hermes-dashboard`, its proxy and `clawbox-tunnel` were left running and free to rewrite `~/.hermes` between the `rm` and the reboot.

Section D is now a step-by-step procedure that mirrors the reset route, edition-aware, with the connectivity-destroying block (Wi-Fi profiles, password reset) **last** — the same ordering and the same reasoning as the route's own abort gate, so a failed wipe leaves the box reachable instead of stranded.

It also carries an explicit warning against the two obvious shortcuts. `rm -rf ~/.hermes` deletes `hermes-agent/`, which is the venv `~/.local/bin/hermes` execs; the shim survives the wipe, so the box is left with a `hermes` command whose interpreter is gone and `clawbox-hermes-dashboard.service` crash-looping on status 127. That is a reflash.

### How the commands were proven

The keep-list `find` invocations were run verbatim against a scratch tree reproducing the real layout. Result: `data/network.env` and `data/llamacpp/models/gemma.gguf` survive; `llamacpp/server.pid` and `server.log` do not; `~/.hermes/hermes-agent/venv/bin/python` survives; every other `~/.hermes` entry **including dotfiles like `.env`** is gone; `~/.openclaw` and `~/.clawkeep` are emptied including hidden entries. That is exactly `DATA_KEEP` + `HERMES_KEEP` + `removeDirectoryContents`.

The `nmcli` and `ollama list` parsers were tested against the box's real output (its wifi TYPE is `802-11-wireless`, not `wifi`; the `sed` form also survives an SSID containing a colon).

## 2. Claims the product does not have

| Claim | Reality |
|---|---|
| QR-card pairing / "pairing wizard" | No route consumes any pairing artefact. The only QR in the product opens `t.me/BotFather`, in the **optional, skippable** last wizard step. |
| VS Code desktop app | Not on *any* edition — component, API route and registry entry all deleted in 41fafc6. `code-server` is absent from the box. |
| ClawKeep on Hermes | `supportedOnEdition: getEdition() !== "hermes"` (`src/lib/clawkeep.ts:655`); the app renders an unavailable card. |
| `clawbox` CLI | No executable exists. `mcp/clawbox-cli.ts` is source-only, mode 100644, never linked or PATH-exposed. |
| `sudo clawbox update` | The wrapper elevates itself; a user `sudo` prefix runs it under root's PATH, where nothing named `clawbox` exists at all. |
| OpenAI API key on Hermes | `API_KEY_PROVIDERS` = openrouter, anthropic, gemini, zai, kimi-coding. OpenAI is sign-in (Codex) only. |
| "Sign in with your Claude account" | The row is `anthropic` — Hermes provider OAuth. There is no `claude-code` row anywhere in the catalog. |
| "No subscription = bring your own key" | The shipped tier table has a €0 "free forever" plan with managed DeepSeek V4 Flash and 1 GB backups. |
| DNS-hijack captive portal | `start-ap.sh:270` `rm -f`s the captive conf, `:271-283` writes upstream forwarding (8.8.8.8/8.8.4.4/1.1.1.1), `:315` *deletes* the port-80 DNAT rule. The box's dnsmasq dir holds only `upstream-dns.conf`. The middleware probe handling is real and is now described as its own separate layer. |
| `/setup-api/tunnel/*` | Dead as far as the UI goes — Remote Control drives `/setup-api/portal/{status,start,stop}`. |

The CLI invocation in the docs was checked end-to-end on the box: `bun run /home/clawbox/clawbox/mcp/clawbox-cli.ts edition` → `hermes`, and `system info` returns real data.

**OpenClaw gateway flows are edition-qualified, not deleted** — they are still true on openclaw/dual. Same for `technical/ai-providers.mdx` and the OpenClaw-only halves of `quick-reference.mdx`.

## 3. Things that were undocumented

- **Portal heartbeat** — destination, cadence, the exact three-field payload, and the two preconditions (paired token *and* a tunnel URL) that mean most boxes send nothing. Documenting the small footprint is a stronger privacy position than silence.
- **The device identity survives a factory reset.** `data/cloudflared/device-id` is deleted but re-derived from `/etc/machine-id`, so the box reappears in the portal as the same device. That is the fact a reseller most needs.
- **`~/.hermes/config.yaml` is generated state.** `hermes config set` (three calls per Local-AI save), `register-mcp.sh` at every boot and `setup-hermes-dashboard-auth.sh` all rewrite it whole, so its shipped comment blocks do not survive. Recorded as a caveat in `filesystem.mdx`.

## 4. Two adjacent unsafe instructions, corrected while editing

Not in the original scope, but both live in the recovery ladder and both are wrong in a way that costs the reader:

- Option C hardcoded `git reset --hard origin/main`, silently moving a `beta` box to a different release line. Now reads `.update-branch`.
- The installer step count said `[1/23] … [23/23]`; `TOTAL_STEPS=26`.

## Not addressed (out of scope)

`b6c` (four-level reasoning) is NOT-REPRODUCED — that sentence exists only on an abandoned branch, and the shipping "eight levels" claim is correct. Two *real* reasoning discrepancies turned up instead and are fixed: ClawBox AI offers seven (its backend rejects `ultra`), and the on-device model has a two-state thinking switch where the docs said it had no control at all.

Stale VS Code / `clawbox` CLI references also exist in `README.md`, `Clawbox.md` and `CLAUDE.md`. Left alone to keep this diff to `docs-site/`; worth a follow-up, particularly `Clawbox.md:148`, which tells the on-device agent it can open an app that does not exist.

## Verification

```
npm test          → 272 files, 3712 tests passed (unchanged; docs-only)
mint validate     → success build validation passed
mint broken-links → success no broken links found
```

Both `mint` checks were negative-controlled first — a deliberate broken link and a deliberate unclosed `<Warning>` each made them fail — so the passes are meaningful rather than the CLI finding no files. All internal links and heading anchors were additionally cross-checked with a script against Mintlify's slug rules.

No app code touched, so no vitest tests were added; the behavioural proof for the reset commands is the scratch-tree run described above.

## Summary by CodeRabbit

- **Documentation**
  - Updated first-boot and quickstart guidance for hotspot-based browser setup, network handoff, updates, and device access.
  - Clarified edition differences across OpenClaw, Hermes, and dual devices, including providers, authentication, reasoning controls, backups, services, and storage.
  - Added current subscription plan details, pricing, usage, models, backups, and support information.
  - Expanded recovery, factory-reset, updating, and CLI instructions.
  - Updated technical references for networking, architecture, filesystem paths, dashboards, credentials, and remote access.


yalexx:
## CI: the `test (test:coverage)` failure was a flake, not a regression from this PR

It is green now — [job 97060825529](https://github.com/ID-Robots/clawbox/actions/runs/32584040383/job/97060825529) passed on the **identical commit with no changes pushed**. Recording the diagnosis rather than just re-running, because the underlying race is real and someone should own it.

### What actually failed

One test, out of 3712:

```
FAIL src/tests/components/mascot-render-gate.test.tsx
     > falls back to the locale's own phrases when no hardcoded quote fits
AssertionError: "🦀! ✨" is not a Japanese line:
  expected [ '⚡ 最強のカニ', '👑 我こそが箱' ] to include '🦀! ✨'
  ❯ src/tests/components/mascot-render-gate.test.tsx:129:66
```

Coverage was **70.61 / 60.62 / 70.23 / 72.64** against thresholds of **63 / 52 / 61 / 65** — the gate passed with margin. The job exited 1 purely on that one assertion.

### Why it cannot be this PR

- The diff is **`docs-site/` only** — 18 `.mdx` files plus `llms.txt`. `git diff --name-only origin/beta...HEAD | grep -v '^docs-site/'` is empty.
- `vitest.config.ts` includes only `src/tests/**` for tests and `src/**/*.{ts,tsx}` for coverage. Nothing in the test job ever reads `docs-site/`.
- The failing assertion is in the mascot/i18n subsystem. This branch touches no mascot test, no `src/lib/mascot-packs/`, no `tamagotchi.ts`, no `Mascot.tsx`.

### The actual race

`"🦀! ✨"` is not a stray string — it is `"{name}! ✨"` from `nameGreetings` in `src/lib/mascot-packs/neutral.ts`, with `{name}` filled from that pack's `nameFallbacks` (`"🦀"`). The neutral pack's own header says what it is for:

> This is what the crab says before the per-locale pack has loaded.

The test sets `locale = "ja"`, then flushes exactly one microtask before firing the frenzy event:

```js
// Let the phrase fetch land before the frenzy picks its lines.
await act(async () => { await Promise.resolve(); });
```

If the phrase fetch resolves more than one microtask deep — which is load-dependent — the Japanese pack has not landed, the mascot correctly renders a neutral line, and the assertion that *every* bubble is in `JAPANESE_POWER` fails. The component behaved correctly; the test's wait is the weak link.

So it is a genuine flake with a specific, fixable cause: that `await Promise.resolve()` should be a `waitFor` on the Japanese pack actually being loaded, rather than a fixed-depth microtask flush.

### Reproduction attempts

25 local runs of the file, 10 of them under 3× parallel component-suite load to imitate a contended CI runner — **all passed**. It is rare: this was the only failure in the last 30 runs of the Tests workflow, across ~15 branches. That rarity is also why I am flagging it rather than treating it as a known-broken test.

### Not fixed here, deliberately

Fixing it means editing `src/tests/components/mascot-render-gate.test.tsx`, which would put a source change into a docs-only PR and make this diff harder to review. It also sits in mascot work that `feat/hermes-pets-mascot` appears to be actively touching, so a fix from here risks colliding. Happy to send it as a separate one-line PR if you want it — say the word.

PR #452, opened 2026-08-23
Installing a skill no longer leaves the chat frozen forever (TASK-508)

yalexx:
## What a customer hits today

Install any skill from the Store on beta `4c07d43` and the chat panel drops a **"Reloading skills… This may take up to 30 seconds"** overlay **that never clears**. Measured on the freshly flashed `.65`: stuck 5 min 26 s (10:05:05 → 10:10:31) before I stopped waiting. The chat is unusable, and only a full page reload recovers it — nothing on screen says so.

Found in the deep exploratory pass Yanko asked for on 2026-08-23. The Store is a headline v4 feature and this is the first thing anyone does with it.

## Root cause

`src/lib/openclaw-config.ts` deliberately removed the gateway restart on skill install. Its own comment explains why: SIGUSR1 means *restart* to OpenClaw, not *reload*, so every install was stopping the gateway; OpenClaw watches its skill roots anyway. That change is right and is untouched here.

`ChatPopup.tsx` was never updated. Its handler still said *"A skill install signals the gateway to restart (SIGUSR1), which drops the WS shortly after this event fires"* and deliberately did **not** force a reconnect, because it was waiting for the post-restart `hello` to clear the overlay. No restart → no dropped socket → no `hello` → the overlay's only exit never fires.

Evidence taken across an install on `.65`:

- `MainPID` 19583 before and after, `NRestarts=0`, `ActiveEnterTimestamp` 06:00:38 UTC while the install ran at 07:05 UTC
- `journalctl -u clawbox-gateway` — no entries in that window
- after a manual page reload, the **same running session** answered *"Yes. I see it listed in my available skills at `~/.openclaw/workspace/skills/weather-forecast/SKILL.md`"*

So the install path is fine. Only the overlay was wrong.

## The fix

The skill path stops waiting for something that will not happen: it keeps the socket, keeps the transcript, and sends one plain question through the same `startRun()` a typed message uses. The agent's reply is the confirmation the overlay was standing in for — and it comes from the session that actually has the skill.

Two things the freeze also swallowed come back: the owner is told the skill is live, and they see the question asked on their behalf rather than an answer appearing from nowhere.

Also corrected, because both were consequences of the same stale assumption:

- the hidden `[System: … your session was refreshed]` preamble — invisible in the bubble, false on the wire — is gone
- the confirm modal no longer warns that installing "will restart the current chat session" and lose unsaved context. It doesn't. Reworded in all nine locales, and `Install Anyway` is now just `Install`.

The skill-only half of the `hello` branch is **deleted**, not left unreachable. Two implementations of skill handling disagreeing with each other is how this bug happened.

## Tests

`src/tests/components/chat-skill-install-no-freeze.test.tsx` drives the real `clawbox-skill-installed` event against a fake gateway: no overlay in any of its three labels, the transcript survives, exactly one `chat.send` goes out naming the skill, the bubble is visible to the owner, the socket is **not** closed, and uninstall behaves the same. **Four of its five cases fail against current beta.**

`src/tests/unit/skill-change-message.test.ts` pins the wording and forbids "refresh"/"restart" from ever coming back into it.

Full suite: **282 files / 3962 tests green**. `tsc --noEmit` clean for every file touched. ESLint on the changed files: one warning fewer than beta, no new ones.

## Hardware

Re-verified after merge on **both** boxes — `.65` (clean install) and `.177` (upgrade path) — by re-running the exact reproduction. Results recorded on TASK-508 and in the TASK-379 log.

## Summary by CodeRabbit

- **New Features**
  - Skill installation, removal, enabling, and disabling now provide clear confirmation messages in the conversation.
  - Conversations remain active and retain their history while skills change.
  - Added localized installation messaging across supported languages.
  - Updated installation actions to use clearer “Install skill” wording.

- **Bug Fixes**
  - Prevented unnecessary reconnects, reload overlays, socket recreation, and transcript loss during skill updates.
  - Skill changes now complete without restarting the conversation, and installed skills are available immediately.


yalexx:
Fixed in `f0a1b2c`-equivalent (latest push) — thanks, this was a real hole.

One deviation from the proposed diff: the guard is in a wrapper, `enqueueRun`, rather than inside `startRun` itself. `startRun` is also what the queue drains **into** — the drain effect calls it as soon as `sending` clears — so putting the `if (sending)` check inside it would let the queue feed itself instead of emptying.

`enqueueRun` is the queue-or-run that `sendVoiceTranscript` already had; it is now shared by both callers rather than duplicated, and `sendVoiceTranscript` is just an alias for it.

Regression test added: fire the install event mid-turn, assert nothing extra goes out, assert the owner can already see the queued line, then end the turn with a real `chat` / `state: final` event and assert the confirmation goes out exactly once.

Suite: 282 files / 3963 tests green.

yalexx:
On the second finding (serialize skill-generated sends synchronously) — the mechanism is real, but I am deliberately not taking it in this PR, for three reasons:

**It is not something I introduced.** `enqueueRun` is `sendVoiceTranscript`'s existing body, moved and shared. The same stale-`sending` window has always been there for voice transcripts; this PR adds a second caller to it, it does not create the race.

**No customer action reaches it.** The event fires once per completed install, from `AppStore.tsx`, behind a confirmation modal and a network round-trip to ClawHub. Two of them cannot land in the same task — there is at minimum a modal click and an HTTP response between any two. I could not construct the failure outside a synthetic double-dispatch.

**The proposed remedy is a rewrite of the run bookkeeping**, not a guard: a synchronous active-run marker, keeping the gateway `runId` separately from `runIdRef` (which today holds the idempotency key), and filtering terminal events by run. That touches every send path in `ChatPopup` — typed, voice, queued, Hermes — in a file that is under on-device release verification right now. A half-correct version of it wedges the composer permanently, which is strictly worse than the bug it fixes: today's worst case is one duplicated question, tomorrow's would be a chat that never accepts input again.

Recorded as a follow-up on TASK-508 so it is not lost. If the team wants it, it deserves its own PR with the voice path in scope, not a rider on a freeze fix.

PR #453, opened 2026-08-23
The Coding Agent stops asking whether you trust your own box (TASK-509)

yalexx:
## What a customer hits today

Open the **Coding Agent**. Clear whatever it shows, get to the prompt, close the window, open it again — and it stops here, every single time:

```
Accessing workspace: /home/clawbox
Quick safety check: Is this a project you created or one you trust?
  1. Yes, I trust this folder
  2. No, exit
```

Found in the 2026-08-23 exploratory pass on beta `4c07d43`, on the freshly flashed `.65`. Both boxes carry the same state, so it is not a clean-install quirk.

## Why it repeats

Claude Code never persists the acceptance for `$HOME`. `~/.claude-ds/.claude.json` **is** written on exit — `lastGracefulShutdown: true`, `lastCost`, `lastSessionId` all update, and so does the file's mtime — but `projects["/home/clawbox"].hasTrustDialogAccepted` stays `false`. `.177` has been through the dialog twice (`projectOnboardingSeenCount: 2`) and is still `false`.

## Why pre-answering it is the right call here

On this appliance the question has no second answer. The folder is the box's own home, reached through the device login, and the alternative on offer is to quit. A security prompt that appears on every launch and can only be answered one way does not protect anybody — it trains the owner to click past prompts without reading them.

Claude Code names this key itself as the supported alternative: *"accept the trust dialog here once interactively, or set `projects[...].hasTrustDialogAccepted` in `<config>`"*.

**Scope is the whole point of the change.** It is seeded only when the harness starts in the clawbox user's home — the directory the Coding Agent app always opens in. Someone who clones a repository and runs `claude-ds` inside it is still asked, because that is the case the dialog exists for. Two tests hold that line.

## How timid the write is

- adds the one missing key, changes nothing else in the config
- **refuses to write over a config it cannot parse** — losing an owner's MCP servers to save them a keypress would be a bad trade
- no-ops when the answer is already there, so it does not rewrite the file on every launch
- temp file + `os.replace`, so a crash mid-write cannot leave half a config
- cannot stop the harness from starting: every failure path exits 0 and Claude Code launches anyway, with the dialog, exactly as today

## Tests

Six new cases in `claude-ds-wrapper.test.ts`, run against the **shipped** `scripts/claude-ds` with a fake home and a stub `claude` — the file's existing pattern. **Three fail against beta.** They cover: seeded for home, *not* seeded for a folder the owner was standing in, other config preserved (including another folder's `false`), unparseable config left untouched, launch survives an unwritable config directory, and no rewrite on the second launch.

Full suite: 282 files / 3969 tests green. `tsc --noEmit` and ESLint clean on the files touched.

## Hardware

The fix shape was proven on `.65` before any code was written — setting the flag by hand made the app open straight to the prompt, and the edit was reverted so the real fix can be proven from a clean state. Re-verified on both boxes after merge; evidence on TASK-509 and in the TASK-379 log.

## Summary by CodeRabbit

- **New Features**
  - Claude Code launched from the home directory now automatically remembers trust approval, avoiding repeated trust prompts.
  - Existing configuration and unrelated settings are preserved.
  - Launches continue normally if trust setup cannot be completed.
- **Bug Fixes**
  - Prevented unnecessary configuration rewrites when trust approval is already recorded.
  - Trust prompts remain unchanged when launching from directories other than the home directory.

### DIFF
--- a/scripts/claude-ds
+++ b/scripts/claude-ds
@@ -141,6 +141,76 @@ export CLAUDE_CODE_EFFORT_LEVEL="${CLAUDE_DS_EFFORT:-max}"
 mkdir -p "$CLAUDE_CONFIG_DIR"
 chmod 700 "$CLAUDE_CONFIG_DIR" 2>/dev/null || true
 
+# PRE-ANSWER THE TRUST QUESTION, FOR THE BOX'S OWN HOME AND NOTHING ELSE.
+#
+# Claude Code asks "is this a project you created or one you trust?" before it
+# will read a folder, and never remembers the answer for $HOME: the flag it
+# writes, projects[<dir>].hasTrustDialogAccepted, stays false there across
+# clean exits that do record lastCost and lastSessionId. So the Coding Agent
+# app — which always opens in the owner's home — asked it on EVERY launch.
+#
+# On this appliance that question has no second answer. The folder is the
+# box's own home, behind the device login, and the dialog's other option is
+# "exit". A security prompt whose only real answer is "yes", shown every single
+# time, does not protect anybody; it teaches the owner to click through
+# prompts without reading them, which is worse than not asking.
+#
+# It is pre-answered ONLY when the harness is starting in that home directory.
+# Somebody who clones a repo and runs `claude-ds` inside it is asked, exactly
+# as before — that is the case the dialog is actually for.
+#
+# Claude Code documents this key as the supported alternative to answering
+# interactively ("accept the trust dialog here once interactively, or set
+# projects[...].hasTrustDialogAccepted in <config>").
+seed_home_trust() {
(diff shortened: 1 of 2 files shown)


yalexx:
Both taken — thanks, they were the two weakest cases in the set.

**Seed-failure test:** you are right that  is a no-op under root, so the test would have passed without reaching the path it names. Skipped under uid 0 with the reason written into the test; CI and a developer machine both run the `unit` project unprivileged, so it still executes where it matters.

**No-rewrite test:** switched from mtime to a byte comparison, as suggested. The fixture is deliberately four-space indented — a rewrite through `json.dump` would flatten it, so the formatting itself is what fails the assertion.

Full suite still green.

PR #471, opened 2026-08-24
Remove the residual root escalation left after the sudoers narrowing (TASK-445)

yalexx:
## Task

**TASK-445 round 2 — root escalation residue after PR #436.**

The post-rebuild revalidation (`findings/revalidate-round1.md` §3 from line 399, "New findings" from line 498) confirmed the clawbox-writable-script root paths from #436 are fixed in the repo, but two things survived:

1. **[MEDIUM]** `/etc/sudoers.d/90-clawbox-nopasswd` still grants `clawbox ALL=(ALL) NOPASSWD: ALL` on the QA box. Not in the repo — a provisioning artefact — but sudo takes the **union** of every drop-in, so *every* narrowing in `config/clawbox-sudoers` is unobservable on a device that has it.
2. **[LOW]** the deployed bundle calls `sudo /usr/local/libexec/clawbox/optimize-ollama.sh` on a box where `/usr/local/libexec` does not exist, so saving a local Ollama model silently skips the q8_0 KV-cache / flash-attention tuning.

## Ruling

> narrow the shipped sudoers drop-in to an explicit allow-list of the exact root commands the product needs […] each as a full path with fixed args where possible, `NOPASSWD` only for that list; keep the box operable (setup wizard, updater, power, wifi, desktop toggle, factory reset all still work — trace every path). Ship `optimize-ollama.sh` under /usr/local/libexec/clawbox via the installer/updater with the same 0755 root-owned convention as the other helpers, or remove the dead call. Migration: the updater must replace the old sudoers file on existing boxes (validate with `visudo -c` before install; on failure keep the old file and log). Tests: a script that parses the allow-list and asserts every `sudo` invocation in the codebase is covered; shellcheck clean.

## Round 2 — why `e2e-install` failed, and the fix

The first push was green on 10 checks and **red on `e2e-install`** ([run 32758542809](https://github.com/ID-Robots/clawbox/actions/runs/32758542809)). `install.sh` itself completed; three assertions in the new `06-sudoers.spec.ts` failed. Two were the product, one was the test. Treated as a product bug, per the brief.

### 1. PRODUCT — the ollama grant never landed on a device that took the test-mode return

`config/sudoers-clawbox-ollama` was installed from `step_performance_mode`, **below that step's `is_test_mode` early return**:

```sh
step_performance_mode() {
  install_root_libexec
  if is_test_mode; then ... return 0; fi     # <-- returns here
  ...
  install_sudoers_dropin ".../sudoers-clawbox-ollama" clawbox-ollama   # <-- never reached
}
```

So the container installed cleanly, quarantined its blanket drop-in, got the narrowed allow-list — and had **no `optimize-ollama.sh` grant at all**. `sudo -l` proved it: 40 grants present, that one missing. On a real box the same shape is one `is_test_mode` away, and a grant that lives behind a Jetson-only step is a grant that silently does not exist on some devices — the same "the narrowing is invisible on the device" failure TASK-445 exists to close, one step removed.

It now installs from **`step_systemd_services`**, next to the primary drop-in: the one step a fresh install *and* the in-app updater (`step_post_update`) both run unconditionally.

**The quarantine stays gated on the PRIMARY allow-list only.** `config/clawbox-sudoers` is what keeps a box operable (wizard, updater, power, hotspot); the ollama grant is one feature's tuning. Gating the removal of a blanket passwordless-root drop-in on a missing KV-cache grant would be the wrong trade in the wrong direction, so the ollama install happens after the quarantine and only warns on failure.

### 2. PRODUCT — `step_ollama_install` still ran the clawbox-writable copy as root

`install.sh` runs as root throughout, and `step_ollama_install` still had `bash "$PROJECT_DIR/scripts/optimize-ollama.sh"` — a root path through a file under clawbox-writable `/home/clawbox/clawbox/scripts`. Exactly the defect #436 closed, missed because §2 of this PR only moved the `step_performance_mode` call site. It now runs `"$ROOT_LIBEXEC_DIR/optimize-ollama.sh"` like the other one, and a test asserts the repo copy is not executed from anywhere in `install.sh`.

### 3. TEST — the `(ALL : ALL) ALL` assertion encoded the wrong invariant

The spec asserted `sudo -l` never renders `(ALL : ALL) ALL`. That line is the **`sudo` group** rule, not a drop-in: `install.sh`'s `step_ensure_user` adds `clawbox` to `sudo`, `video`, `audio`, `i2c`, `gpio` on every device (`install.sh:863`), and the distro's `%sudo ALL=(ALL:ALL) ALL` **demands a password**. It is the owner's own administrator account, not a path the web server — which only ever runs `sudo -n` — can take. `quarantine_overbroad_sudoers` deliberately never touches it: stripping it would lock the only administrator out of an appliance with no console. So the assertion would have failed on a real box for the same reason it failed in the container.

Replaced with a **behavioural probe** — stronger than the string match and indifferent to how sudo renders its rules:

```ts
for (const probe of ["id -u", "cat /etc/shadow", "install -m 0755 /bin/true /usr/local/bin/pwn"]) {
  // no line in the allow-list names any of these; a blanket rule makes them all succeed
  expect(await sudoNonInteractive(probe)).toBe("DENIED");
}
```

The `NOPASSWD:\s*ALL` assertions and the per-file `/etc/sudoers.d` scan are unchanged.

### 4. The checker caught beta's new call site

The rebase onto current beta brought `src/lib/local-ai-runtime.ts` (#472), which calls `sudo -n /usr/bin/systemctl {enable --now|start|stop} ollama.service`. `check-sudoers-coverage.sh` failed in **both** directions, which is what it is for: the call spreads its argv (`["-n", ...argv]`) so it was **unresolved**, and the `start`/`stop ollama.service` grants it needs were reported **unused**. Declared in `DECLARED_ARGV` with the three literal module-level argv arrays; no grant was added or widened.

## The final allow-list (needs human eyes)

41 grants. Every one is a full path with fixed args, `(root)` runas only, no bare `ALL`. Both spellings of a unit ship because sudo matches argv exactly and `clawbox-gateway` / `clawbox-gateway.service` are different strings to it.

| Grant | What breaks without it |
|---|---|
| `systemctl restart clawbox-gateway{,.service}` | gateway restart after a config write, clawkeep restore |
| `systemctl stop clawbox-gateway{,.service}` | factory reset, updater |
| `systemctl --runtime {mask,unmask} clawbox-gateway{,.service}` | factory reset |
| `systemctl restart clawbox-setup{,.service}` | web-server restart (`force-update.sh`) |
| `systemctl {start,stop} clawbox-browser{,.service}` | Browser app |
| `systemctl {start,stop,restart,enable,disable} clawbox-tunnel{,.service}` | Remote Control |
| `systemctl {enable,disable} --now ollama.service` | Settings → Local Models |
| `systemctl {start,stop} ollama.service` | on-demand Local AI wake / idle stop (#472) |
| `systemctl reset-failed clawbox-*` | updater + installer hand-off |
| `systemctl start --no-block clawbox-*` | updater + installer hand-off |
| `systemctl start clawbox-root-update@*.service` | password change, hostname, AP restart, llama.cpp install |
| `systemctl reboot` / `systemctl poweroff` | power menu, factory reset |
| `apt-get update -qq`, `apt-get install -y -qq chromium{,-browser}`, `dpkg --configure -a`, `snap install chromium` | Browser app first-run install |
| `/usr/local/libexec/clawbox/clawbox-desktop-mode.sh --{enable,disable}` | Settings → Desktop |
| `/usr/local/libexec/clawbox/clawbox-power-mode.sh --{balanced,performance}` | Settings → Performance mode |
| `/usr/local/libexec/clawbox/optimize-ollama.sh` | saving a local Ollama model |

Two wildcards remain, both deliberate and both bounded by a root-owned target: `clawbox-*` (reset-failed / start --no-block) and `clawbox-root-update@*.service`, whose unit `ExecStart`s the root-owned `/usr/local/libexec/clawbox/clawbox-root-step.sh` — asserted by `06-sudoers.spec.ts`. `nmcli` has no grant: WiFi goes through polkit (`config/49-clawbox-updates.rules`).

`bash scripts/check-sudoers-coverage.sh --list` prints this table's source of truth plus all 42 resolved call sites.

## How verified (round 2)

```
$ bash scripts/check-sudoers-coverage.sh
check-sudoers-coverage: OK — 41 grants, 42 resolved sudo invocations, 0 gaps

$ npx vitest run --config vitest.config.ts
  Test Files  395 passed (395)
       Tests  5893 passed (5893)

$ bun run build   → succeeded, build-info.json commit=f7662e1
$ npx shellcheck install.sh scripts/check-sudoers-coverage.sh
    → checker clean; install.sh only pre-existing SC2329/SC1090 notes, none on changed lines
$ npx eslint e2e-install/06-sudoers.spec.ts src/tests/unit/install-sudoers-migration.test.ts → exit 0
$ bash -n install.sh → OK
```

The `e2e-install` harness was also run **locally in Docker** (`bunx playwright test --config e2e-install/playwright.config.ts --grep "root escalation surface"`) against a container `install.sh` actually provisioned — see the run's result below/in the checks.

Four new/changed unit tests pin the fix: both managed drop-ins install from `step_systemd_services`; **no** `install_sudoers_dropin` call exists outside it (byte-range assertion, so a future move into a conditional step fails the build); the quarantine precedes the ollama install; and neither `step_performance_mode` nor `step_ollama_install` runs the repo copy of `optimize-ollama.sh`.

### 1. The migration — `install.sh`

Shipping a narrow file was only half a fix. Three new functions, called from `step_systemd_services` (which fresh installs run directly and existing devices reach through the in-app updater → `step_post_update`):

- **`quarantine_overbroad_sudoers()`** moves any `/etc/sudoers.d` drop-in granting the clawbox service user a bare `NOPASSWD ALL` into `/var/lib/clawbox/sudoers-quarantine/<name>.<UTC timestamp>` (root-only dir 0700, file 0400) — kept so the removal is explainable and reversible rather than the file just vanishing.
  Deliberately narrow: **only** a rule whose user spec is `clawbox` or `%clawbox`. An operator's own `%sudo`/`%admin` rule and the distro default in `/etc/sudoers` are never inspected — removing one of those could lock the only administrator out of an appliance nowhere near a keyboard. If removing files makes `visudo -c` stop passing (a quarantined drop-in can define an alias another one uses), **every file goes back**.
- **`install_sudoers_dropin()`** replaces `cp → visudo -cf → rm + exit 1`. That order turned a typo in the repo into a device with **no** drop-in at all — every systemctl the web server needs failing on a password prompt nobody can answer, on a box with no console. It now stages the candidate **outside** `/etc/sudoers.d` (sudo parses everything in that directory, so an unvalidated file there is already live), validates it, installs atomically via `install(1)`, re-checks the whole set, and rolls the previous file back on failure. A bad candidate leaves the installed file untouched and logs why.
- **`sudoers_grants_blanket_nopasswd()`** — the detector. Handles line continuations, comments, `%clawbox`, `NOPASSWD:ALL` with no space, `(ALL:ALL)` runas, and `PASSWD:`-tagged entries.

**Order is asserted:** the allow-list is installed *first*, and the quarantine only runs if that succeeded — so the device is never left with neither.

### 2. `optimize-ollama.sh`

Already shipped root:root 0755 by `install_root_libexec()`, which both `step_systemd_services` and `step_resource_limits` call, so `step_post_update` reaches it on existing devices. This PR adds the missing guarantee: `step_performance_mode` now runs `"$ROOT_LIBEXEC_DIR/optimize-ollama.sh"` instead of the clawbox-writable repo copy. That is the copy the sudoers grant names, so running it during install *is* the check that `install_root_libexec` really put it there. A test enumerates every `/usr/local/libexec/clawbox/*` path any grant names and asserts `install_root_libexec` installs it and `scripts/<name>` exists.

### 3. The guard — `scripts/check-sudoers-coverage.sh` (new, `bun run check:sudoers`)

The allow-list only *stays* an allow-list if drift fails the build. The checker parses both drop-ins and every sudo invocation under `src/`, `mcp/`, `scripts/`, and fails in **both** directions:

- an invocation with no grant → the device would hit a password prompt;
- a grant with no invocation → privilege handed out for free (this is how a list creeps back toward `ALL`);
- **fail-closed**: a call site whose argv it cannot resolve is an *error*, not a pass. Dynamic argv must be declared in `DECLARED_ARGV` — keyed on the **call's source text**, so unrelated edits above it are fine but editing the call invalidates the declaration and forces a re-review — or exempted in `EXEMPT_CALLS` with a written reason.

It also rejects a bare `ALL` Cmnd, a non-`(root)` runas, and any line it cannot parse.

Three sites are exempt on purpose, all because the target lives somewhere clawbox can write and a NOPASSWD grant would recreate exactly the defect #436 closed: `clawbox update` → `sudo bash install.sh`, `runHermesCli({sudo:true})` → `~/.local/bin/hermes` (its `sudo -n` fails closed in ms; the only caller treats it as non-fatal), and `scripts/force-update.sh`'s operator-interactive `sudo -u` / `chown`.

### 4. Grants pruned (reverse direction found 10)

- `start` for `clawbox-gateway`, `clawbox-setup`, `clawbox-tunnel` (6 lines): every caller uses `restart`, which starts a stopped unit anyway. `clawbox-browser` keeps its `start` — the Browser app really calls it.
- `clawbox-ap` `restart`/`stop` (4 lines): Settings → Hotspot goes through `clawbox-root-update@restart_ap.service`, `stop-ap.sh` runs unprivileged, every other AP restart is inside `install.sh` as root.

`nmcli` needs no grant at all — WiFi runs through polkit (`config/49-clawbox-updates.rules`), not sudo.

## How verified

Every command run in the worktree at `f77552a`:

```
$ bash scripts/check-sudoers-coverage.sh
check-sudoers-coverage: OK — 39 grants, 39 resolved sudo invocations, 0 gaps

$ shellcheck --version → 0.10.0
$ shellcheck scripts/check-sudoers-coverage.sh          → exit 0
$ shellcheck <the new install.sh sudoers block>          → exit 0

$ visudo -cf config/clawbox-sudoers          → config/clawbox-sudoers: parsed OK
$ visudo -cf config/sudoers-clawbox-ollama   → config/sudoers-clawbox-ollama: parsed OK

$ npx vitest run --config vitest.config.ts
  Test Files  390 passed (390)
       Tests  5826 passed (5826)

$ bun run build   → succeeded, build-info.json commit=f77552a
$ npx eslint <the 4 changed/added TS files>  → exit 0
$ npx tsc --noEmit -p tsconfig.json          → 21 errors, byte-identical to the
                                                21 on origin/beta (pre-existing)
```

**40 new tests.**

`src/tests/unit/install-sudoers-migration.test.ts` (25) sources the real functions out of `install.sh` — never a copy — and runs them against a temp `/etc/sudoers.d` with fake root-capable tools (`visudo -cf` goes to the *real* `/usr/sbin/visudo`; only the whole-set `-c`, which a non-root test cannot read, is scripted). Covers: the exact QA-box file detected; spacing/`%clawbox`/`(ALL:ALL)`/line-continuation variants detected; the shipped drop-ins, commented-out rules, `PASSWD:`-tagged `ALL`, and `%sudo`/`%admin`/other-user rules **not** detected; install at 0440 root-owned; a bad candidate leaves the installed file intact; missing source is a failure not a no-op; staging leaves nothing behind and never lands in `/etc/sudoers.d`; rollback of a previous file and removal of a first-time install when `visudo -c` breaks; quarantine keeps a 0400 copy, skips managed drop-ins, leaves the operator's rule alone, is idempotent, and restores everything when the removal breaks `visudo -c`.

`src/tests/unit/sudoers-coverage.test.ts` (15) drives the checker against a fixture root that shares the real `src/`/`mcp/` trees, asserting it passes as shipped and fails on: an uncovered call, a removed grant, an unused grant, an unresolvable argv, a blanket `ALL`, a non-root runas, an unparseable line — and that it does *not* mistake `echo "Run: sudo …"` for an invocation. One test traces the brief's operability list end to end and asserts each is still a resolved call site: `clawbox-root-update@{set_hostname,restart_ap,chpasswd}` (wizard, password, hotspot), `reboot`/`poweroff` (power menu), the gateway mask/stop/unmask trio (factory reset), both desktop-mode and both power-mode flags (Settings toggles), `ollama.service` enable/disable and `optimize-ollama.sh` (local models), tunnel `restart`/`enable` (Remote Control).

`e2e-install/06-sudoers.spec.ts` (7, read-only) makes the same assertions against the **real container filesystem** after `install.sh` has run — which is the layer that failed on the QA box. The container's `clawbox ALL=(ALL) NOPASSWD:ALL` drop-in (written by `e2e-install/Dockerfile` so volume permissions work pre-install) is now a genuine fixture: `install.sh` quarantines it exactly as on a device, so a regression in `quarantine_overbroad_sudoers` fails the first test. It also asserts every granted libexec helper is `755 root:root` under root-owned dirs, no grant points into `/home/clawbox/clawbox`, the root-update unit ExecStarts the root-owned entrypoint, and `visudo -c` still passes.

`85-tunnel.spec.ts`'s setup/teardown moved from `dockerExec(sudo …, {user: "clawbox"})` to plain root — planting a stub binary in `/usr/local/bin` is harness work, not something the product does, and after this change `sudo tee` from clawbox correctly hits a prompt.

Two existing assertions moved with the change, both keeping their intent:
- `root-steps.test.ts` pinned the literal line `"  install_root_libexec\n  # Install sudoers rules"`; it now asserts the **order** of `install_root_libexec` vs `install_sudoers_dropin` inside `step_systemd_services`, which is the actual invariant.
- `install-foreign-edition-teardown.test.ts` required a `start clawbox-gateway` grant to justify the mask; it now accepts `restart` too, which brings a stopped unit up just the same — the escalation the mask blocks is unchanged.

## What is NOT covered

- **Not verified on hardware.** Box `.71` is owned by the hardware-test lane this round, so nothing here was exercised on a real Jetson. The migration is proven by the unit tests (real functions, real `visudo -cf`, temp `/etc/sudoers.d`) and by `06-sudoers.spec.ts` against a container `install.sh` actually provisioned. Code, tests, and the findings evidence are the ground truth as the round-2 addendum instructs.
- **The other two `NOT-APPLIED` sub-claims in §3 are not re-fixed here** — `c445c` and `c445a` were already correct in the repo; the revalidation's verdict was that `/etc` on that one box was stale (Aug 21 copies, no `/usr/local/libexec` at all) because no installer or updater run had landed since. Any `sudo bash install.sh` or in-app update applies them, and this PR's migration rides the same path.
- **`sudo -n hermes gateway install --system` stays ungranted** (Telegram gateway install on the Hermes SKU). `HERMES_BIN` is under `/home/clawbox/.local/bin`, which clawbox owns and can rewrite; granting it would be the same one-step local root TASK-445 exists to close. Documented as an exemption with that reason; the caller already treats the failure as non-fatal. Making that path work needs a root-owned wrapper — out of scope here, flagging it.
- **`update/run` remains anonymously reachable during a genuine first-boot window** (`setup-api-gate.ts:53`) and `updater.ts` still hard-syncs during `git_pull`. Both are documented, deliberate carve-outs from TASK-443/#436, unchanged by this PR.
- The quarantine detector splits Cmnd lists on top-level commas and does not model a `Runas_Spec` containing a comma (`(root, clawbox)`). That fails *closed* — such a rule is not quarantined, i.e. status quo — and no such rule has been observed.

## Summary by CodeRabbit

- **Security Improvements**
  - Tightened privileged service controls and removed unnecessary passwordless permissions.
  - Added safeguards for secure privilege-rule installation, rollback, and integrity verification.
  - Improved protection against unsafe configuration and credential files.

- **Reliability Improvements**
  - Gateway setup now reports success only when changes are running and applied.
  - Added clearer warnings when restarts are refused or fail.
  - Improved service recovery by clearing stale failure states before restarting.

- **Quality Improvements**
  - Added automated validation and expanded coverage for installation, privilege rules, updates, and tunnel setup.


KrasimirKralev (commented):
## Adversarial review — TASK-445 round 2

Reviewed at `f7662e1`, test-merged onto current `beta` (`7de5b39`, i.e. including #485/#489/#490 coding-agent and #488 factory-reset-reauth) in a throwaway worktree, plus a **read-only** pre-state capture from the Hermes QA box `.165`. Nothing installed, nothing changed, nothing merged.

**Verdict: needs changes before merge — two blockers.** Neither is in the allow-list itself. The allow-list is right; the machinery around it has one silent-failure path, and the quarantine has one un-modelled consequence on the Hermes SKU.

---

### What I verified as solid

The core audit came back clean, and this is the part I expected to find holes in:

* **Completeness of the allow-list.** I enumerated every `sudo` invocation in `src/`, `mcp/`, `scripts/`, `config/`, `install.sh` and the e2e harness independently of `check-sudoers-coverage.sh`, then test-merged the PR onto current `beta` and re-ran the checker there: `OK — 41 grants, 42 resolved sudo invocations, 0 gaps`. No missing grant for any product flow I could reach — wizard, updater, power menu, factory reset, hotspot, desktop toggle, Local Models, Browser app, Remote Control, ClawKeep restore.
* **The 10 pruned grants are genuinely dead.** `start clawbox-gateway|clawbox-setup|clawbox-tunnel` — every caller uses `restart`, including the two user-facing instruction strings (`setup-api/gateway/route.ts:95`, `ClawKeepApp.tsx:1921`), which stay covered by `restart` + `reset-failed clawbox-*`. `clawbox-ap` restart/stop — no TS call site; `config/clawbox-ap.service` runs `stop-ap.sh` as its own `ExecStop` (root), `scripts/ap-watchdog.sh` runs from a root timer unit, and Settings → Hotspot goes through `clawbox-root-update@restart_ap.service` (`system/hotspot/route.ts:142`). Confirmed, not just asserted.
* **nmcli/WiFi genuinely needs no grant** — every `nmcli` call site (`lib/network.ts`, `wifi/*`, `system/hotspot`, `setup/reset:115,129`) runs unprivileged. See the correction under Minors on *which* polkit file does that, though.
* **The updater has no chicken-and-egg.** Every privileged updater step runs through `clawbox-root-update@<step>.service` (root, `ExecStart` = root-owned `clawbox-root-step.sh`); the only sudo the updater itself takes is `reset-failed clawbox-*` / `start [--no-block] clawbox-root-update@*` — both preserved, and both wildcards match the templated unit names. `step_systemd_services` (where the migration happens) is itself inside a root step, so there is no window where the updater loses the root it needs to finish.
* **The detector matches the real artefact.** Box `.165` carries `/etc/sudoers.d/clawbox-nopasswd` = exactly `clawbox ALL=(ALL) NOPASSWD: ALL` (32 bytes), and `sudo -l` there shows `(ALL) NOPASSWD: ALL` sorting *after* the narrow `clawbox` file — the union problem, measured. I extracted `sudoers_grants_blanket_nopasswd()` out of `install.sh` and ran it against that exact byte string: detected. Negative controls (`config/clawbox-sudoers`, `config/sudoers-clawbox-ollama`): not detected. The `CLAWBOX_SUDOERS_MANAGED` basename comparison is exact, so `clawbox-nopasswd` is correctly *not* treated as the managed `clawbox` file.
* **`optimize-ollama.sh`**: install location (`install_root_libexec`, `root:root 0755`), grant target and both call sites now agree, and `.165` already shows the file present at `/usr/local/libexec/clawbox/optimize-ollama.sh 755 root:root`. Moving both drop-ins to `step_systemd_services` is the right call — the `is_test_mode` early return really did make the ollama grant conditional.
* **Ordering and fail-safety** of install-then-quarantine, the staging directory outside `/etc/sudoers.d`, the whole-set `visudo -c` re-check, the restore-everything path when a removal breaks the set, and the deliberate refusal to touch `%sudo`/`%admin` — all correct, and the 85-tunnel harness change (root instead of clawbox-with-sudo) is the right fix rather than a test weakened to pass.
* **The `set -e` shape I went looking for isn't there**: `for m in "${CLAWBOX_SUDOERS_MANAGED[@]}"; do [ "$base" = "$m" ] && managed=1 && break; done` returning 1 does not trip `errexit`. Verified on bash 5.2.
* **The checker does run in CI**, contrary to CodeRabbit's `package.json:27` finding — `src/tests/unit/sudoers-coverage.test.ts:69` runs it against the real `REPO` root, so the `test` job enforces it. (It self-skips without `perl`; CI has it.)

---

### Blocker 1 — `install_sudoers_dropin` runs with `errexit` suppressed, so its unchecked writes fail silently

CodeRabbit raised this at `install.sh:2734` and its bot marked it **"Addressed in commits 0d768b7 to f7662e1"**. It is not addressed — the code at `f7662e1` is unchanged, and the reasoning that might have dismissed it (`set -euo pipefail` will abort anyway) does not hold here.

Both call sites invoke the function in a condition context:

```sh
if install_sudoers_dropin "$PROJECT_DIR/config/clawbox-sudoers" clawbox; then   # install.sh:2941
install_sudoers_dropin ".../sudoers-clawbox-ollama" clawbox-ollama || echo ...   # install.sh:2957
```

bash disables `errexit` for the **entire body** of a function called from an `if`/`||` condition. Verified:

```
$ bash -c 'set -euo pipefail; f(){ /bin/false; echo "  CONTINUED past the failure"; return 0; }; if f; then echo "RESULT: SUCCESS"; fi'
  CONTINUED past the failure
RESULT: SUCCESS
```

So every unchecked mutating command inside the function is silent: `cat "$src" > "$staged"`, `chown`, `chmod`, `install -m 0440 "$staged" "$dest"`, `cat "$dest" > "$backup"`, and the rollback `install ... "$backup" "$dest"`.

The failure chain, on a device whose eMMC filled up mid-update (the realistic trigger on an 8 GB board):

1. `install ... "$dest"` fails → `$dest` is unchanged, or on a fresh install never created.
2. `visudo -c` passes anyway — the previously installed set is still on disk.
3. The function returns **0**.
4. `quarantine_overbroad_sudoers` runs and removes `/etc/sudoers.d/clawbox-nopasswd`.
5. The device now has **neither** the blanket grant nor the allow-list. Password prompt on `systemctl restart clawbox-gateway`, on `clawbox-root-update@chpasswd`, on `reboot` — on an appliance with no console. That is precisely the outcome the new ordering exists to prevent.

A second variant: `cat "$src" > "$staged"` truncating under ENOSPC can still leave a file `visudo -cf` parses (a valid prefix of the allow-list), so a *partial* allow-list installs and validates. And the rollback path prints `"rolled $dest back"` whether or not the rollback actually succeeded.

Suggested fix, beyond CodeRabbit's `install`-exit-status diff: gate the quarantine on positive proof rather than on a return code. After the install, before returning 0, assert the installed file is what was intended —

```sh
cmp -s "$staged" "$dest" || { echo "Error: $dest does not match the candidate; keeping existing grants" >&2; rm -f "$staged" "$backup"; return 1; }
```

— and check `cat`/`chown`/`chmod`/the rollback `install` explicitly too, since `errexit` will not do it for you here.

---

### Blocker 2 — on the Hermes SKU the quarantine removes root that a shipped feature is using today, and the UI reports success anyway

`EXEMPT_CALLS` documents `sudo -n hermes gateway install --system` as ungranted, non-fatal, "the only caller … treats the failure as non-fatal". Two things are off.

**(a) There are two sudo paths, not one.** `ensureHermesGateway()` (`hermes-telegram.ts:432`) takes `sudo: systemScope` on the **restart** branch (`:442`) as well as the install branch (`:463`). The restart branch is the one that runs on every configuration save on an already-provisioned box — which is every Hermes box in the field.

**(b) The failure is invisible.** `runHermesCli` **resolves** on a non-zero exit (`hermes-cli.ts:117` — `child.on("close", … resolve({code, …}))`); it does not throw. `hermesGatewayStatus()` then runs `hermes gateway status` *without* sudo, sees the old gateway process still up, and returns `running: true`. So `telegram/configure/route.ts:69-87` falls straight through to:

```ts
return NextResponse.json({ success: true, reset: tokenChanged, restarted: true });
```

Same shape in `whatsapp/configure:135`, `discord/configure:129`, `hermes-email.ts:113,150`, `whatsapp-pairing.ts:682`. The owner changes a bot token / mailbox password, the UI says it restarted, and the gateway keeps serving the old config until something else restarts it.

**This is a live regression, not a theoretical one.** From `.165` (read-only, `sudo -n -l`, nothing executed):

```
CLAWBOX_EDITION=hermes
hermes-gateway.service   loaded active running   (User=clawbox, ExecStart=/home/clawbox/.hermes/.../venv/bin/python -m hermes_cli.main gateway run)
/home/clawbox/.local/bin/hermes  -rwxr-xr-x clawbox clawbox

$ sudo -n -l /home/clawbox/.local/bin/hermes gateway restart --system
/home/clawbox/.local/bin/hermes gateway restart --system        # permitted today, via clawbox-nopasswd
```

After the quarantine that becomes `sudo: a password is required`.

**A clean fix for the restart half, in-pattern and with zero added privilege:** `hermes-gateway.service` is a root-owned unit in `/etc/systemd/system` that runs `User=clawbox`. Restarting it grants clawbox nothing it does not already have — it is strictly less privilege than the current `restart clawbox-gateway` grant. So:

```
clawbox ALL=(root) NOPASSWD: /usr/bin/systemctl restart hermes-gateway.service
clawbox ALL=(root) NOPASSWD: /usr/bin/systemctl restart hermes-gateway
```

and have `ensureHermesGateway` take the `systemctl` path when `scope === "system"` instead of `hermes gateway restart --system`. That removes the clawbox-writable `HERMES_BIN` from the root path entirely, which is better than what the box does today.

The first-time **install** half is the one that genuinely needs root to write a unit file — the existing `clawbox-root-update@<step>.service` template is already granted and already root-owned (`clawbox-root-step.sh`), so a `hermes_gateway_install` step id would cover it without a new grant or a new wrapper.

If neither lands in this PR, then at minimum the routes must stop reporting `restarted: true` when the restart was denied — a `warning` is the difference between "a feature is degraded" and "a feature is degraded and the box says it isn't".

---

### Minors

1. **`install_sudoers_dropin` silently tightens a shared directory.** `install -d -o root -g root -m 0700 "$SUDOERS_STAGING_DIR"` chmods **`/var/lib/clawbox`** itself — currently `0755 root:root` on `.165` — from 0755 to 0700, as a side effect of sudoers staging. Nothing breaks today (`clawbox-power-mode.sh`'s `l4t_dfs.conf` and `ensure-vnc-on-first-boot.pending` are root-only), but `docs/clawkeep-restic-migration.md:151` already plans `/var/lib/clawbox/clawkeep.cred` *owned by clawbox* in that directory, and a 0700 root parent kills traversal for it. Stage in a dedicated `/var/lib/clawbox/.sudoers-staging` (0700) instead of hardening the shared parent. Also: a kill between `mktemp` and `rm -f` leaves `.sudoers-candidate.XXXXXX` in a shared dir forever.
2. **PR body / comment cite the wrong polkit file.** "WiFi goes through polkit (`config/49-clawbox-updates.rules`)" — `install.sh:3418` explicitly `rm -f /etc/polkit-1/rules.d/49-clawbox-updates.rules` and installs `config/49-clawbox-updates.pkla` instead; that `.pkla` is what carries the `org.freedesktop.NetworkManager.*` actions. `49-clawbox-updates.rules` is a dead file in the repo and only covers `clawbox-root-update@` anyway. The conclusion is right, the cited mechanism isn't.
3. **Most sudo call sites still omit `-n`.** `ai-models/configure:1826`, all of `browser/manage`, `browser/route:200`, `clawkeep/restore:55`, `llamacpp/install:123,129`, `setup/reset:190-389,553`, `system/credentials:135,138`, `system/hostname:80`, `system/hotspot:142`, `system/power:49`, `local-models:579`, `openclaw-config:1116`, `system-profile:98`. With the list narrowed, an argv mismatch is now the expected failure mode, and without `-n` it manifests as a route handler hanging for the full timeout (up to 120 s in `browser/manage`) instead of failing in milliseconds with `sudo: a password is required` in the journal. Worth a follow-up, and worth teaching `check-sudoers-coverage.sh` to require `-n` on every non-exempt call site — that turns "we got the argv right" into something the build proves.
4. **`06-sudoers.spec.ts` asserts substrings of `sudo -l`, not that sudo would actually match the argv the code passes.** The negative direction is genuinely behavioural (the three `sudo -n` probes are good, and using the container's seeded blanket drop-in as a migration fixture is a nice touch), but the positive direction is `expect(listed).toContain(needle)`, which cannot catch a fixed-args mismatch — the exact failure mode the allow-list introduces. `sudo -n -l -- <exact argv>` is non-mutating and exits non-zero when the argv is not permitted; driving it from `check-sudoers-coverage.sh --json`'s resolved call-site list would make the container prove every one of the 42 sites end to end. That would have caught the `enable --now` class of bug without anyone reasoning about it.
5. **Two CodeRabbit findings on the checker are still open and both are fail-open**, which matters more than usual for a script whose contract is "fail closed": `walk()` returning silently when a scan root is missing (`:200`), and the `$src =~ /sudo/` gate being case-sensitive (`:405`) — a file that only ever names an imported `SUDO_BIN` is skipped whole. `/sudo/i` and an explicit `-d` check are one-liners.
6. `install.sh:3374` (CodeRabbit, open): `"$ROOT_LIBEXEC_DIR/optimize-ollama.sh"` is unguarded in both `step_performance_mode` and `step_ollama_install`, while the `clawbox-power-mode.sh --apply` call a few lines above it *is* guarded with `|| echo warning`. Not a regression (the old `bash "$PROJECT_DIR/scripts/optimize-ollama.sh"` had the same exposure), but the inconsistency is right there in the same function and it aborts `step_resource_limits` behind it.
7. Detection gap worth one line of documentation: the quarantine only walks `/etc/sudoers.d`. A `clawbox … NOPASSWD: ALL` line appended to `/etc/sudoers` itself by some future provisioning script would keep the narrowing decorative and nothing would say so. Not touching `/etc/sudoers` is the right call; *reporting* when it contains a `clawbox`/`%clawbox` blanket rule would close the loop without the lockout risk.

---

### Summary

The narrowing itself is the strong part of this PR and I could not break it: I looked specifically for a grant whose fixed args disagree with its caller, for a path only reachable on the Hermes SKU, for a call site added to `beta` after this branch was cut, and for a step that skips the migration — and found none. The reverse-direction check finding 10 dead grants, and the checker catching #472's `local-ai-runtime.ts` in both directions during the rebase, are exactly the evidence that the guard works.

What needs to change is on either side of the list: a function that can report success after failing to write the file the whole migration depends on, and a quarantine that withdraws root the Hermes messaging stack is quietly using today while the UI keeps saying "restarted". Blocker 2 in particular is the kind of thing that only shows up on a real Hermes box, and `.165` says it will.

— Krasimir's review pass. Read-only throughout; `.165` was inspected with `ls` / `sudo -n -l` / `systemctl cat` only, nothing installed or modified.

KrasimirKralev:
## Agent-capability audit: does the narrowing break the AI agent?

The earlier review verified the allow-list against the **app's** 42 sudo invocations. This is the
missing half — what the **Hermes/OpenClaw agent** needs, since it runs as `clawbox` with a real
terminal tool and today inherits whatever `clawbox` can do. Repo audit + read-only forensics on
the QA box (166 sessions / 447 messages of real agent history) and on a freshly provisioned box.

**Short answer: the allow-list needs no additions for agent capability. Zero (b)-class gaps.**

### Evidence

**1. Empirical.** Full agent transcript store on the QA box: 166 sessions, 447 messages, 21
`terminal` tool calls. `sudo` appears in **3** messages, of which exactly **one** is an
invocation — and it is an `||` fallback for a cosmetic convenience:

```
cp ~/generated-images/x.jpg /var/www/html/ 2>/dev/null || sudo cp ~/generated-images/x.jpg /var/www/html/ ...
```

The next two turns are the interesting part: the agent checked, found no web root, said *"No web
root — I'll spin up a tiny local HTTP server"*, and served the file with an unprivileged
`python3 -m http.server`. **It already degrades gracefully the moment it gets a definite "no".**
The other two hits are the mascot line `"sudo make me a sandwich"` and the agent *describing*
`sudo bash install.sh`. `~/.hermes/logs/agent.log` (3 MB, 4 days): 0 hits.

**2. Contractual.** No MCP tool invokes `sudo`. The two that reach a privileged app route —
`system_power` (`systemctl reboot|poweroff`) and `browser_open` (`systemctl start
clawbox-browser.service`) — are both allow-listed. The tool descriptions already disclaim the
rest: `update_check` *"never installs anything… tell the user to install it from Settings"*,
`wifi_scan` *"this does not connect to anything"*. `preferences_set` is charset-restricted to 5
keys, so the desktop/power-mode grants aren't even agent-reachable. `coding_agent_run` is
already `setpriv --no-new-privs` + `Bash(sudo:*)` denied, and nothing broke.

**3. Skills.** ~60 installed skills on the QA box. Every `sudo` in them is an optional
`apt`/`snap` install in a *fallback* branch of a reference doc (`cowsay`, `boxes`, `toilet`,
`jp2a`, `texlive-full`, `latexdiff`, `nvidia-cuda-toolkit`, `ufw`). Two mention
`sudo loginctl enable-linger`, which ClawBox does not need — installing the gateway as a
**system** unit is precisely how `ensureHermesGateway` avoids linger. No bundled ClawBox skill
exists; skill installs go through the Hermes CLI **without** sudo.

### Classification

| Class | Count | Notes |
|---|---|---|
| (a) legitimate, already covered | all of it | reboot/poweroff, browser+chromium, ollama, tunnel, `clawbox-root-update@*`, skills (no sudo needed) |
| (b) legitimate, **missing** | **1**, and it's an *app* path, not an agent path | `systemctl restart hermes-gateway[.service]` — blocker B2 |
| (c) unbounded by nature | 1 attempt / 166 sessions | ad-hoc `sudo <anything>`. **No grant proposed.** See below. |
| (d) must-not-have | the blanket drop-in | plus two escalations this PR doesn't close, below |

**On (c)** — please don't add a package-install grant. Any `apt-get install` form broad enough to
be useful is a root escalation in one move (`-o APT::Update::Pre-Invoke=`), which is exactly the
SEC-2 reasoning behind the exact-match chromium grants. Measured demand is ~0. The right fix is
UX, not privilege: add `sudo`/`su` to `DANGEROUS_PATTERNS` in `mcp/lib/jobs.ts` so `bash` returns
a clean refusal the agent can relay ("that's a Settings action") instead of the agent watching a
command die on a password prompt. Separate, non-blocking follow-up.

### Failure-mode UX (measured, not guessed)

Ran `sudo -u nobody -- sudo …` on the QA box (an identity with no matching rule):

| Context | Result | Time |
|---|---|---|
| no TTY (`execFile`/`spawn`), without `-n` | `sudo: a terminal is required to read the password…` | **30 ms** |
| no TTY, with `-n` | `sudo: a password is required` | **28 ms** |
| **with a PTY** (`script -qec`), without `-n` | **hangs** on `[sudo] password for …` | unbounded |

So the "route hangs for 120 s" concern is real but only in PTY contexts, and the app's call sites
all spawn without a PTY. The ~20 sites still missing `-n` should get it anyway — the guarantee
should come from the flag, not from how the child happens to be spawned, and `Defaults use_pty`
(present in `/etc/sudoers` on the box) is one config change from turning that into a hang.

### Two things this PR does NOT close — worth naming so they aren't mistaken for solved

**1. `config/49-clawbox-updates.pkla` makes the systemd half of the allow-list advisory.**

```ini
Identity=unix-user:clawbox
Action=org.freedesktop.systemd1.manage-units
ResultAny=yes
```

No unit filter. The `clawbox` user can start/stop/restart/enable **any** unit on the box with no
sudo and no password — including `clawbox-root-update@<any of the 50 dispatcher steps>.service`.
Verified live: `pkcheck --action-id org.freedesktop.systemd1.manage-units` returns authorized,
and an unprivileged `systemctl restart <unit>` succeeds. The second block grants unfiltered
`NetworkManager.settings.modify.system` + `network-control` the same way.

The scoped alternative is already in the repo — `config/49-clawbox-updates.rules:6` filters on
`unit.indexOf("clawbox-root-update@") === 0` — but `step_polkit_rules` installs the `.pkla` and
`rm -f`s the `.rules`. That's not an oversight: the Jetson base ships polkit **0.105** (confirmed
on the box), which only reads `.pkla`, and `.pkla` can't express a per-unit filter. So this needs
a design decision, not a one-liner, and `/setup-api/install/run-step/route.ts:83-89` deliberately
calls `systemctl` without sudo and works *only* because of this grant. Filing it as a separate
issue rather than expanding this PR — but I'd argue #471 shouldn't be described as closing the
clawbox→root path while this is open.

**2. `install.sh` is clawbox-writable and is what `clawbox-root-step.sh` execs as root.**
`clawbox-root-step.sh:14-15` already says so. Edit → `systemctl start clawbox-root-update@build`
→ root. Acknowledged in-tree; just noting it stays open.

### Verdict

- Agent capability after the narrowing: **unchanged in every user-visible respect.**
- The one grant to add is B2's `systemctl restart hermes-gateway[.service]`, which is
  **strictly less** privilege than the existing `restart clawbox-gateway` grant: verified against
  the Hermes CLI source that `gateway install --system` writes a root-owned unit with
  `User=clawbox` and `StartLimitIntervalSec=0` (so no `reset-failed` grant is needed either), and
  it lets us stop `sudo`-ing the **clawbox-writable** `/home/clawbox/.local/bin/hermes` — the
  thing `EXEMPT_CALLS` currently has to excuse.

Full report with file/line citations and the raw query output is in the follow-up PR.

KrasimirKralev:
## Follow-up PR for the two blockers: #495

Opened against **this branch** (`fix/hermes-sudoers-445-r2`), not beta, so it reviews and lands as part of #471 rather than rewriting it. → **#495**

**B1 — a failed sudoers install could still trigger the quarantine.** `install_sudoers_dropin` (install.sh:2680) returns 0 after an `install(1)` that never wrote: both call sites run it in a condition context, which suspends `set -e` for the whole function body, and the closing `visudo -c` (install.sh:2724) then validates the *old* file on disk. The caller reads that 0 as permission to remove `/etc/sudoers.d/90-clawbox-nopasswd`, so a device whose only grant is the blanket one ends with **neither** file. #495 checks every step inside the function, compares the destination **byte-for-byte** with the staged candidate (a truncated allow-list still parses, so only a comparison catches it), calls the function plainly with an explicit status, and gates the quarantine on `cmp -s` between the shipped file and what is actually in `/etc/sudoers.d/clawbox` — proof about the device, not about a code path. Regression tests lift the real gate out of `step_systemd_services` and assert the blanket drop-in survives a refused install, a truncated install and a failing `visudo -c`.

**B2 — the `hermes gateway restart` exemption.** Detail in #495; the short version is that `ensureHermesGateway` takes `sudo: systemScope` on the **restart** branch too (`hermes-telegram.ts:442`), which is the path every messaging config save hits on a provisioned box — and `runHermesCli` resolves on non-zero (`hermes-cli.ts:117`) while the follow-up status probe runs unprivileged, so a refused restart came back as `{restarted: true}`. Replaced with `systemctl restart hermes-gateway[.service]`, and `ensureHermesGateway` now returns an `applied` flag so nothing claims a restart it did not achieve.

**Validation, stated honestly:** the allow-list parses under a real `visudo`, `check-sudoers-coverage.sh` is `OK — 43 grants, 43 resolved sudo invocations, 0 gaps`, and the B1 gate passes 8/8 against a real `visudo` on Linux. **No live `/etc/sudoers.d` swap was performed** — both boxes moved off my subnet mid-task and the owner's box was mid-testing throughout. The outstanding check is one install/update run on a box that still carries the blanket drop-in, which is what `e2e-install/06-sudoers.spec.ts` covers.

One note for when you check the new grant on a device: `hermes-gateway.service` does not exist until the first messaging platform is configured — that unit is what `ensureHermesGateway` installs. On a box that never had Telegram/WhatsApp/Discord/Email set up, `systemctl show hermes-gateway.service` reports "no such file". It is not a dead grant, just an unexercised one.

I have not merged anything and will not — both #471 and #495 stay your call.

KrasimirKralev:
## Separate finding: the polkit rule makes the systemd half of this allow-list advisory

Raising this on its own so it does not get lost inside the follow-up PR. **It is not something #495 fixes, and I am deliberately not attempting a fix** — it needs a design decision, and the obvious one-line change would break the updater.

`config/49-clawbox-updates.pkla`, installed by `step_polkit_rules`:

```ini
[Allow clawbox to manage ClawBox systemd services]
Identity=unix-user:clawbox
Action=org.freedesktop.systemd1.manage-units
ResultAny=yes
ResultInactive=yes
ResultActive=yes
```

**No unit filter.** The `clawbox` user — which is the web server, the in-UI terminal and the agent's shell — can start, stop, restart and enable **any** systemd unit on the box with no sudo and no password. That includes `clawbox-root-update@<any dispatcher step>.service`, which is the same escalation the `systemctl start clawbox-root-update@*.service` grant in this PR is scoped to allow.

Verified live on the QA box (read-only):

```
$ pkcheck --action-id org.freedesktop.systemd1.manage-units --process $$
$ echo $?
0                                    # authorized

$ systemctl restart <a real unit>    # no sudo
$ echo $?
0
```

The second block of the same file grants unfiltered `NetworkManager.settings.modify.system` + `network-control` the same way, so Wi-Fi reconfiguration is equally sudo-free.

### Why this is not a one-liner

The scoped version already exists in the repo — `config/49-clawbox-updates.rules:6` filters on `unit.indexOf("clawbox-root-update@") === 0`. But `step_polkit_rules` installs the **`.pkla`** and `rm -f`s the `.rules`, and that is correct for the hardware: the JetPack base ships **polkit 0.105** (confirmed on the device: `pkaction version 0.105`, `policykit-1 0.105-33`), which reads `.pkla` only and has no JS rules engine at all. `.pkla` syntax cannot express a per-unit filter — it keys on action id, not on the object being acted upon. So the options are roughly:

1. Backport/ship a newer polkit on the appliance image, then use the existing `.rules` file.
2. Drop the `manage-units` grant from the `.pkla` entirely and route every call that relies on it through the sudoers grants instead.
3. Accept it and stop describing the clawbox→root path as closed.

Option 2 is a real behaviour change and needs care: `src/app/setup-api/install/run-step/route.ts:83-89` deliberately calls `systemctl` **without** sudo and works only because of this grant, and `src/lib/local-ai-runtime.ts:241` tries `sudo -n` first and falls back to an unprivileged `systemctl` — the fallback is load-bearing on any box whose installed drop-in predates the `start`/`stop ollama.service` grants this PR adds. Narrow the polkit rule without checking those and Local AI standby breaks silently, in the same shape as the earlier "Settings says Local AI is on, `ollama.service` is dead underneath" bug.

### What I would suggest

Not blocking #471 on it — the narrowing is still a real improvement and the blanket drop-in still has to go. But I would avoid describing #471 as closing the clawbox→root path while this is open, and I would file it as its own task under the TASK-442 epic so the decision is made rather than inherited.

For completeness, the third path in the same class, already acknowledged in-tree at `config/clawbox-root-step.sh:14-15`: `install.sh` lives in the clawbox-writable project tree and is what `clawbox-root-step.sh` execs as root. Edit it, start `clawbox-root-update@build.service`, get root. Unfixable without a root-owned copy of the installer, but worth being in the same ticket so the three are traded off together rather than one at a time.

KrasimirKralev:
## Upstream check: do the harnesses actually require passwordless sudo?

Before we close the remaining gaps it was worth establishing whether `clawbox ALL=(ALL) NOPASSWD: ALL` is something **OpenClaw or Hermes expect**, or something **we added ourselves**. Answer: it's ours. Both upstreams document the opposite, and Hermes prescribes an allow-list by name.

Method: upstream docs + upstream source, plus read-only forensics on a live Hermes box (`hermes-agent` v0.20.5). Full write-up with all citations kept locally (`clawbox-upstream-sudo-expectations.md`); condensed here for the record.

### Hermes (NousResearch)

- Its own installer prints, three times (`scripts/install.sh:1169,1185,1569`):
  > `sudo is needed ONLY to install optional system packages (...) via your package manager.`
  > `Hermes Agent itself does not require or retain root access.`
- Install docs: *"Running Hermes as a dedicated unprivileged user (e.g. a `hermes` systemd service account, or **any user without `sudo` access**) is supported."*
- It **does** manage systemd itself — but expects a human at a sudo prompt, never a service account with NOPASSWD. `hermes_cli/gateway.py:2937-2941`:
  > `Scope choice: always the least-surprising, no-privilege option — user-scope systemd unit on Linux ... Users who want a boot-time system service still run `hermes gateway install --system` explicitly (that path prompts and requires root; we never self-elevate from an installer).`
  and `gateway.py:2854-2858`: *"a non-root user is never handed a 're-run yourself under sudo' recipe"*.
- It actively stops **the agent** using sudo: `sudo`/`doas`/`su` are in `_UNSAFE_ROOT_BINARIES` and `_UNSAFE_CLASS_PATTERNS` (`approvals_suggest.py:31,67,99`, *"Dangerous root binaries never become globs"*), plus an unconditional *"sudo stdin guard"* alongside the hardline blocklist (`approvals_test.py:29,98-103`).
- **The FAQ prescribes exactly what this PR does.** Section *"`sudo` not working via messaging gateway"*:
  > **Solution:** Avoid `sudo` in messaging — ask the agent to find alternatives · If you must use `sudo`, **configure passwordless sudo for specific commands in `/etc/sudoers`** · Or switch to the terminal interface for administrative tasks: `hermes chat`

### OpenClaw

- The managed daemon is a **systemd user service** (`docs.openclaw.ai/install`: *"Linux/WSL2: systemd user service via the same commands"*). No root on the happy path. Its own code only *prints* sudo hints (`src/daemon/systemd-system.ts:156`); the one place it spawns sudo is opt-in behind an explicit `sudoMode` (`src/daemon/systemd-linger.ts:54-71`).
- Security guide: *"Prefer a dedicated OS user account for the Gateway if the host is shared."* · *"Use sandboxing and host isolation for strong boundaries."* · *"For mixed-trust or adversarial-user operation, split trust boundaries: separate gateway + credentials, ideally separate OS users or hosts."*
- Upstream's twin of our `agent-interface.mdx:109` line (secure-file-operations): *"...**not a sandbox**. Host filesystem permissions, OS users, containers, and the agent/tool policy still define the real blast radius."*
- Its exec auto-reviewer is told to treat *"chmod/chown, rm/mv sensitive paths, **sudo**, ssh/scp/rsync, and secret paths as high security risk"* (`src/agents/exec-auto-reviewer.prompt.ts:12`).

### Per-operation verdict

Mapped every grant in `config/clawbox-sudoers` on this branch to who needs it. **0 of 16 categories are required by a harness.** 13 are ClawBox product features (gateway/setup/browser/tunnel restarts, ollama, `clawbox-root-update@*`, reboot/poweroff, chromium install, desktop/power mode, hotspot), 2 are our own defects, 1 is a provisioning artifact. Neither harness asks for a single one of them — `clawbox-gateway` is a system unit because *we* chose that; upstream ships a user unit.

The only harness-adjacent need is **`systemctl restart hermes-gateway.service`** (B2 / #495), and it is a ClawBox product path (Settings → Telegram save), not an upstream requirement.

### Three live findings from the owner's box today (read-only)

1. **`hermes-gateway.service` is installed, `enabled` AND `active`** — root-owned, `User=clawbox`, `StartLimitIntervalSec=0`, `KillSignal=SIGTERM`/`TimeoutStopSec=60`. So B2 is not hypothetical, and the restart grant is confirmed zero-escalation. This corrects the earlier note that we never install it.
2. **The blanket rule is still live**: `sudo -n -l` as `clawbox` returns the full allow-list **and** `(ALL) NOPASSWD: ALL`. Until `quarantine_overbroad_sudoers` actually runs on provisioned devices, the allow-list is decorative.
3. **The polkit hole is on this box too**: `pkcheck --action-id org.freedesktop.systemd1.manage-units` → **rc=0** for `clawbox`, same as on `.71`. The systemd half of the allow-list stays advisory until `49-clawbox-updates.pkla` is scoped (TASK-539).

### Recommendation

**Narrowing is upstream-safe — there is no upstream contract to preserve, and nothing must be kept for harness compatibility.** This PR converges on upstream's documented posture rather than diverging from it.

Two things that must **not** be granted:

- `sudo` on `/home/clawbox/.local/bin/hermes` — clawbox-writable (`-rwxr-xr-x clawbox:clawbox`, re-verified today), so it's one-step local root. Upstream anticipates the underlying PATH problem for service accounts and prescribes *symlinking the launcher into a system location*, not sudo. Take the `systemctl restart` grant instead.
- Any broad `apt-get install` form — already documented here as SEC-2.

Comparable practice, briefly: Home Assistant OS runs every add-on in *"protection-enabled mode, which prevents the app from getting any rights on the system"* by default, with `full_access` as a flagged opt-in. Narrowing the agent user's sudo is the normal posture for this class of appliance; blanket NOPASSWD is the outlier.

This covers the upstream-compatibility half only — it does not adjudicate the 5 gaps in Mike's independent audit, and nothing above argues for widening the allow-list to close any of them.

PR #474, opened 2026-08-24
Hermes MCP/tooling round 2: skill uninstall collision, webapp leak, clawhub ids, tunnel status (TASK-453)

yalexx:
## Task

**TASK-453 round 2** — MCP/tooling defects that survived PR #435, plus the tunnel-status false negative found in the smoke pass.

Evidence: `hermes-night/findings/revalidate-round1.md` §6 (from line 817, "New findings" line 952), §7 (line 1003, "New findings" line 1240), §12 (line 2192, new finding line 2396). Every defect below was observed live on the Hermes QA box on 2026-08-24.

## Ruling

> Deliverables: fix each with a regression test (mock the builtin/store collision for (1); temp dir for (2); a fake search fixture for (3)); tunnel status reads the real unit state and URL. Keep to `mcp/`, the skills/apps/tunnel routes and their libs. PR base beta, do not merge.

### 1. HIGH — `skill_uninstall` reported CONFLICT after a *successful* uninstall (`mcp/tools/skills.ts`)

A regression #435 introduced. The post-condition was `after?.some(sk => sk.name === name)` — name only. A store skill can **shadow** a builtin of the same name (the README's own worked example: `skill_install official/pdf` → `skill_uninstall pdf`), and the installed list is keyed by name, so removing the store copy does not empty the name — the builtin resurfaces under it with `origin: "builtin"`. Live: the hub entry *was* removed and the tool answered `CONFLICT "The device did not remove \"pdf\" — it is still installed"` with `next: "Do not retry. Tell the user the device refused."`

Now only a surviving **store** row counts as a failure, and where the identifier is known, only the *same* identifier does. The pre-condition also prefers the removable store row when both are listed. On success after an un-shadowing, the tool says so — otherwise the agent reads the still-present name off a later `skill_list` as a failed uninstall.

### 2. MEDIUM — `app_uninstall` left the deployed webapp on disk and served (`src/app/setup-api/apps/uninstall/route.ts`)

The route removed the OpenClaw skill dir, the icon and the prefs, never `data/webapps/<appId>/`. Live: after `app_uninstall qa-t453a-revalidate` the desktop said it was gone while `/setup-api/webapps?app=qa-t453a-revalidate` still returned 200 with the page. It now deletes the deployed directory (traversal-guarded), which is what `code_project_delete`'s own guidance already promises.

### 3. MEDIUM — `skill_search` returned ids `skill_install` could not resolve (`src/lib/hermes-skills.ts`, install + inspect routes)

Root cause, from the box's own index and the Hermes CLI source:

- ClawHub is the **only** registry in `hermes-index.json` with unprefixed identifiers — all **69 150** of its rows are a bare slug with empty `repo`/`path`. That is 69 150 of 90 605 rows: three quarters of the store. `official`/`github`/`skills.sh`/`lobehub`/`browse-sh` are all prefixed.
- `hermes skills install` routes a **slash-less** argument through `_resolve_short_name()` (`hermes_cli/skills_hub.py:585`), which requires an exact match on the catalog **NAME**. A ClawHub row's name is its display name (`"QR Code Decode"`), its identifier the slug (`qrcode-decode`). Nothing ever matched, so the CLI printed a "did you mean…" table, **exited 0 having installed nothing**, and our lock check answered 502 `"Skill could not be resolved"` — for an id `skill_search` had just handed out. The tool's `next` ("pass the exact id it returned") made that a guaranteed retry loop.
- `ClawHubSource._parse_identifier` accepts `clawhub/<slug>`, and the slash makes the CLI skip short-name resolution entirely.

New pure helper `cliInstallIdentifier(id, source)` maps a bare slug to `clawhub/<slug>`; the install route uses it for the CLI argument only (`id` stays as the customer clicked it for the catalog lookup, the lock check and the audit log), and the lock check accepts either spelling. The same mapping is applied to the inspect route's docs phase, where `hermes skills inspect <bare slug>` dead-ended every clawhub card on "Skill not found" for the identical reason. This fixes the Settings → Skills Install button as well as the MCP tool.

### 4. LOW — `ai_list_models` reported `in_use` as two empty strings (`mcp/tools/ai.ts`)

`??` only guards null/undefined; `/setup-api/hermes/models` answers with empty **strings** on an unconfigured device, so the `"unknown"` fallback never fired and the tool returned `{"provider":"","model":""},"thinking":""`. Now blank and whitespace-only fields read `"unknown"`.

### 5. LOW — MCP `write_file`/`read_file`/`edit_file` echoed a bare basename (`mcp/tools/coding.ts`)

A relative `file_path` resolves against `CLAWBOX_ROOT`, not the spawning process's cwd, so `"Created qa-t453a-rel.txt"` identified neither of the two plausible trees. All headline echoes are now the resolved absolute path. (`list_directory` already did this.) OpenClaw-only family, dormant on Hermes.

### 6. HIGH — `/setup-api/tunnel/status` denied a tunnel that was serving the public internet (`src/lib/tunnel.ts`, `src/lib/cloudflared.ts`)

`src/lib/tunnel.ts` only ever recognised a `cloudflared` it had spawned itself and keyed everything off `data/tunnel.pid`. The shipped device runs `clawbox-tunnel.service`, whose `scripts/run-tunnel.sh` writes no pid file. Live:

```
$ systemctl is-active clawbox-tunnel  -> active
$ cat data/cloudflared/tunnel.url     -> https://stat-door-tournament-resorts.trycloudflare.com
$ ls data/tunnel.pid                  -> does not exist
$ curl -b jar .../setup-api/tunnel/status
  {"enabled":false,"running":false,"tunnelUrl":null,"error":null,"cloudflaredInstalled":true}
```

The owner is told remote access is off while the whole desktop is on the public internet behind one password — in the one state where they most need the URL and the off switch.

`getTunnelStatus()` now consults `systemctl is-active clawbox-tunnel` first and reads the URL from `data/cloudflared/tunnel.url`, falling back to the unit's journal (new `readTunnelUrlFromJournal()`, quick-tunnel hostnames only, newest wins). It adds `service` and `managedBy` so support can tell "off" from "on, run by systemd". Two write-path lies of the same class go with it: `startTunnel()` no longer spawns a **second** cloudflared alongside a running unit (that published a second public hostname nothing tracked or stopped), and `stopTunnel()` says it cannot stop the unit rather than returning `{success:true}` — naming Settings → Remote Control, which can.

## How verified

All commands run in the worktree at `83f43c5`.

**Full suite, build, tool contract, coverage:**

```
$ bun run test
  Test Files  397 passed (397)
        Tests  5886 passed (5886)      # run 3x after the rebase onto a63bec6

$ bun run build
  ✓ Compiled successfully — build-info.json: commit=2865760 buildId=MRii7EqAsrmg6idVehLSw

$ bun run mcp/check-tools.ts
  Tool contract OK.                     # 40 hermes / 52 openclaw unchanged

$ npx vitest run --coverage
  All files | 78.5 stmts | 70.05 branch | 78.01 funcs | 80.81 lines
  # ratchet is 63 / 52 / 61 / 65 — no threshold error
```

**Every new test was confirmed to FAIL against the pre-fix code** (restore the old file, run, restore the fix):

```
$ git show HEAD~1:src/app/setup-api/apps/uninstall/route.ts > route.ts && npx vitest run …/uninstall-webapp.test.ts
  × deletes data/webapps/<appId>/ so the URL stops serving
  × leaves every other deployed webapp alone
  Tests  2 failed | 2 passed (4)

$ git show HEAD~1:mcp/tools/skills.ts > skills.ts && npx vitest run …/mcp-tool-honesty.test.ts
  × reports success when removing a store skill that was shadowing a builtin
  × prefers the removable store row when a builtin of the same name is listed too
  Tests  2 failed | 31 passed (33)

$ git show HEAD~1:src/app/setup-api/hermes/skills/install/route.ts > route.ts && npx vitest run …/skills-install-clawhub.test.ts
  × installs a bare ClawHub slug that search returned
  × works on a device whose catalog index has not been built yet
  Tests  2 failed | 2 passed (4)
```

**Root-cause evidence gathered read-only from the QA box (.71), nothing mutated:**

```
$ python3 … hermes-index.json
  total 90605
  Counter({'clawhub': 69150, 'skills.sh': 19967, 'lobehub': 505, 'browse-sh': 440,
           'github': 438, 'official': 104, 'claude-marketplace': 1})
  clawhub  total 69150  no-slash 69150     # every other source: no-slash 0
  QR: {"name":"QR Code Decode","source":"clawhub","identifier":"qrcode-decode","repo":"","path":""}

$ curl https://clawhub.ai/api/v1/skills/qrcode-decode   -> 200  (the registry HAS it)
$ sed -n 582,586p hermes_cli/skills_hub.py
  # If identifier looks like a short name (no slashes), resolve it via search
  if "/" not in identifier:
      identifier = _resolve_short_name(identifier, sources, c)
      if not identifier: return          # exits 0, installs nothing
$ sed -n 2455,2470p tools/skills_hub.py
  """Accepts a bare slug, ``clawhub/<slug>``, ``@owner/slug``, …"""
```

**Test files added / extended:**

| file | covers |
|---|---|
| `src/tests/unit/mcp-tool-honesty.test.ts` (+5) | builtin/store name collision both ways; the same-identifier survivor still fails; `ai_list_models` blanks and whitespace |
| `src/tests/routes/apps/uninstall-webapp.test.ts` (new, 4) | real temp dir: webapp gone, neighbours untouched, no-webapp case, traversal refused |
| `src/tests/unit/hermes-skill-install-id.test.ts` (new, 6) | fake search fixture — one row per source, every returned id maps to something installable and still passes `checkInstallIdentifier` |
| `src/tests/routes/hermes/skills-install-clawhub.test.ts` (new, 4) | faithful CLI fake (slash-less ⇒ "did you mean" + exit 0 + empty lock); asserts the argv the route hands the CLI |
| `src/tests/unit/mcp-file-tool-paths.test.ts` (new, 4) | write→read→edit round trip on a relative path under a temp `CLAWBOX_ROOT` |
| `src/tests/unit/tunnel.test.ts` (+6) | the exact live shape (active unit, no pid file), journal fallback, `activating`, `failed`/`inactive`, spawned tunnel, both write-path guards |
| `src/tests/unit/cloudflared.test.ts` (+4) | journal reader: newest hostname, no-URL, journalctl missing, non-quick-tunnel host rejected |
| `src/tests/routes/tunnel.test.ts` (+1, 2 updated) | route passes `service` / `managedBy` / URL through |

## What is NOT covered

- **The skills ROUTES still fabricate.** `POST /setup-api/hermes/skills/uninstall` still answers `200 {"ok":true}` for a name it never removed, and `GET …/inspect?id=<nonsense>` still synthesises a record — so Settings → Skills (`HermesSkillsStore.tsx:189-208`, `useSkillDetail.ts:94-151`) still shows a false success and an invented skill. Only the MCP layer was hardened, in #435 and here. Separate defect, separate UI surface; not in this brief.
- **No live verification on .71.** The box is owned by a hardware-test lane that reboots it and runs the updater, and installing/uninstalling a skill or bouncing `clawbox-tunnel` are mutations the brief forbids. Everything above is read-only inspection of the box's index, the Hermes CLI source and the ClawHub API, plus tests. The ClawHub install path is proven against a faithful fake of the CLI, not against the real one.
- **`stopTunnel()` still cannot stop the unit** — it now refuses honestly and points at `POST /setup-api/portal/stop`. Making the legacy `/setup-api/tunnel/*` surface actually control the unit (or deleting it, which round 1 suggested) is a larger call about which surface owns remote access.
- **`clawkeep-restore-ungated`** (revalidate §7, NOT-FIXED) is untouched — it is a ClawKeep route, not in this brief.
- **Degraded-catalog provenance.** When the index has not been built, a bare slug is mapped to `clawhub/` with no catalog record to confirm the source. That is deliberate: ClawHub is the only registry that emits bare slugs, and Hermes' own fallback is a cross-registry fuzzy match its authors call provenance-unsafe. A bare id declared as any *other* source is left alone.
- One full-suite run out of four showed `1 failed` and did not name the test before I could capture it; three consecutive re-runs were green (5886/5886). I could not reproduce it and do not believe it is mine, but flagging it rather than omitting it.

## Summary by CodeRabbit

* **New Features**
  * Tunnel status now shows service state, management method, and published URL.
  * File tools report complete resolved paths.
  * Skill installation supports ClawHub identifiers and improved removal of store-installed skills.
  * Uninstalling an app also removes its deployed files safely.
* **Bug Fixes**
  * Prevented duplicate tunnel processes and accurately report unsupported stop requests.
  * Empty AI model details now display as “unknown.”
  * Improved skill inspection and installation across identifier formats.
* **Documentation**
  * Updated Remote Control documentation with current tunnel behavior and limitations.


KrasimirKralev:
Heads-up on a direction change that touches one line of this PR's write-up, so it doesn't surprise you later.

Your "What is NOT covered" section notes `clawkeep-restore-ungated` (revalidate §7) as untouched, correctly out of scope here, and the TASK-453 approach around ClawKeep was to **edition-gate it** on Hermes. The owner has since asked for the opposite: **make ClawKeep actually work on Hermes** rather than gate it off. That is now #500.

To be explicit about the boundary: **#500 changes no file this PR touched and reverts none of your fixes.** The skill uninstall/collision work, the webapp leak, the ClawHub identifier mapping, `ai_list_models`, the MCP file-path echoes and the tunnel-status work are all untouched. The only overlap is conceptual — the ClawKeep edition gate from `8c82c72` is removed, because the feature is no longer absent on that SKU.

`clawkeep-restore-ungated` is resolved by #500, in the other direction: the restore route had no edition gate while the UI gated the whole app off. Now that ClawKeep archives either agent, an ungated route is the correct state, and the honesty work moved to where it belongs — a cross-edition snapshot is refused by name before anything is verified or moved, since one portal account gets one R2 prefix and the snapshot list legitimately contains the other edition's backups (including a converted box's own).

Two things from your work fed straight into #500 and are worth recording:

- **The Hermes SKU has no `clawbox-gateway`.** ClawKeep restarted `clawbox-gateway.service` after every restore. Install removes and masks that unit on Hermes, so the restart always failed while `clawbox-hermes-dashboard` — the process actually holding the restored files open — was never touched: the restore reported success and the agent kept serving pre-restore state. Same false-success class as the tunnel-status finding in this PR, and I would not have looked there without it.
- **Your WAL warning on the QA box was right and pre-empted a real bug.** 2.6 MB of `-wal` against a 2.8 MB `state.db`. The archived copy goes through sqlite's online-backup API so it is checkpointed and self-contained, but restore was leaving the *old* database's sidecars beside it, and sqlite replays a WAL because the file is there, not because it matches. Fixed and regression-tested before it ever ran. Thank you.

Also worth flagging for whoever picks up the skills routes: your NOT-covered note that `POST /setup-api/hermes/skills/uninstall` still answers `200 {"ok":true}` for a name it never removed, and `inspect` still synthesises a record, is still open. #500 does not touch it.

KrasimirKralev:
Cross-posting a direction change on one item from this task's notes, so it is on the record next to the work rather than only in a PR that supersedes it.

**TASK-453's ClawKeep line was "edition-gate ClawKeep" on Hermes. The owner has since asked for the opposite: make it work there.** That has now landed as #500 — ClawKeep backs up and restores the Hermes agent (`~/.hermes`: `config.yaml`, `.env`, `state.db`, memories, skills, plugins, hooks, cron, pairing, plus the shared identity bridge), with the ~1.5 GB `hermes-agent/` checkout, the venv and every cache deliberately excluded.

**Nothing in #474 was touched or reverted.** This PR's own "What is NOT covered" already drew the boundary — *"`clawkeep-restore-ungated` (revalidate §7, NOT-FIXED) is untouched — it is a ClawKeep route, not in this brief"* — and no file in #500 overlaps this diff. The `skill_uninstall` collision, the webapp leak, the ClawHub identifier mapping, `ai_list_models`, the MCP path echoes and the tunnel-status work are all untouched and still stand.

On `clawkeep-restore-ungated` specifically: it **dissolves** rather than getting fixed. The ClawKeep *routes* were never edition-gated — only the UI was (`8c82c72` gated the app behind `supportedOnEdition`, leaving the routes reachable). That asymmetry was a defect while the feature could not work on Hermes. Now that it can, ungated routes are the correct state, and the UI gate is what had to go.

Two things from this task's evidence turned out to matter directly to the ClawKeep work, so credit where it is due:

- The **tunnel-status finding's shape** — a route that keys off state it spawned itself rather than the state the shipped device is actually in — is exactly the bug class I hit in ClawKeep's restore path: it restarted `clawbox-gateway` after a Hermes restore, and install *removes and masks* that unit on Hermes, so the restart always failed while `clawbox-hermes-dashboard` (the process actually holding the restored files open) was never touched. Restore reported success and the agent kept serving pre-restore state. Same shape, different route.
- Running against real hardware rather than a fixture found three more that reading the code did not: empty directories (`hooks/`, `pairing/`) aborting a whole restore, absolute symlinks (the identity bridge) being dropped **while restore still reported success**, and stale sqlite WAL sidecars surviving a swap. This PR's insistence on live evidence over inference was the right instinct and I followed it.

@Yalexx-ID no action needed here — this is a note, not a request to change anything in #474.

PR #477, opened 2026-08-24
Verify the setup-time Hermes OAuth routes are behind the box password, and pin them there (TASK-527)

KrasimirKralev:
Task: TASK-527 — lock the setup-time Hermes provider-OAuth routes behind the box password.

**Outcome: this is a verification, not a new gate.** TASK-443 (#436, merged into
beta earlier today) already closed the hole. The honest deliverable is a live
verification plus the regression coverage that was missing, and no second fence
across the same door.

---

## The exposure as reported

During first boot the device broadcasts an **open `ClawBox-Setup` AP**, and the
inline provider-OAuth relay lives under `/setup-api/hermes/oauth/`:

| route | what it does |
|---|---|
| `start` | opens a real OAuth session against the owner's Hermes dashboard and hands back the provider's `auth_url` / `user_code` |
| `submit` | feeds that session the authorization code the user pasted — the step that mints a stored provider credential |
| `cancel` | destroys an in-flight session |
| `poll` | reads a session's status back |

The gate in front of `/setup-api/*` used to be a **deny-list**
(`PRE_AUTH_SENSITIVE_PREFIXES`) evaluated whenever `setup_complete !== true`.
It was written around the read-only `setup/status` route, and every path nobody
had remembered to name was served with no credential. None of these four was
ever named, so all four answered anyone in radio range for the whole first-boot
window.

## Why nothing needs re-gating

TASK-443 **inverted that deny-list to an allow-list** (`src/lib/setup-api-gate.ts`).
The list holds only what wizard steps 1–3 call; the default is now 401. The
OAuth family is not on it and does not belong on it — the flow is step 4
(AIModelsStep), which runs after CredentialsStep has set the password and been
handed a session cookie, so the wizard reaches it authenticated and needs no
carve-out.

Both halves confirmed on the bench box (hermes edition, provisioned, app on :80):

```
# no session
GET    /setup-api/hermes/oauth                  -> 401
POST   /setup-api/hermes/oauth/start            -> 401
POST   /setup-api/hermes/oauth/submit           -> 401
DELETE /setup-api/hermes/oauth/cancel           -> 401
GET    /setup-api/hermes/oauth/poll?...         -> 401
GET    /setup-api/setup/status                  -> 200   (public by design)

# after POST /login-api with the box password
GET    /setup-api/hermes/oauth                  -> 200   provider list
POST   /setup-api/hermes/oauth/start            -> 200   live PKCE session + auth_url
DELETE /setup-api/hermes/oauth/cancel           -> 200   reaches the dashboard
POST   /setup-api/hermes/oauth/submit           -> 404   "Unknown or expired session" (synthetic id)
GET    /setup-api/hermes/oauth/poll?...         -> 404   "Session not found or expired" (synthetic id)
```

So the routes refuse a stranger and still work for the owner — the wizard's own
sign-in flow is intact, end to end, against the real dashboard.

**`poll` stays gated too.** The brief allowed leaving the read path exempt if it
returns only status booleans. It does relay only non-credential fields
(`session_id`, `status`, `error_message`, `expires_at`, whitelisted by
`relayJson`), but under an allow-list the exemption would have to be *added*
rather than removed, and there is no caller that needs it: the session id it
polls can only have come from `start`, which is gated. Leaving it off the list
is both stronger and less code.

## What this PR actually adds

Coverage for the two ways the gate could quietly stop holding, plus the
rationale next to the list so the next editor sees it.

- **The allow-list holds PREFIXES.** One plausible addition —
  `/setup-api/hermes`, or the oauth path itself, to read the provider list a
  step earlier — reopens all four leaves at once, and beta's existing assertion
  on the parent path `/setup-api/hermes/oauth` would still pass. Now asserted
  against the list directly, and the four leaves are named individually in
  `middleware.test.ts` rather than inferred from the parent.
- **The wizard handoff.** The cookie `system/credentials` mints on the initial
  password set could stop being one middleware accepts. The existing suites
  check those halves separately and neither would notice they no longer meet —
  and that failure strands a fresh box mid-wizard, with no way in but a password
  the user chose thirty seconds ago. Now driven end to end with `@/lib/auth`
  unmocked, so the cookie under test is the one the route really signs.
- Also pinned: the post-setup 401 (a different code path — the session gate, not
  the allow-list), and the trailing-slash and mixed-case spellings.

The comment added to `setup-api-gate.ts` records why the subtree's absence from
that list is load-bearing rather than an oversight.

## Deliberately not done

An in-handler `requireSession` on `start` / `submit` / `cancel`, the way TASK-443
gave the destructive handlers a self-contained second line via `@/lib/route-auth`.
There is a real argument for it — credential-minting routes are at least as
sensitive as destructive ones, and 443's own reasoning is that one hand-maintained
list in front of a ~100-route surface is one `startsWith` away from serving a
route it meant to refuse. It was left out because the door is already shut and a
second, divergent gate on the same paths is worth an explicit decision rather
than a side effect of this task. The structural test above covers the same
regression at the list. **Happy to add it if you want it — say the word.**

## Verification

Run on the bench box in an isolated clone (`git clone --shared` + hardlinked
`node_modules`), against a clean checkout of `origin/beta` as the baseline.
Nothing deployed, no service touched, `~/clawbox` untouched.

| | beta `4cbc4af` | this branch `1c87d36` |
|---|---|---|
| `npx vitest run --config vitest.config.ts` | 400 files, 5924 tests | **401 files, 5933 tests — 0 failed** |
| `npx tsc --noEmit` | 21 errors | **21 errors — the identical set**, diff empty |
| `npx eslint <changed files>` | — | clean, exit 0 |

The +1 file and +9 tests are this PR's own (5 in the new file, 4 leaves added to
`middleware.test.ts`'s existing case). All 21 tsc errors are pre-existing and in
`src/tests/unit/*` files this PR does not touch. CI is green on this head:
`test`, `e2e`, `e2e-install`, `Analyze`, `CodeQL`, `build-identity`.

### The guards were checked against the regression they claim to catch

Adding `"/setup-api/hermes/oauth"` to `BOOTSTRAP_ALLOWED_PREFIXES` on the device
fails **7 assertions** — the behavioural one (`refuses a stranger on the open AP
for the whole pre-setup window`), the structural one (`keeps the OAuth subtree
off the bootstrap allow-list itself`), and all five leaf cases in
`middleware.test.ts`. Reverted after the check; tree clean.

### About the flaky runs

Two full-suite runs on the box while it was loaded (load average 8–14, other
work running alongside) reported failures — 7 tests across 5 files on the
branch, and separately **4 tests across 3 files on untouched beta**, a different
set each time. Every one was a `.test.tsx` component/timing suite; none was in
this PR's diff, and none was `middleware.test.ts` or the new file.

Pinned down rather than waved off. `chat-spoken-reply.test.tsx` was the
repeatable one: byte-identical on both trees, it failed 2 of 3 isolated runs at
load ~12 and then passed **4/4 on beta and 4/4 on the branch** at load ~5. The
run recorded in the table above is the branch at load 3.66 — fully green. The
three changed suites were also run 10× consecutively: 10/10 clean.


KrasimirKralev:
@coderabbitai review

KrasimirKralev:
Closing: #478 is merged to beta and supersedes this for TASK-527. The setup-time Hermes OAuth routes are gated there, so this branch would only re-land the same guard against a moved base.

PR #485, opened 2026-08-25
Coding agent: delegate coding tasks to a headless Claude Code run

yalexx:
## Coding agent: let the assistant delegate a whole coding task to a headless Claude Code run

The desktop's "Coding Agent" app has been a terminal typed into `claude-ds` (Claude Code on the box's own ClawBox AI plan). This turns that same wrapper into something the **assistant** can use: it hands over a task, `claude-ds -p` works in the background inside one folder, and the summary comes back through the MCP tools.

### What's in it

- **`src/lib/coding-agent.ts`** — the runner. Lives in the web server (the MCP process is reaped after ten idle minutes; a run outlives that). Spawns `~/.local/bin/claude-ds -p --output-format stream-json …` by absolute path with an **explicit environment** (a login shell's PATH, `CLAWBOX_ROOT`; never `SESSION_SECRET`/`CLAWBOX_MCP_TOKEN`), task on stdin, parses the stream into a run record persisted at `data/coding-agent-runs.json` (0600, temp+rename), and settles runs the previous server left "running" at boot.
  - `--permission-mode acceptEdits`, `--tools` cut to files/search/Bash, a Bash **allow-list** (build/test/package tooling, read-only git) plus an explicit **deny-list** (`sudo`, `rm`, `curl`, `git push`, `systemctl`, the device CLIs…), and Read/Edit/Write **deny rules** for the credential folders `file-guard` protects and this checkout's `data/`. `--setting-sources user` keeps the OS checkout's own CLAUDE.md out of a project that sits under it. Every denial Claude Code would have asked about is silently refused in `-p` mode and **counted** on the run.
  - Working folder: a code project by default; any other folder must be inside the home, not protected, and **not the ClawBox checkout itself**.
  - One run at a time, 20 min / 60 turns / cost cap; `resume_run_id` continues a finished session in the same folder.
- **`/setup-api/coding-agent/{status,enable,run,runs,stop}`** — `enable` is **owner-only** (`hasOwnerSession`, refuses the MCP bearer: the agent must not grant itself a delegated shell). `run` answers **409** while the switch is off / harness not ready / busy so the MCP maps it to CONFLICT-do-not-retry. `runs?id=&wait=` long-polls up to two minutes.
[...]

PR #489, opened 2026-08-26
Coding agent: start a run with no Linux capabilities

yalexx:
Follow-up to #485. Found by the adversarial review pass of the merged change, confirmed on the device — and then **partly refuted by the review's own verification stage**, which changed why this is worth doing. Both halves are below, because the first framing was wrong and the record should say so.

### What is measurably true

`clawbox-setup.service` grants the web server three **ambient** capabilities so it can manage WiFi and bind port 80:

```
config/clawbox-setup.service:47
AmbientCapabilities=CAP_NET_BIND_SERVICE CAP_NET_ADMIN CAP_NET_RAW
```

Ambient capabilities are inherited across `execve` — that is what they are for — so every coding run started holding all three:

```
web server  (pid 72052)  CapAmb: 0000000000003400   <- NET_BIND_SERVICE|NET_ADMIN|NET_RAW
gateway     (pid 63862)  CapAmb: 0000000000000000
```

Reachable from inside a run, not theoretical. `python3` is on the Bash allow-list, so no Claude Code tool policy applies to what the interpreter then does:

```
run-93nboaal  status=completed  turns=2  permissionDenials=0
summary: ['CapInh:\t0000000000003400', 'CapPrm:\t0000000000003400',
          'CapEff:\t0000000000003400', 'CapBnd:\t000001ffffffffff',
          'CapAmb:\t0000000000003400']
```

(The file-access guard held throughout — the same probe via `grep`/`cat`/Read was refused five times. The interpreter is the path that walked around it.)

### What I first claimed, and why it was wrong

I originally called this a breach of the bar this feature is held to — *"no more than the agent already has through its own shell tool"*. A verifying agent refuted that, and checking on the box, it is right:

- `config/clawbox-sudoers` grants `clawbox` NOPASSWD `systemctl start clawbox-root-update@*.service`; `config/clawbox-root-step.sh:97` ends `exec /bin/bash "$ENTRYPOINT" --step "$step"` with `ENTRYPOINT=$PROJECT_DIR/install.sh`, and `install.sh` is `-rwxr-xr-x clawbox clawbox` — writable by the same user.
- The agent's own `bash` tool applies no sudo filtering (`mcp/lib/jobs.ts`, `mcp/tools/coding.ts`).

So on the OpenClaw edition the agent's shell can already reach **root**, which subsumes these three capabilities. `mcp/README.md` says as much in plain words: *"OpenClaw + `bash` = the agent can reach anything the device user can."* This was never an escalation relative to the agent.

### Why it still lands

Least privilege, not a breach fix. The coding run is the *unattended, prompt-driven* path — the one with an explicit Bash allow-list, file deny rules and folder containment, all built on the premise that a delegated run gets only what the task needs. A run that builds and tests code in one folder has no use for `CAP_NET_ADMIN` (reconfigure interfaces, routes, firewall) or `CAP_NET_RAW` (raw sockets), and it was getting them for free from an unrelated unit setting. Dropping them costs nothing and stops an unrelated configuration change from silently following the web server into every run.

### The change

```
setpriv --ambient-caps=-all --inh-caps=-all --no-new-privs -- ~/.local/bin/claude-ds -p …
```

Both sets emptied before the run starts; `--no-new-privs` stops it regaining through a setuid binary what those flags took away. `setpriv` is util-linux, on every ClawBox; readiness counts it among the things a run needs and `startRun` resolves it again and **refuses** rather than falling back to a spawn that keeps the capabilities. No existing guard was relaxed.

### Verified on hardware, after the fix

Same probe, same allow-listed interpreter, rebuilt box — see the comment below: `CapAmb: 0000000000000000`, run completes normally.

### Tests

`buildSpawnArgv` is pinned exactly (flags before the separator, wrapper after — get that order wrong and setpriv hands the flags to the wrapper), readiness reports the new requirement, and the existing fake-wrapper integration tests now exercise the whole setpriv chain: a malformed invocation would stop the wrapper running and fail them all.

### DIFF
--- a/mcp/README.md
+++ b/mcp/README.md
@@ -272,6 +272,13 @@ shell tool, not less and not more:
   the run is using — but never the run's own `data/code-projects/<id>`), are
   denied to Claude Code's own Read/Edit/Write. A guard rail, not a sandbox —
   the same caveat as `bash`;
+- the run holds **no Linux capabilities**. `clawbox-setup.service` gives the web
+  server `CAP_NET_BIND_SERVICE`, `CAP_NET_ADMIN` and `CAP_NET_RAW` ambiently for
+  WiFi management and port 80, and ambient capabilities are inherited across
+  `execve` — so a run used to start with all three while the agent's own shell
+  tool, spawned by the gateway, had none. The wrapper is now spawned through
+  `setpriv --ambient-caps=-all --inh-caps=-all --no-new-privs`, and a box
+  without `setpriv` reports not-ready rather than running with them;
 - the folder must be a code project or a directory inside the home that is
   neither protected nor the ClawBox checkout itself, so a prompt-injected
   "fix the OS" cannot edit the running product in place;
--- a/src/components/CodingAgentPanel.tsx
+++ b/src/components/CodingAgentPanel.tsx
@@ -25,6 +25,10 @@ interface Readiness {
   wrapperInstalled: boolean;
   claudeInstalled: boolean;
   clawaiConnected: boolean;
+  /** setpriv, which strips the web server's network capabilities off a run.
+   *  Not given a row of its own: it is present on every ClawBox, and when it
+   *  is not, `problems` says so in the owner's words. */
+  capabilityDropAvailable: boolean;
   problems: string[];
 }
 
--- a/src/lib/coding-agent.ts
+++ b/src/lib/coding-agent.ts
@@ -33,6 +33,18 @@
  *     Claude Code's own Read/Edit/Write as well. That is a guard rail against a
  *     mistake, not a sandbox: a shell can spell a path in ways no pattern list
  *     enumerates, exactly as mcp/README.md says of `bash`.
+ *   - The run starts with NO Linux capabilities. `clawbox-setup.service` grants
+ *     the web server `CAP_NET_BIND_SERVICE`, `CAP_NET_ADMIN` and `CAP_NET_RAW`
+ *     as AMBIENT capabilities so it can manage WiFi and bind port 80, and
+ *     ambient capabilities are inherited across execve by design — so without
+ *     this a run held `CapAmb=0x3400` while the agent's own shell tool, which
+ *     the gateway spawns, held none. Measured on a real box: a run asked for
+ *     `python3 -c "…/proc/self/status…"` — an allow-listed interpreter, so no
+ *     tool policy applied — and printed the three capabilities back. That is
+ *     more power than the bar this feature is held to, so the wrapper is
+ *     spawned through `setpriv` with the ambient and inheritable sets emptied
+ *     and no-new-privs set. Readiness refuses to start a run when `setpriv` is
+ *     missing rather than quietly running with the capabilities.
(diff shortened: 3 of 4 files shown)


yalexx:
## Verified on the box (192.168.50.65), after the fix

Rebuilt on this branch and restarted: `build-info.json` → `commit=a6bb1be branch=fix/coding-agent-drop-capabilities`.

Readiness now counts the capability stripper:

```json
{"enabled":true,"ready":true,
 "readiness":{"ready":true,"wrapperInstalled":true,"claudeInstalled":true,
              "clawaiConnected":true,"capabilityDropAvailable":true,"problems":[]},
 "running":0,"harnessCommand":"claude-ds","maxTaskChars":4000}
```

Then the **same probe as before the fix** — same task, same allow-listed `python3`, so nothing about the measurement changed except the code under it:

| | run | permission denials | what the run printed |
|---|---|---|---|
| before | `run-93nboaal` | 0 | `CapAmb: 0000000000003400` |
| after | `run-shy6zu77` | 0 | `CapAmb: 0000000000000000` |

After, verbatim from the run's own summary:

```
['CapInh:\t0000000000000000', 'CapPrm:\t0000000000000000',
 'CapEff:\t0000000000000000', 'CapBnd:\t000001ffffffffff',
 'CapAmb:\t0000000000000000']
```

Inheritable, permitted, effective and ambient are all empty. The bounding set is untouched, which is expected — `setpriv` was not asked to shrink it, and an empty ambient set is what decides what the process actually holds.

The run completed normally in 2 turns, so the `setpriv` chain execs the wrapper and the wrapper execs Claude Code as before: dropping the capabilities cost the feature nothing.

yalexx:
**Correction to my own framing, for the record.**

I opened this claiming the inherited capabilities exceeded the bar this feature is held to. The review's verification stage refuted that, and it is right — I checked on the box:

- `config/clawbox-sudoers` lets `clawbox` start `clawbox-root-update@*.service` without a password; `config/clawbox-root-step.sh:97` `exec`s `$PROJECT_DIR/install.sh` as root; `install.sh` is owned and writable by `clawbox`.
- The agent's own `bash` tool filters no sudo at all.

So the agent's shell on an OpenClaw box already reaches root, which subsumes `CAP_NET_ADMIN`/`CAP_NET_RAW`. `mcp/README.md` states this openly ("the agent can reach anything the device user can"). The capability inheritance was **not** an escalation relative to the agent, and I have rewritten the PR description to say so.

The change still stands on least privilege: an unattended run confined to one folder by an allow-list has no use for interface, routing or raw-socket powers it was inheriting from an unrelated unit setting, and dropping them costs the feature nothing (verified: the run completes normally with all four sets empty).

Separately, and out of scope for this PR: this dev box also carries `/etc/sudoers.d/clawbox-nopasswd` → `clawbox ALL=(ALL) NOPASSWD: ALL`. It is **not** shipped by this repo — nothing in `install.sh` or `config/` writes it — so it looks like a local convenience on this spare unit. Worth knowing when reasoning about what an agent can do here versus on a customer device.

PR #495, opened 2026-08-26
Make the sudoers quarantine prove the allow-list landed, and fix the Hermes gateway restart (TASK-445 follow-up)

KrasimirKralev:
Follow-up to #471 — **targets `fix/hermes-sudoers-445-r2`, not `beta`**, so it lands with the narrowing rather than after it. Please review it as part of #471; it is not independently mergeable.

Closes the two blockers from the deep review of #471, plus a documentation correction. The agent-capability research that motivated the second one is posted as a comment on #471.

---

## B1 — a failed sudoers install could still trigger the quarantine

`install_sudoers_dropin` (install.sh:2680) could return **0 after an `install(1)` that never wrote anything**, and `step_systemd_services` used that success as permission to remove `/etc/sudoers.d/90-clawbox-nopasswd`. On a device whose only grant is the blanket one — which is every shipped box — the sequence ends with **neither** file: no working sudo at all, on an appliance with no console.

Two defects combined:

**1. Both call sites ran the function in a condition context.** install.sh:2950 (`if install_sudoers_dropin …; then`) and install.sh:2961 (`… || echo`). Bash suspends `set -e` for the whole *dynamic extent* of a command being tested, so every unchecked line inside the function body was silently non-fatal too:

| install.sh | unchecked | what it hides |
|---|---|---|
| 2694 | `cat "$src" > "$staged"` | ENOSPC → a truncated allow-list that still **parses**, so `visudo -cf` passes and it installs |
| 2716 | `cat "$dest" > "$backup"` | a backup that cannot restore |
| 2723 | `install … "$staged" "$dest"` | **the blocker** |
| 2728 | rollback `install … "$backup" "$dest"` | prints `rolled … back` whether or not it did |

**2. The function's own success signal could not see the failure.** The closing `visudo -c` (install.sh:2724) re-reads whatever is on disk, so after a failed install it validates the **old** file and answers 0.

### Fix

* Every step inside the function carries its own check — the function no longer depends on `set -e` at all.
* The destination is verified **byte-for-byte** against the staged candidate (`cmp -s`). A return code cannot distinguish "wrote nothing" from "wrote everything"; a comparison can, and it also catches the truncation case that `visudo` structurally cannot.
* The caller invokes it **plainly** and reads an explicit `sudoers_status`, so the condition-context trap cannot come back.
* The quarantine is gated on a second, load-bearing proof: `cmp -s "$PROJECT_DIR/config/clawbox-sudoers" "$SUDOERS_DIR/clawbox"` — a fact about the **device**, not about a code path.
* The rollback now reports what actually happened, so a stranded device is distinguishable from a safely-restored one in the install log.

### Tests

`src/tests/unit/install-sudoers-migration.test.ts` gains a suite that lifts the **real gate** out of `step_systemd_services` (rather than re-implementing it, so it cannot drift) and runs it against a temp `/etc/sudoers.d` with a fake `install(1)` that can refuse (`INSTALL_FAIL_DEST`) or exit 0 having written a prefix (`INSTALL_TRUNCATE_DEST`). Three regression cases assert the blanket drop-in **survives**: refused install, truncated install, failing `visudo -c`. Plus a happy-path case asserting it *is* quarantined once the allow-list really landed.

---

## B2 — `hermes gateway restart` is not install-time-only, and it reported false success

The `EXEMPT_CALLS` rationale in `scripts/check-sudoers-coverage.sh:136` said the sudo'd Hermes CLI call is install-time-only and its failure non-fatal. **Both halves are wrong for the restart branch.**

* `ensureHermesGateway` takes `sudo: systemScope` on the **restart** branch too (`src/lib/hermes-telegram.ts:442`). That is the path every Telegram / WhatsApp / Discord / Email config save hits on a provisioned Hermes box — `telegram/configure/route.ts:69`, `discord/configure/route.ts:129`, `whatsapp/configure/route.ts:135`, `hermes-email.ts:113` and `:150`, `whatsapp-pairing.ts:682`.
* It execs `/home/clawbox/.local/bin/hermes`, which the clawbox user **owns and can rewrite** (verified on a device: `-rwxr-xr-x 1 clawbox clawbox`). That is why it could not be granted and had to be exempted — and therefore why the restart silently stops working once the blanket grant is gone.
* The failure was invisible: `runHermesCli` **resolves** on a non-zero exit (`src/lib/hermes-cli.ts:117` — it only rejects on spawn failure, timeout or abort), and the status check that follows runs `hermes gateway status` **without** sudo. It sees the still-running old process, answers `running: true`, and `telegram/configure/route.ts:75` replies **`{restarted: true}`**. The owner's new bot token then quietly kept not working.

### Fix — a zero-escalation grant

`hermes gateway install --system --run-as-user clawbox` writes a **root-owned** unit that runs `User=clawbox`. Verified against the installed CLI's unit template (`hermes_cli/gateway.py:3514`):

```
[Service]
User={username}          # = clawbox
Group={group_name}
ExecStart={venv python} -m hermes_cli.main gateway run
Restart=always
StartLimitIntervalSec=0
```

So `systemctl restart hermes-gateway.service` starts a process the clawbox user could have started itself — **strictly less** privilege than the existing `restart clawbox-gateway` grant, whose unit has no `User=` at all. `gateway.py:3878` prints `sudo systemctl restart <svc>` as the CLI's own remediation hint, so this is its sanctioned path.

* System scope now runs `sudo -n /usr/bin/systemctl restart hermes-gateway.service`.
* Two grants added (`config/clawbox-sudoers`), bare + `.service`, no wildcard — profile units (`hermes-gateway-<profile>.service`) are a separate privilege question and ClawBox only ever runs the default profile. No `start` (restart starts a stopped unit) and no `reset-failed` (`StartLimitIntervalSec=0` means there is no start limit to clear).
* User scope stays on the CLI — `systemctl --user` from a system service would target root's session bus — but its **exit code is now checked** instead of assumed.
* `gateway install --system` stays deliberately **ungranted**: it writes into `/etc/systemd/system` and has no safe Cmnd spelling against a clawbox-writable binary. It fails fast under `-n`, and the new `applied` flag carries that outward instead of it disappearing into a status probe.

A note on the unit name, since it will look odd on a box you check today: `hermes-gateway.service` does not exist until the first messaging platform is configured — that is what `ensureHermesGateway` installs. On a box that has never had Telegram/WhatsApp/Discord/Email set up, `systemctl show hermes-gateway.service` reports "no such file", and the grant is simply not exercised yet. It is not a dead line.

### Fix — stop faking the result

`ensureHermesGateway` now returns `HermesGatewayEnsureResult` = status **plus `applied`**: did the change it attempted actually take? Every caller requires `running && applied` before claiming a restart, and degrades to the existing *"Saved — will apply on next gateway restart"* warning otherwise. `stopHermesEmailPolling` gains a `"restart-failed"` outcome so `/email/configure` warns instead of reporting that receiving stopped.

Making `applied` a **required** field was deliberate: it forced every existing mock in the suite to state explicitly whether it was modelling a real restart or a refused one.

### Trade-off, stated plainly

The CLI's restart first attempts a SIGUSR1 **graceful drain** of in-flight turns (`gateway.py:4082-4160`); `systemctl restart` relies on the unit's own `KillSignal=SIGTERM` / `TimeoutStopSec` instead. The CLI already falls through to the identical forced `systemctl restart` whenever the drain does not finish in budget, so this is its own fallback path taken directly. Flagging it so it is a decision, not a surprise.

---

## Validation — what is live-verified and what is not

**Live-verified on a device (read-only, before the boxes went off-subnet):**
- `/home/clawbox/.local/bin/hermes` is `clawbox:clawbox 0755` — the binary the old path sudo'd is user-writable.
- The Hermes CLI unit template writes `User={username}` into a root-owned `/etc/systemd/system/hermes-gateway.service`, and `systemd_restart(system=True)` reduces to `systemctl restart <svc>`.
- Denied-sudo timing: **28 ms** with `-n`, **30 ms** without one in a no-TTY context (how the app spawns children); **unbounded hang** on `[sudo] password for …` when a PTY is attached. So `-n` is doing real work here.
- `/etc/sudoers.d/90-clawbox-nopasswd` (`clawbox ALL=(ALL) NOPASSWD: ALL`) is still live on the QA box, i.e. the migration this PR guards is the one that matters.

**Harness-verified on Linux (Ubuntu 22.04, real `visudo`, LF checkout of this branch):**
- `visudo -cf config/clawbox-sudoers` → `parsed OK`.
- `scripts/check-sudoers-coverage.sh` → `OK — 43 grants, 43 resolved sudo invocations, 0 gaps`.
- The B1 gate, run as install.sh really runs it: blanket quarantined on the happy path; blanket **survives** a refused install, a truncated install and a failing `visudo -c`. 8/8.

**NOT verified:** no live `/etc/sudoers.d` swap was performed. Both boxes moved off this machine's subnet mid-task, and the owner's box was mid-testing and off limits for sudoers changes throughout. I would rather say so than imply device coverage I do not have. A live install/update run on a box that still carries the blanket drop-in is the one check outstanding — it is also exactly what `e2e-install/06-sudoers.spec.ts` and the e2e-install container exercise.

**Test/lint:**
- `eslint` on every changed file: **0 problems**.
- `tsc --noEmit`: **0 new errors** vs the base branch (diffed error lists).
- `vitest`: 52 unit + 259 route tests across the affected areas pass. The full suite has pre-existing Windows-environment failures (missing `flock`, path separators) in unrelated files; none of the changed files appear in them.
- `src/tests/unit/install-sudoers-migration.test.ts` self-skips off Linux (`CAN_RUN` needs `/usr/sbin/visudo`), so its new cases were additionally run through a standalone harness on Linux — result above.

---

## Not addressed here, on purpose

`config/49-clawbox-updates.pkla` grants `unix-user:clawbox` the **unfiltered** `org.freedesktop.systemd1.manage-units` action with `ResultAny=yes`, which makes the systemd half of this allow-list advisory. Details, evidence and the JetPack constraint are in a separate comment on #471 — it needs a design decision rather than a patch, and it is yours to route.

## Summary by CodeRabbit

* **Bug Fixes**
  * Gateway configuration now reports success only when changes are actually applied, not merely when the gateway is running.
  * Added clear pending or failed-restart warnings across Discord, Telegram, WhatsApp, and email setup.
  * Email polling now distinguishes failed gateway restarts from successful shutdowns.

* **Security**
  * Restricted service restart permissions to the required gateway action.
  * Hardened permission installation and recovery to preserve existing rules when validation fails.

### DIFF
--- a/config/clawbox-sudoers
+++ b/config/clawbox-sudoers
@@ -62,6 +62,36 @@ clawbox ALL=(root) NOPASSWD: /usr/bin/systemctl enable  clawbox-tunnel
 clawbox ALL=(root) NOPASSWD: /usr/bin/systemctl disable clawbox-tunnel.service
 clawbox ALL=(root) NOPASSWD: /usr/bin/systemctl disable clawbox-tunnel
 
+# Hermes' messaging gateway. This is the unit `hermes gateway install --system`
+# writes, and it is the process that RECEIVES Telegram / WhatsApp / Discord /
+# Email — so every config save on a provisioned Hermes box restarts it
+# (src/lib/hermes-telegram.ts ensureHermesGateway). It is NOT install-time-only.
+#
+# This grant is a strict REDUCTION in privilege, not an addition:
+#
+#   * The unit is root-owned in /etc/systemd/system and runs `User=clawbox`
+#     (Hermes writes it that way — `gateway install --system --run-as-user
+#     clawbox`). Restarting it therefore starts a process the clawbox user could
+#     have started itself. That is less than the `restart clawbox-gateway` grant
+#     above, whose unit has no User= at all.
+#   * It REPLACES `sudo -n /home/clawbox/.local/bin/hermes gateway restart
+#     --system`. That binary, and every directory above it, is clawbox-owned and
+#     clawbox-writable — exactly the shape the optimize-ollama.sh grant was moved
+#     to /usr/local/libexec to avoid. It could never be granted; the coverage
+#     checker had to carry it in EXEMPT_CALLS instead, which meant the restart
+#     silently failed on a narrowed box and the route still answered "restarted".
+#
+# No `start`: the restart grant starts a stopped unit too. No `reset-failed`:
+# the generated unit sets StartLimitIntervalSec=0, so there is no start limit to
+# clear. No wildcard: profile units (hermes-gateway-<profile>.service) are a
+# different privilege question and ClawBox only ever runs the default profile.
+# The first-time `gateway install --system` stays deliberately UNGRANTED — it
+# writes into /etc/systemd/system and cannot be expressed safely against a
+# clawbox-writable binary; it fails fast under `sudo -n` and the route now says
+# so instead of reporting success. TASK-445 follow-up.
+clawbox ALL=(root) NOPASSWD: /usr/bin/systemctl restart hermes-gateway.service
+clawbox ALL=(root) NOPASSWD: /usr/bin/systemctl restart hermes-gateway
+
 # Ollama, for Settings -> Local Models. Ollama is a system unit and it holds
 # real RAM on an 8 GB box, so the owner has to be able to stop it and have it
 # stay stopped; until this grant existed nothing in the UI could. `--now` is
--- a/install.sh
+++ b/install.sh
@@ -2686,14 +2686,30 @@ install_sudoers_dropin() {
     return 1
   fi
 
-  install -d -o root -g root -m 0755 "$SUDOERS_DIR"
-  install -d -o root -g root -m 0700 "$SUDOERS_STAGING_DIR"
+  install -d -o root -g root -m 0755 "$SUDOERS_DIR" || return 1
+  install -d -o root -g root -m 0700 "$SUDOERS_STAGING_DIR" || return 1
 
   local staged
   staged="$(mktemp "$SUDOERS_STAGING_DIR/.sudoers-candidate.XXXXXX")" || return 1
-  cat "$src" > "$staged"
-  chown root:root "$staged"
-  chmod 0440 "$staged"
+  # Checked, not assumed. Both call sites invoke this function in a CONDITION
+  # context (`if install_sudoers_dropin …`, `… || echo`), and bash disables
+  # `set -e` for the whole dynamic extent of a command being tested. So every
+  # step in here has to carry its own `|| return 1`: an unchecked failure does
+  # not abort the script, it falls through to the next line and reports success.
+  # A truncated-but-parseable candidate — a `cat` that hit ENOSPC halfway down
+  # the allow-list — validates under visudo and installs cleanly. TASK-445.
+  if ! cat "$src" > "$staged"; then
+    rm -f "$staged"
+    echo "Error: could not stage $src; keeping the existing $dest" >&2
+    return 1
+  fi
+  if ! cmp -s "$src" "$staged"; then
+    rm -f "$staged"
+    echo "Error: staged copy of $src is truncated; keeping the existing $dest" >&2
+    return 1
+  fi
+  chown root:root "$staged" || { rm -f "$staged"; return 1; }
+  chmod 0440 "$staged" || { rm -f "$staged"; return 1; }
 
   if ! visudo -cf "$staged" >/dev/null 2>&1; then
     rm -f "$staged"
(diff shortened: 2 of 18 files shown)


KrasimirKralev:
## ⚠️ One thing I need you to confirm rather than take on trust

CI caught the new grant on your tripwire in `install-foreign-edition-teardown.test.ts:369` — *"the gateway's mask is justified by a sudoers grant no Hermes unit has"*, whose comment says: *"If an equivalent grant ever appears for a Hermes unit, this test fails and the teardown needs revisiting."* It fired exactly as designed. `a273101` revisits it; **this is the one change in the PR that relaxes a guard you wrote on purpose, so please rule on it.**

My reasoning, so you can disagree with the argument and not just the diff:

The rule the test encoded was *"no Hermes unit may appear in sudoers at all"*, and that is stricter than the reason behind it. The two directions of the teardown defend different kinds of thing:

* **`clawbox-gateway` is foreign on hermes** because it is an unauthenticated agent surface on :18789. That is a **security boundary** — so `step_edition_gateway_state` removes the unit file *and* masks it, and the `clawbox-gateway` grants are dead there. The mask is what justifies keeping the grants.
* **`hermes-gateway` is foreign on openclaw** because two harnesses polling one Telegram token deadlock each other (`install.sh:497-507`). That is a **functional conflict**, and `step_edition_foreign_teardown` deliberately only stops and disables it — it says so in its own operator message: *"Stopped and disabled only — nothing was masked and no unit file was removed."*

And it **cannot** mask that one: `hermes-gateway.service` is written by the upstream Hermes installer, not by us (your own test right below asserts `config/hermes-gateway.service` must not exist). A persistent mask leaves a symlink to `/dev/null` at the path a later `hermes gateway install --system` writes to — the exact trap `step_edition_gateway_state`'s unmask branch exists to undo.

So masking is off the table, which means the tripwire's blanket rule was the only thing holding the line for that unit. I considered three ways out and rejected two:

1. **Route the restart through `clawbox-root-update@`** (which needs no new grant). Rejected: the dispatcher execs `install.sh`, which lives in the clawbox-writable project tree — `clawbox-root-step.sh:14-15` says so. That turns "restart a `User=clawbox` process" into "run install.sh as root". Strictly worse than the grant.
2. **Add the mask to the teardown.** Rejected for the /dev/null reason above, and because unmasking would have to land in the hermes install path you and Mike are actively building on.
3. **Ship no grant and let the restart fail honestly.** Viable — `applied: false` would degrade to *"Saved — will apply on next gateway restart"* rather than lying — but it leaves messaging config saves not applying on a narrowed box, which is a real regression for a real feature.

I took the grant, and kept the tripwire with teeth:

* `hermes-dashboard` (a web surface) stays **completely** ungranted — unchanged assertion.
* `hermes-gateway` gets **`restart` and nothing else**. The assertion is now an anchored exact match against both spellings, so `start`, `enable`, `stop` or `unmask` each fail it — a new capability trips the wire, a cheaper spelling of an existing one does not.
* The teardown must still stop+disable it, and `hermes-gateway.service` must still be in `FOREIGN_EDITION_UNITS`, so the grant can only ever restart a gateway that **belongs** on the device.

The thing that tips it for me: the grant *replaces* `sudo -n /home/clawbox/.local/bin/hermes gateway restart --system`. That binary is `clawbox:clawbox 0755` (verified on a device) — passwordless root on a file the clawbox user can rewrite is one-step local root, and it is why `check-sudoers-coverage.sh` had to carry it in `EXEMPT_CALLS` rather than grant it. Swapping it for a systemctl restart of a root-owned unit that runs `User=clawbox` removes a real escalation and adds none.

If you would rather keep the original blanket rule, say so and I will take option 3 — the code change (systemctl instead of the writable binary, plus the honest `applied` flag) still stands on its own without the grant, and the only cost is that messaging config saves report *"applies on next gateway restart"* until the gateway is restarted some other way.

CI is otherwise green: this was the only failure.

KrasimirKralev:
@coderabbitai review

PR #502, opened 2026-08-27
Root steps stop running code the clawbox user can rewrite, and the grants lose their wildcards (TASK-445 GAP 2/2b/3)

KrasimirKralev:
Closes **GAP 2, GAP 2b, GAP 3, MINOR 6 and MINOR 7** from Mike's independent audit of #471 (`445-strict-verify`), plus two more members of the GAP 2b class the audit did not reach. MINOR 5 was already fixed by #495 (`fc15707d`). **GAP 1 (the unfiltered polkit `manage-units` grant) is deliberately NOT in this PR** — it is TASK-539 and lands in its own PR right after this one, so the two can be reviewed and reverted independently.

Base is `fix/hermes-sudoers-445-r2` (#471's head), stacked on #495.

---

## GAP 2 — root no longer runs code the clawbox user can rewrite

The granted chain is

```
clawbox  --sudo-->     systemctl start clawbox-root-update@<step>.service
         --systemd-->  /usr/local/libexec/clawbox/clawbox-root-step.sh   (root:root)
         --exec-->     /home/clawbox/clawbox/install.sh --step <step>
```

Only the middle link was root-owned. Confirmed again on a box flashed **today** (2026-08-27, main @ `227c26f1`):

```
$ ls -ld /home/clawbox/clawbox /home/clawbox/clawbox/install.sh
drwxrwxr-x 17 clawbox clawbox   4096 /home/clawbox/clawbox
-rwxr-xr-x  1 clawbox clawbox 221077 /home/clawbox/clawbox/install.sh
```

and the steps install.sh dispatches go on to run **more of that same tree** as root: `scripts/start-ap.sh` (recover, restart_ap), `scripts/launch-browser.sh`, `scripts/setup-hermes-edition.sh`, `scripts/setup-tunnel.sh`, `scripts/install-voice.sh`. So the grant also meant *"clawbox may choose the program root runs"* — passwordless local root in two moves for anything with clawbox-level code execution: the web server, the in-UI terminal, the agent's shell.

The tree cannot be moved out of clawbox's reach — the updater has to replace it and the app has to build in it — so root now **refuses to run code it did not record**.

* **`config/clawbox-root-manifest.sh`** (new, root-owned, installed to `/usr/local/libexec/clawbox/`) writes and verifies a sha256 manifest of `install.sh`, `scripts/` and `config/` at `/etc/clawbox/root-exec.manifest` (0644 root:root, staged inside root-owned `/etc/clawbox`, never `/tmp`).
* **`config/clawbox-root-step.sh`** verifies it before the exec.
* **`install.sh`** re-records it wherever the covered tree legitimately changes: the bootstrap block (right after a successful `git reset --hard origin/<branch>`, before the re-exec) and `sync_repo_to_update_target` (which both `step_git_pull` and `step_bootstrap_updater` go through).

Three design decisions are load-bearing, and each of them is there because the strict version bricks devices:

1. **The gate covers the pinned steps, not the update family.** The update family is exactly the set that legitimately rewrites the covered files — and it is not always install.sh doing the rewriting: `src/lib/updater.ts:498` runs its own `fetch` / `reset --hard` / `clean -fd` as the clawbox user before starting the rebuild step, and `scripts/force-update.sh` does the same by hand. Verifying there fails those flows at their next step and leaves the device refusing *every* root step afterwards. Instead the update family re-records as its first action, which is also what heals a tree replaced from the outside. This is not a hole the allow-list leaves open: GAP 3 below removes every sudo grant for a self-updating instance.
2. **An added file is not tampering.** Root only ever executes files install.sh names explicitly, all of which are recorded, so an unrecorded file is not something root can be made to run. Failing on additions turns any stray file under `scripts/` into a permanent refusal. `scripts/__pycache__` alone would have done it: `gateway-pre-start.sh` (the gateway's `ExecStartPre`) imports `scripts/gateway_origins.py`, so CPython writes a `.pyc` there the first time the gateway starts — after the manifest was written, and again under a new name after any python3 minor-version bump. `__pycache__`/`node_modules`/`.venv` are pruned from the record for the same reason.
3. **Manifest first, dispatcher second.** `install_root_libexec` writes the record and installs the dispatcher only if that succeeded — a dispatcher newer than its manifest refuses every root step. On failure the previous dispatcher stays where it is and the run records a provision failure, the same rule `install_sudoers_dropin` follows for the allow-list.

`scripts/root-update-step.sh` — the shim for field devices whose deployed unit still names it — now delegates to the root-owned dispatcher instead of exec'ing install.sh directly, so a device still on the old unit is inside the fix rather than outside it. It falls through to the legacy path only when `/usr/local/libexec/clawbox` does not exist at all (a genuinely pre-migration device, which needs that one run to install the dispatcher); when the directory exists but the dispatcher does not, it refuses.

## GAP 2b — and two more of the same class the audit did not reach

`step_chpasswd` read `$PROJECT_DIR/data/.chpasswd-input` — clawbox-writable — and piped it straight into `chpasswd`, with every guard living on the *unprivileged* side in `src/lib/chpasswd.ts`. Dropping `root:<new>` in there and starting the granted unit set **root's** password. The root side now requires exactly one record, naming exactly `$CLAWBOX_USER`, with a non-empty password and no CR; refuses a symlinked input; and scrubs the file whether it accepted it or not.

Reviewing that fix turned up two live siblings, both **worse**, because they `source` rather than parse — i.e. arbitrary root *code* execution, not just a wrong password:

* **`install.sh::read_configured_hostname`** did `. "$PROJECT_DIR/data/hostname.env"` as root. That file is written by the web server, and `clawbox-root-update@set_hostname.service` is one of the four instances the allow-list grants. `printf 'id > /tmp/pwn\n' > data/hostname.env` + the granted unit was root, in one move.
* **`install.sh`** `source`d `$PROJECT_DIR/data/network.env` at the **top level**, i.e. on every root run of the script, `--step chpasswd` included. A root-owned copy at `/etc/clawbox/network.env` already existed and is what the unit loads.
* **`scripts/start-ap.sh`** `source`d `data/hotspot.env`. `clawbox-ap.service` and `clawbox-ap-watchdog.service` carry no `User=`, so it runs as root — and `clawbox-root-update@restart_ap.service` is granted.

All three now parse a `KEY=VALUE` line instead of evaluating the file, and reject a symlinked input. The rule is one line long: **root parses files under `data/`, never evaluates them.**

TASK-445's "no git/network work in password changes" still holds and is now asserted: `chpasswd` is not in `SELF_UPDATING_STEPS`, so the dispatcher pins it with `CLAWBOX_INSTALL_BOOTSTRAPPED=1`.

**Recorded residual:** `clawbox` is in the `sudo` group, so setting the *clawbox* account's own password is still a route from clawbox code execution to an interactive root shell. That is the owner's administrator account; removing it would lock the only administrator out of a console-less box. What this closes is the part that was never intended — changing a *different* account's password, root's included.

## GAP 3 — no wildcards, because a prefix was not a scope

`reset-failed clawbox-*`, `start --no-block clawbox-*` and `start clawbox-root-update@*.service` looked scoped and were not. sudoers(5) matches a command's arguments as **one concatenated string**, so `*` spans whitespace — and `systemctl start` takes a **list** of units.

Reproduced against **sudo 1.9.9**, the Ubuntu 22.04 vintage the appliance ships, with `sudo -U clawbox -l <cmd>` (tests matching, executes nothing). Same host, same probes, old rules vs this PR:

| probe | old rules | this PR |
|---|---|---|
| `start clawbox-root-update@chpasswd.service ssh.service` | ALLOWED | **denied** |
| `start --no-block clawbox-root-update@llamacpp_install.service ssh.service` | ALLOWED | **denied** |
| `reset-failed clawbox-root-update@chpasswd.service ssh.service` | ALLOWED | **denied** |
| `start --no-block clawbox-setup.service ssh.service` | ALLOWED | **denied** |
| `reset-failed clawbox-a ssh.service` | ALLOWED | **denied** |
| `start clawbox-root-update@git_pull.service` | ALLOWED | **denied** |
| `start clawbox-root-update@build.service` | ALLOWED | **denied** |
| `start clawbox-root-update@bootstrap_updater.service` | ALLOWED | **denied** |
| `start --no-block clawbox-root-update@post_update.service` | ALLOWED | **denied** |
| all 11 commands the product actually issues | ALLOWED | **ALLOWED** |

Regex Cmnd arguments would say this directly, but they arrived in sudo 1.9.10 — after the version on the device — so the rules are **enumerated**: one exact `(verb, unit)` pair per command the product issues. Only four `clawbox-root-update@` instances are reachable from the web server (`chpasswd`, `set_hostname`, `restart_ap`, `llamacpp_install`), which is also what makes the update family sudo-unreachable and lets GAP 2's gate exclude it safely.

## MINOR 6 / MINOR 7

* `setup/reset/route.ts` started ollama with a bare `systemctl start ollama` and no sudo — it worked only through GAP 1's polkit grant, and `ollama` would not have matched the `ollama.service` rule anyway. It now calls the shared `startOllamaService()` from `src/lib/local-ai-runtime.ts`, which is where the argv is already pinned to the Cmnd_Spec, passes `-n`, keeps the unprivileged dev fallback, and waits for the API — replacing a hand-rolled retry loop.
* `sudoers_grants_blanket_nopasswd` records its accepted residual in a comment: a blanket line in `/etc/sudoers` itself, a `User_Alias`, and an over-broad-but-not-`ALL` rule are knowingly out of scope, with the behavioural `sudo -n` probes in `e2e-install` as the compensating control.

---

## Mechanical enforcement — a regression fails CI

`scripts/check-sudoers-coverage.sh` gains two **shape** invariants on top of its existing coverage checks, because coverage never made a grant safe:

1. **No wildcards.** Any `*` or `?` anywhere in a Cmnd_Spec is a build failure. Grant matching is exact comparison now that wildcards are rejected at parse time.
2. **Root-owned targets only.** Every granted command must resolve under `/bin`, `/sbin`, `/usr/bin`, `/usr/sbin` or `/usr/local/libexec/clawbox`. A grant on `/home/clawbox/clawbox/install.sh` or `/home/clawbox/.local/bin/hermes` fails by construction — GAP 2's shape written as a build rule. A relative command is rejected too: `secure_path` is a convenience, not a privilege boundary.

Verified firing (each appended to a fixture allow-list, checker run, exit 1):

```
/usr/bin/systemctl start --no-block clawbox-*        -> uses a wildcard
/usr/local/libexec/clawbox/*.sh                      -> uses a wildcard
/home/clawbox/clawbox/install.sh --step build        -> outside every root-owned prefix
/home/clawbox/.local/bin/hermes                      -> outside every root-owned prefix
systemctl reboot                                     -> relative command
```

New/changed tests:

* `src/tests/unit/root-exec-manifest.test.ts` (new) — drives the real shipped `clawbox-root-manifest.sh` and `clawbox-root-step.sh` with their constants retargeted at a temp tree: records + verifies; refuses a rewritten `install.sh`, a rewritten `scripts/start-ap.sh`, a removed file, a missing manifest; **does not** refuse an added file or a `__pycache__` churn; refuses a filename sha256sum would have to escape while accepting one containing `*`; refuses **and never execs**; fails closed with no verifier; pins `chpasswd` while letting the update family through. Also covers `install_root_libexec`'s manifest-before-dispatcher ordering and the keep-the-old-dispatcher fallback.
* `src/tests/unit/install-chpasswd-validation.test.ts` (new) — runs the real `step_chpasswd` with `/usr/sbin/chpasswd` redirected to a recorder (accepts `clawbox:`, keeps colons, refuses `root:`, another account, a smuggled second record, a CR, an empty password, a symlinked input — asserting each time that **nothing** reached `chpasswd`), and a second block proving root no longer *evaluates* `hostname.env` or `hotspot.env`.
* `src/tests/unit/sudoers-coverage.test.ts` — eight new cases for the two shape invariants.
* `src/tests/unit/root-steps.test.ts` — the wildcard tripwire is inverted (no grant may contain `*`/`?`, the four reachable instances must be named, no self-updating step may be), plus a case asserting the verification happens **before** the exec.
* `e2e-install/06-sudoers.spec.ts` — on a device install.sh really provisioned. The probes are **behavioural** (`sudo -n <cmd>`, classified on sudo's own "a password is required"): `sudo -n -l` cannot be used here, because install.sh puts clawbox in the `sudo` group and `%sudo ALL=(ALL:ALL) ALL` makes `sudo -l` answer "yes, with a password" for everything on the box. Appended-unit probes use a deliberately non-existent unit name so a regression fails the test without starting anything real.

## Provisioning: the blanket rule ships from the factory

Worth stating plainly, because it changes what "fixed" means. Nothing in this repo writes `clawbox ALL=(ALL) NOPASSWD: ALL`. It is created by the **flash pipeline**, twice — once into the rootfs image before flashing and once over SSH during first-boot setup — so every device is *born* with it and gets it re-created on every provisioning run.

Measured on the box flashed today:

```
$ ls -l /etc/sudoers.d/
-r--r----- 1 root root 8587 Aug 27 09:41 clawbox            # the allow-list, written by install.sh
-r--r----- 1 root root   32 Aug 27 08:33 clawbox-nopasswd   # the blanket rule, written by the flasher
$ sudo -n id -u >/dev/null 2>&1 && echo ESCALATED || echo DENIED
ESCALATED
```

08:33 is the flasher; 09:41 is install.sh. That ordering is why `quarantine_overbroad_sudoers` (#471) is the only thing that can clear it from devices already in the field, and why the image itself needs fixing so new devices are born clean — tracked separately.

The e2e container is therefore seeded with **both** blanket names seen on real devices, spelled `NOPASSWD:ALL` and `NOPASSWD: ALL`, and both must end up in quarantine — the detector parses the rule rather than matching text, and the fixture now looks like a real device.

## What is deliberately still open

**GAP 1 — the unfiltered polkit `manage-units` grant — remains open, and TASK-445 cannot be marked done while it is.** `config/49-clawbox-updates.pkla` grants `org.freedesktop.systemd1.manage-units` with no unit condition (`.pkla` on polkit 0.105 cannot express one), which is `systemd-run` — arbitrary root — for the clawbox user, and `step_polkit_rules` (`install.sh`) actively `rm -f`s the correctly-scoped `.rules` twin. Tracked as TASK-539; the PR follows this one.

Three notes from this work that feed straight into it:

* Routing the updater's polkit call sites through sudo will need **enumerated** grants for the `clawbox-root-update@<step>.service` instances they use, since wildcards are now a build failure. That is the intended cost: each new root entrypoint gets reviewed as a privilege boundary.
* Granting the update family through sudo also re-opens the question GAP 2's gate side-steps today. The right answer there is to move the git work itself to the root side (delete the `fetch`/`reset --hard`/`clean -fd` in `src/lib/updater.ts:498` and let the existing `git_pull` / `bootstrap_updater` root steps do it, since they already sync *and* re-record), not to widen the gate.
* `config/49-clawbox-updates.rules` — the scoped twin `step_polkit_rules` deletes — already documents the bounded set (`unit.indexOf("clawbox-root-update@") === 0`), which is exactly the list the sudoers migration needs.

Two root-executes-clawbox-writable paths found while doing this are **not** fixed here, because neither is reachable through a sudo grant and both are boot units that need hardware to re-test: `clawbox-ap.service` / `clawbox-ap-watchdog.service` (`ExecStart` inside `/home/clawbox/clawbox/scripts`, no `User=`) and `clawbox-firstboot-vnc.service`. Their *inputs* are fixed above; their *entrypoints* want the same libexec treatment as `clawbox-root-step.sh` in a follow-up.

## Summary by CodeRabbit

* **Security**
  * Added integrity checks for privileged update operations, detecting modified, deleted, or replaced files before execution.
  * Restricted passwordless administrative commands to explicitly approved services and arguments.
  * Improved protection against unsafe configuration files, symlinks, and untrusted input.

* **Reliability**
  * Gateway restarts now report success only after changes are applied.
  * Improved rollback handling for failed administrative configuration updates.
  * Simplified local AI service startup and unavailable-service handling.

* **Bug Fixes**
  * Corrected email polling status when gateway restarts fail.
  * Improved hotspot configuration parsing and restart-status warnings.

### DIFF
--- a/config/clawbox-root-manifest.sh
+++ b/config/clawbox-root-manifest.sh
@@ -160,11 +160,45 @@ verify_manifest() {
     || die "$PROJECT_DIR does not match $MANIFEST_FILE (a covered file changed or is gone)"
 }
 
+# Check ONE already-opened copy against what the manifest recorded for a path.
+#
+# `--verify` answers a question about the project tree, and the answer is stale
+# the moment it returns: the clawbox user can replace a file between the check
+# and the exec, and a tight rewrite loop wins that race. So the root dispatcher
+# copies the file it is going to run into a root-only directory FIRST and then
+# asks about the copy — which is the same bytes it will execute, and which
+# clawbox cannot touch.
+#
+#   clawbox-root-manifest.sh --verify-file <recorded path> <file to hash>
+verify_file() {
+  local rel="$1" actual="$2" want="" got h p
+  [ -n "$rel" ] && [ -n "$actual" ] || die "usage: $0 --verify-file <recorded path> <file>" 64
+  [ -f "$MANIFEST_FILE" ] || die "no manifest at $MANIFEST_FILE"
+  [ -f "$actual" ] || die "$actual is missing" 66
+
+  # Read the recorded hash out of the sha256sum-format manifest by exact path
+  # match. write_manifest refuses names it would have to escape, so the path
+  # column is the plain name (with a leading `*` in binary mode).
+  while read -r h p; do
+    p="${p#\*}"
+    if [ "$p" = "$rel" ]; then
+      want="$h"
+      break
+    fi
+  done < "$MANIFEST_FILE"
+  [ -n "$want" ] || die "$rel is not in $MANIFEST_FILE"
+
+  got="$(sha256sum < "$actual")"
+  got="${got%% *}"
+  [ "$want" = "$got" ] || die "$actual does not match what $MANIFEST_FILE recorded for $rel"
+}
+
 case "${1:-}" in
   --write)  write_manifest ;;
   --verify) verify_manifest ;;
+  --verify-file) verify_file "${2:-}" "${3:-}" ;;
   *)
-    echo "usage: $0 --write|--verify" >&2
+    echo "usage: $0 --write|--verify|--verify-file <recorded path> <file>" >&2
     exit 64
     ;;
 esac
--- a/config/clawbox-root-manifest.sh
+++ b/config/clawbox-root-manifest.sh
@@ -1,0 +1,151 @@
+#!/usr/bin/env bash
+#
+# Integrity manifest for the code root executes on the clawbox user's behalf.
+#
+# The privilege chain is:
+#
+#   clawbox  --sudo-->  systemctl start clawbox-root-update@<step>.service
+#            --systemd-->  /usr/local/libexec/clawbox/clawbox-root-step.sh (root:root)
+#            --exec-->     /home/clawbox/clawbox/install.sh --step <step>
+#
+# Only the middle link is root-owned. install.sh is `clawbox:clawbox 0755` in a
+# `clawbox:clawbox 0775` directory — install.sh itself hands the tree back with
+# `chown -R clawbox:clawbox` on every root run — and the steps it dispatches go
+# on to run more of the same tree as root (scripts/start-ap.sh,
+# scripts/launch-browser.sh, scripts/setup-hermes-edition.sh, …). So anything
+# with clawbox-level code execution — the web server, the in-UI terminal, the
+# agent's shell — could rewrite the program root was about to run and then
+# trigger a granted step. That is passwordless local root in two moves, and it
+# is the defect TASK-445 was filed about.
+#
+# Moving the tree out of clawbox's reach is not an option: the updater has to be
+# able to replace it, and the app has to be able to build in it. So instead the
+# root side REFUSES to run code it did not record. This file writes and checks
+# that record:
+#
+#   * install.sh writes the manifest at the end of every root-side install and
+#     immediately after every successful `git reset --hard` to the update branch
+#     (install.sh's bootstrap block and sync_repo_to_update_target). Those are
+#     the only two ways the covered files are supposed to change.
+#   * clawbox-root-step.sh verifies it before exec'ing anything. A tampered or
+#     unrecorded tree fails the step instead of running as root.
+#
+# What this does and does not buy:
+#
+#   * It closes the "rewrite install.sh, then start a granted unit" path — the
+#     dispatcher refuses before the exec.
+#   * It does NOT make the box safe against someone who can already run code as
+#     root, and it does not authenticate the UPDATE itself: an update legitimately
+#     replaces the covered files and re-records them. The update path is gated
+#     on the dashboard session instead (TASK-445's "require auth for update").
+#
+# Usage (root only):
+#   clawbox-root-manifest.sh --write     record the tree as it is now
+#   clawbox-root-manifest.sh --verify    exit 0 if it still matches, 65 if not
+#
+# Installed by install.sh::install_root_libexec to
+# /usr/local/libexec/clawbox/clawbox-root-manifest.sh, root:root 0755.
+
+set -euo pipefail
+
+# Hard-coded on purpose. Every value below selects WHICH code root executes, so
+# none of them is overridable from the environment: this script runs from a
+# systemd unit reached through a NOPASSWD sudoers grant, and an env escape hatch
+# would be a second way to point root at a file the clawbox user chose.
+PROJECT_DIR="/home/clawbox/clawbox"
+MANIFEST_DIR="/etc/clawbox"
+MANIFEST_FILE="/etc/clawbox/root-exec.manifest"
+
+# Everything the clawbox-root-update@ chain can end up running as root:
+# install.sh, the scripts it hands to bash, and the config/unit files it installs.
+# Runtime state — data/, .next/, node_modules/, .git/ — is deliberately NOT
+# covered: it is clawbox's to write and root never executes it, so covering it
+# would turn every build into a manifest mismatch.
+COVERED_PATHS="install.sh scripts config"
+
+die() {
+  echo "clawbox-root-manifest: $1" >&2
+  exit "${2:-65}"
+}
+
+# Covered files, relative to PROJECT_DIR, NUL-delimited and byte-sorted.
+# Callers must already be in PROJECT_DIR.
+#
+# `-type f` excludes symlinks deliberately: a symlink is a way to make a
+# recorded name resolve to unrecorded content, so one appearing under a covered
+# path shows up as a missing file and fails verification rather than passing it.
+covered_files() {
+  local p
+  local -a args=()
+  for p in $COVERED_PATHS; do
+    [ -e "$p" ] && args+=("$p")
+  done
+  [ "${#args[@]}" -gt 0 ] || return 1
+  find "${args[@]}" -type f -print0 | LC_ALL=C sort -z
+}
+
+# The path column of a sha256sum-format manifest. write_manifest refuses names
+# sha256sum would have to escape, so stripping the fixed-width prefix is exact.
+# The separator is two spaces in text mode and " *" in binary mode; coreutils on
+# the device emits the former, but accept both so the manifest stays readable
+# wherever it was generated.
+manifest_paths() {
+  sed -n 's/^[0-9a-f]\{64\} [ *]//p' "$MANIFEST_FILE"
+}
+
+write_manifest() {
+  cd "$PROJECT_DIR" || die "$PROJECT_DIR is missing" 66
+
+  local f bad=0
+  while IFS= read -r -d '' f; do
+    case "$f" in
+      *$'\n'*|*\*) bad=1; break ;;
+    esac
+  done < <(covered_files)
+  [ "$bad" -eq 0 ] || die "refusing to record a path containing a backslash or a newline"
+
+  install -d -o root -g root -m 0755 "$MANIFEST_DIR" || die "cannot create $MANIFEST_DIR" 66
+
+  # Staged inside the root-owned /etc/clawbox, never /tmp: a world-writable
+  # staging directory is one more place to race the file root ends up trusting.
+  local tmp
+  tmp="$(mktemp "$MANIFEST_FILE.XXXXXX")" || die "cannot stage a manifest" 66
+  if ! covered_files | xargs -0 -r sha256sum > "$tmp"; then
+    rm -f "$tmp"
+    die "cannot hash $PROJECT_DIR" 66
+  fi
+  if ! chmod 0644 "$tmp"; then
+    rm -f "$tmp"
+    die "cannot set the manifest mode" 66
+  fi
+  if ! mv -f "$tmp" "$MANIFEST_FILE"; then
+    rm -f "$tmp"
+    die "cannot install $MANIFEST_FILE" 66
+  fi
+}
+
+verify_manifest() {
+  [ -f "$MANIFEST_FILE" ] || die "no manifest at $MANIFEST_FILE"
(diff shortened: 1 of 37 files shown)


KrasimirKralev:
@coderabbitai review

KrasimirKralev:
## Review dispositions (CodeRabbit, ASSERTIVE profile)

All seven inline findings triaged; six fixed, one already covered.

| Finding | Severity | Disposition |
|---|---|---|
| `config/clawbox-root-step.sh` — **verify-to-exec race**: `--verify` returns, then bash opens `$ENTRYPOINT`; clawbox can swap install.sh in between | 🔴 Critical | **Fixed.** The dispatcher now `install`s install.sh into `/run/clawbox` (tmpfs, `0700 root:root`) at `0500 root:root`, hashes **that copy** through a new `clawbox-root-manifest.sh --verify-file`, and execs the copy. Same bytes checked and run, in a directory clawbox cannot reach. The residual — the scripts install.sh itself opens later — is now recorded in the code where the decision is, with the follow-up it points at. |
| `config/clawbox-sudoers` — `set_hostname` / `restart_ap` have no `reset-failed`, so repeated failures hit systemd's start limit and every later start is refused | 🟡 Minor | **Fixed.** The three callers (`system/hostname`, `setup/reset`, `system/hotspot`) now `reset-failed` first, as the chpasswd and llamacpp hand-offs already did, and the two grants are added. |
| `install.sh` — `SUDOERS_STAGING_DIR=/var/lib/clawbox` staged at `0700 root:root` on a **shared** directory | 🟡 Minor | **Fixed.** Staging moved to `/var/lib/clawbox/sudoers-staging`; the shared parent keeps its default mode. `clawbox-power-mode.sh`'s clock snapshot and the first-boot VNC marker live in that parent. |
| `scripts/check-sudoers-coverage.sh` — sudo compares the command path as a string and never canonicalises it, so `/usr/bin/../home/clawbox/…` passes the root-owned prefix test | 🟡 Minor | **Fixed.** Any `.` or `..` component in a granted path is a build failure. Verified firing. |
| `scripts/start-ap.sh` — a CRLF `hotspot.env` leaves `\r` on the SSID/PSK, which reaches `nmcli` as an argv byte | 🟡 Minor | **Fixed.** The trailing CR is stripped before quote handling, in both new parsers. |
| `src/app/setup-api/setup/reset/route.ts` — `getOllamaBaseUrl()` honours `OLLAMA_HOST`, so factory reset could `/api/delete` against a **remote** Ollama | 🟠 Major | **Fixed.** The delete loop is pinned back to `http://127.0.0.1:11434`. What it cleans is this device's own models under `/usr/share/ollama`; `startOllamaService()` (which is about the local unit) stays. |
| `src/tests/routes/discord|whatsapp/*` — add a running-but-not-applied gateway case | 🔵 Trivial | **Not this PR.** Those files and that condition come from #495, which this PR is stacked on; the finding is against its diff, not mine. |

### And one the review did not reach

Fixing the `source`-as-root class turned up a fourth member after the three in the commit above: **`scripts/ap-watchdog.sh:60`** did `. "$HOTSPOT_ENV"` in a subshell. `clawbox-ap-watchdog.service` carries no `User=`, so that ran as root **on a timer** — plant the payload as clawbox, wait twenty seconds. The subshell protected the watchdog's own variables and nothing else. It parses now, and a test asserts the payload does not run.

PR #508, opened 2026-08-27
Remove the unfiltered polkit manage-units grant and route root steps through one root-owned launcher (TASK-539)

KrasimirKralev:
Closes **GAP 1** from Mike's audit of #471 — the last open item on TASK-445, and the one he rated CRITICAL. Base is `beta`, on top of #471 (merged).

---

## What it was

`config/49-clawbox-updates.pkla`, installed by `install.sh::step_polkit_rules`:

```
[Allow clawbox to manage ClawBox systemd services]
Identity=unix-user:clawbox
Action=org.freedesktop.systemd1.manage-units
ResultAny=yes / ResultInactive=yes / ResultActive=yes
```

No unit condition — a `.pkla` on polkit 0.105, which is what JetPack ships, cannot express one. And `org.freedesktop.systemd1.manage-units` is the action systemd checks for `StartTransientUnit`, so it is `systemd-run /bin/sh -c …`: arbitrary root, no password, for the account the web server, the in-UI terminal and the agent's shell all run as. **Every narrowing in `config/clawbox-sudoers` was bypassable while it existed.**

Worse, `step_polkit_rules` actively `rm -f`'d `/etc/polkit-1/rules.d/49-clawbox-updates.rules` — the correctly scoped twin already in this repo (`unit.indexOf("clawbox-root-update@") === 0`) — which is how the unscoped `.pkla` became the only authority.

Reproduced read-only on a board flashed from `main` this morning (2026-08-27, OpenClaw edition):

```
org.freedesktop.systemd1.manage-units        AUTHORIZED
org.freedesktop.systemd1.manage-unit-files   NOT-AUTHORIZED   # control
org.freedesktop.systemd1.reload-daemon       NOT-AUTHORIZED   # control
org.freedesktop.hostname1.set-hostname       NOT-AUTHORIZED   # control
```

The three controls are what make that a statement about the `.pkla` and not about a permissive polkit.

## Why removal rather than narrowing

The upstream research settles it: **neither harness needs this.** Hermes' installer states three times that it "does not require or retain root access", its docs list "any user without sudo access" as supported, and its own FAQ prescribes exactly the shape this PR moves to — *"configure passwordless sudo for specific commands in /etc/sudoers"*. OpenClaw's daemon is a systemd **user** service. Of the grant categories in `config/clawbox-sudoers`, none is required by a harness: they are ClawBox product features, two ClawBox defects, and one provisioning artifact.

And `.pkla` cannot be narrowed — that is the whole reason this was stuck. sudoers can.

## What replaces it

One root-owned launcher, granted once:

```
clawbox ALL=(root) NOPASSWD: /usr/local/libexec/clawbox/clawbox-run-root-step.sh
```

Not ~25 enumerated `clawbox-root-update@<step>.service` instances: that is 50 lines of string matching nobody reviews, and the wildcard that would compress them is exactly what TASK-445 removed. `config/clawbox-run-root-step.sh` instead:

* takes a **STEP** name, never a unit name, and builds the unit itself;
* checks it against `WEB_ROOT_STEPS` — 25 names, deliberately narrower than the root dispatcher's 48, which also covers operator-only steps — with `src/tests/unit/root-steps.test.ts` pinning the shell list to `src/lib/root-steps.ts`;
* refuses a second argument, so "`systemctl start` takes a LIST of units" cannot come back through it;
* clears a previous failure itself, so no caller needs its own grant and no step becomes permanently unstartable against systemd's start limit (`clawbox-root-update@.service` does not set `StartLimitIntervalSec=0`).

No argument spec in the Cmnd, deliberately: the arguments are validated by root-owned code rather than by sudo's string matching — the same shape as the existing `optimize-ollama.sh` grant, and the same shape the deleted `.rules` file expressed.

**Verified against the real script on Ubuntu 22.04**, with `/usr/bin/systemctl` redirected to a recorder (nothing started):

```
=== permitted ===                    === refused (rc 64, systemctl never called) ===
chpasswd                        ok   git_pull            build           rebuild
--no-block llamacpp_install     ok   recover             polkit_rules    systemd_services
set_hostname                    ok   ../../etc/shadow    chpasswd.service
restart_ap                      ok   CHPASSWD            <empty>
post_update                     ok   chpasswd ssh        --no-block chpasswd ssh
                                     clawbox-root-update@chpasswd.service

what reached systemctl:  reset-failed clawbox-root-update@<step>.service
                         start [--no-block] clawbox-root-update@<step>.service
```

## Call sites

All seven move to `src/lib/root-step-runner.ts::startRootStep()`: `updater.ts` (both starters), `install/run-step`, `system/credentials`, `setup/reset` (chpasswd **and** set_hostname), `system/hostname`, `system/hotspot`, `llamacpp/install`. Every one passes `sudo -n`, so a device missing the grant fails in milliseconds rather than blocking a route handler on a prompt nobody can answer.

`step_polkit_rules` stops deleting `49-clawbox-updates.rules` and installs it instead — inert on polkit 0.105, correct on anything newer.

## The NetworkManager stanza stays

It names five specific NM actions rather than a category, `nmcli` needs them for the wifi and hotspot flows, and Mike's audit assessed it as legitimately scoped. Narrowing it further means routing every `nmcli` write through sudo, which is a separate change with its own hardware verification.

## Enforcement

`scripts/check-sudoers-coverage.sh` gains the call shape that made this possible to miss in the first place: `execFile("sudo", argv)` where `argv` is a **variable**. Without it, a helper that builds its own argument list is invisible to the whole check — its grant reads as unused, and a new one would read as covered.

New/changed tests:

* `src/tests/unit/root-step-launcher.test.ts` (new) — runs the shipped launcher with systemctl stubbed: permitted steps produce exactly the two intended commands; a step outside `WEB_ROOT_STEPS`, a non-identifier, a unit name, an empty step and a second argument all exit 64 with nothing reaching systemctl.
* `src/tests/unit/root-steps.test.ts` — the launcher's list must equal `WEB_ROOT_STEPS`, must be inside the dispatcher's allow-list and strictly narrower, must cover every UI step; and no grant may name a `clawbox-root-update@` unit any more.
* `e2e-install/06-sudoers.spec.ts` — on a real provisioned device: `pkcheck --action-id org.freedesktop.systemd1.manage-units` must be **NOT-AUTHORIZED** for clawbox (with `manage-unit-files` NOT-AUTHORIZED and `NetworkManager.wifi.scan` AUTHORIZED as controls, so a pass is about this grant and not about a broken probe), the installed `.pkla` must not mention `manage-units` and must still carry the NM stanza, and `sudo -n` must deny every `clawbox-root-update@` unit while allowing the launcher.

## Rebased onto #534

#534 ("verify the sudoers declarations instead of trusting them") landed under this branch and changed two things here:

* `%DECLARED_ARGV` entries are now `{ argv, resolve, unverified }` rather than a bare list, and every dynamic argument must be named in one of the latter two. The `root-step-runner.ts` declaration is re-expressed in that schema, with `argv` under `unverified` and the reason spelled out: `startRootStep()` assembles the list imperatively (`["-n", ROOT_STEP_LAUNCHER]`, then an optional `--no-block`, then the step), so there is no single const array for a symbol lookup to read back. The **step** stays deliberately un-enumerated — that is the whole design.
* The scanner's new `execFile("sudo", <variable>)` branch now goes through `declared_argv()` so the declaration is verified, not merely looked up. (The pre-rebase spelling read `%DECLARED_ARGV` directly and would have dereferenced a hash ref as an array.)

**#534 does not make that scanner branch redundant.** Removing it and the declaration together, on top of current beta, makes the checker report:

```
UNUSED grants (1):
  config/clawbox-sudoers:158  /usr/local/libexec/clawbox/clawbox-run-root-step.sh
```

— the only grant this PR adds becomes invisible to the check. Keeping the declaration but dropping the branch fails the other way, with #534's new stale-declaration error. The two are complementary: #534 verifies that a declaration still matches the code; this branch is what makes the call shape visible in the first place.

`install-sudoers-migration.test.ts` also gained a "grants exactly the systemctl commands the product issues" list in #534. Its eight `clawbox-root-update@` entries are gone here, because the four instances they named are no longer started by a systemctl grant at all.

## What this removes, and what it keeps

**Removed** — `org.freedesktop.systemd1.manage-units` for `unix-user:clawbox`, the entire first stanza of `config/49-clawbox-updates.pkla`. That is the action systemd checks for `StartTransientUnit`, so it was `systemd-run /bin/sh -c …`: arbitrary passwordless root for the account the web server, the in-UI terminal and the agent's shell all run as.

**Removed** — eight sudoers Cmnd_Specs, the enumerated `reset-failed` / `start` pairs for `clawbox-root-update@{chpasswd,set_hostname,restart_ap,llamacpp_install}.service`.

**Added** — exactly one: `/usr/local/libexec/clawbox/clawbox-run-root-step.sh`, root-owned, no argument spec, which validates the step itself.

**Kept, unchanged** — all five NetworkManager actions (`wifi.scan`, `network-control`, `settings.modify.system`, `wifi.share.open`, `wifi.share.protected`). `nmcli` needs them for the wifi and hotspot flows and they name specific actions rather than a category.

Also kept: every other grant in `config/clawbox-sudoers` (gateway/setup/browser/tunnel service control, ollama, apt/dpkg/snap, reboot/poweroff, the desktop and power mode helpers). This PR changes the root-step surface and nothing else.

The migration order is the part worth reviewing on a device: `step_rebuild_reboot` calls `step_systemd_services` — which installs the launcher (`install_root_libexec`) and then the new allow-list — **before** `step_polkit_rules` removes the grant. A box is therefore never left with neither.

## Still to do before this merges — this stays a DRAFT

**Hardware verification of the product flows, which has not been completed.** The e2e-install container exercises the allow-list and the polkit state, not a Jetson's radio.

Progress so far: a box was brought to beta head with the in-app updater and a full pre-change baseline captured (`manage-units` AUTHORIZED; `manage-unit-files`, `reload-daemon` NOT-AUTHORIZED; all five NM actions AUTHORIZED; polkit 0.105). It went offline before the branch could be installed on it, so **nothing on this PR has been exercised on hardware yet.**

A scripted one-pass validation is ready to run the moment a box is free — preflight and baseline, migration through the product's own updater, the privilege boundary, then wifi / hotspot / VNC / a full in-app update / the remaining call sites, then teardown.

Two things that verification has to settle, and cannot be settled from the repo:

1. **Does polkit 0.105 re-read the `.pkla` after `step_polkit_rules` overwrites it?** `cp` replaces the file, and nothing reloads polkit. If `manage-units` only becomes NOT-AUTHORIZED after a reboot, `step_polkit_rules` needs an explicit reload and that is a defect in this PR, not an environment quirk.
2. **Does every root step still start through the launcher on a real device**, including a full in-app update with the grant already gone — where a missing grant surfaces as `Command failed: /usr/bin/sudo` rather than anything about polkit.

Static analysis of the sibling call sites is complete: every `startRootStep()` caller is listed above, and a sweep of the tree for unprivileged system-scope `systemctl` writes finds none left — the remaining bare `systemctl` calls are reads (`show`, `is-active`, `is-enabled`), `--user`-scope calls, or run in already-root contexts (`clawbox-firstboot-vnc.service`, `clawbox-ap.service` and `clawbox-ap-watchdog.service` all carry no `User=`).

## Summary by CodeRabbit

* **Security**
  * Privileged setup and update operations now run through a validated, root-owned launcher.
  * Direct systemd management access is restricted, while NetworkManager scanning remains available.

* **Bug Fixes**
  * Improved validation prevents unauthorized or malformed root-step requests.
  * Root-step failures are reset before retrying, with optional non-blocking execution preserved.

* **Tests**
  * Added coverage for allowed steps, rejected inputs, sudo permissions, and launcher behavior.

### DIFF
--- a/config/49-clawbox-updates.pkla
+++ b/config/49-clawbox-updates.pkla
@@ -1,10 +1,3 @@
-[Allow clawbox to manage ClawBox systemd services]
-Identity=unix-user:clawbox
-Action=org.freedesktop.systemd1.manage-units
-ResultAny=yes
-ResultInactive=yes
-ResultActive=yes
-
 [Allow clawbox to manage NetworkManager]
 Identity=unix-user:clawbox
 Action=org.freedesktop.NetworkManager.wifi.scan;org.freedesktop.NetworkManager.network-control;org.freedesktop.NetworkManager.settings.modify.system;org.freedesktop.NetworkManager.wifi.share.open;org.freedesktop.NetworkManager.wifi.share.protected
--- a/config/clawbox-run-root-step.sh
+++ b/config/clawbox-run-root-step.sh
@@ -1,0 +1,91 @@
+#!/usr/bin/env bash
+#
+# The web server's one way to start a root update step.
+#
+# Until TASK-539 this went through polkit: src/lib/updater.ts and
+# /setup-api/install/run-step called `/usr/bin/systemctl start
+# clawbox-root-update@<step>.service` with no sudo at all, and
+# config/49-clawbox-updates.pkla authorised it. That grant is
+# `org.freedesktop.systemd1.manage-units` with no unit condition — `.pkla` on
+# polkit 0.105 (what JetPack ships) cannot express one — and manage-units is the
+# action systemd checks for StartTransientUnit. So it was `systemd-run /bin/sh
+# -c …`: arbitrary root, no password, for the account the web server, the in-UI
+# terminal and the agent's shell all run as. It made the whole sudoers allow-list
+# bypassable, which is why it had to go rather than be narrowed.
+#
+# sudoers CAN express a scope, so the operation moves here: one root-owned
+# entrypoint, granted once, that decides for itself which unit it will start.
+# The scoped polkit twin this replaces said the same thing —
+# config/49-clawbox-updates.rules restricts to `clawbox-root-update@` — but only
+# polkit >= 0.106 reads it.
+#
+# What this deliberately does NOT do: take a unit name. It takes a STEP name,
+# from a list narrower than the dispatcher's, and builds the unit itself. The
+# dispatcher (clawbox-root-step.sh) validates the step again on the other side
+# and is the authority on what root then runs; this is the outer bound on what
+# the WEB SERVER may ask for.
+#
+# Usage (root, from a sudoers grant):
+#   clawbox-run-root-step.sh [--no-block] <step>
+#
+# Installed by install.sh::install_root_libexec to
+# /usr/local/libexec/clawbox/clawbox-run-root-step.sh, root:root 0755.
+
+set -euo pipefail
+
+# Keep in step with WEB_ROOT_STEPS in src/lib/root-steps.ts —
+# src/tests/unit/root-steps.test.ts fails the build otherwise. Every name here
+# is a root entrypoint the web server can trigger without a password, so adding
+# one is a privilege decision, not a config line.
+WEB_ROOT_STEPS="
+ai_tools_install apt_update bootstrap_updater chpasswd chromium_install
+clawkeep_install cloudflared_install ffmpeg_install fix_git_perms gateway_setup
+hermes_edition llamacpp_install nvidia_jetpack ollama_install openclaw_config
+openclaw_install openclaw_patch openclaw_setup performance_mode post_update
+rebuild_reboot restart_ap set_hostname vnc_install vnc_refresh
+"
+
+NO_BLOCK=""
+if [ "${1:-}" = "--no-block" ]; then
+  NO_BLOCK="--no-block"
+  shift
+fi
+
+step="${1:-}"
+if [ "$#" -ne 1 ]; then
+  echo "usage: $0 [--no-block] <step>" >&2
+  exit 64
+fi
+
+# Reject anything that is not a plain step identifier before it reaches a unit
+# name. systemd instance names cannot contain `/`, but they can contain plenty
+# else, and this is the root side of the boundary.
+case "$step" in
+  *[!a-z0-9_]*|"")
+    echo "clawbox-run-root-step: refusing malformed step name: $step" >&2
+    exit 64
+    ;;
+esac
+
+permitted=1
+for allowed in $WEB_ROOT_STEPS; do
+  if [ "$allowed" = "$step" ]; then
+    permitted=0
+    break
+  fi
+done
+if [ "$permitted" -ne 0 ]; then
+  echo "clawbox-run-root-step: step not permitted from the web server: $step" >&2
+  exit 64
+fi
+
+service="clawbox-root-update@${step}.service"
+
+# Clear a previous failure first, so a step that failed once is not permanently
+# unstartable. Best-effort: a unit that never ran has nothing to reset.
+/usr/bin/systemctl reset-failed "$service" >/dev/null 2>&1 || true
+
+if [ -n "$NO_BLOCK" ]; then
+  exec /usr/bin/systemctl start --no-block "$service"
+fi
+exec /usr/bin/systemctl start "$service"
--- a/config/clawbox-sudoers
+++ b/config/clawbox-sudoers
@@ -146,14 +146,24 @@ clawbox ALL=(root) NOPASSWD: /usr/bin/systemctl stop    ollama.service
 # Adding an instance here adds a root entrypoint — the granted step runs as root
 # through /usr/local/libexec/clawbox/clawbox-root-step.sh — so review it as a
 # privilege boundary, not as a config line. TASK-445.
-clawbox ALL=(root) NOPASSWD: /usr/bin/systemctl reset-failed clawbox-root-update@chpasswd.service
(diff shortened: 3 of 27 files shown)


KrasimirKralev:
Not ready to merge: the code and CI evidence are in, the **hardware verification is not**.

Before this leaves draft, on a real Jetson (192.168.50.71 is free, freshly flashed OpenClaw, pinned to `beta`):

1. `pkcheck --action-id org.freedesktop.systemd1.manage-units --process $$` as clawbox must go from AUTHORIZED to **NOT-AUTHORIZED**. It is AUTHORIZED there right now — measured read-only this morning, with `manage-unit-files`, `reload-daemon` and `hostname1.set-hostname` all NOT-AUTHORIZED as controls.
2. An unprivileged `systemctl restart <unit>` must be denied.
3. Every product flow that used the grant must still work: UI power off/reboot, gateway restart, local-AI enable/disable, wifi change, hotspot on/off, and a full in-app update through the `clawbox-root-update@*` steps.

The e2e-install container covers the allow-list, the .pkla contents and the pkcheck answer, but it has no radio — (3) cannot be answered there. A feature silently breaking is worse than the hole this closes.

PR #514, opened 2026-08-27
Coding agent: re-advertise the coding_agent_* tools when the owner flips the switch

KrasimirKralev:
## The defect

Flipping the coding agent ON never reaches the running agent.

`coding_agent_run`, `coding_agent_status` and `coding_agent_stop` are registered **conditionally**: the ClawBox MCP server asks `/setup-api/coding-agent/status` once while it boots (`mcp/lib/context.ts` `probeCodingAgent`) and `mcp/tools/coding-agent.ts` opens with

```ts
if (!ctx.codingAgent) return;
```

so a `false` withholds the whole family. That gate is right — an always-409 tool opens the agent's per-server circuit breaker and takes *every* ClawBox tool offline with it, which is the same reason `email_list`/`email_read` are gated.

The problem is the probe is taken **once**, and the MCP server is a long-lived stdio child. `src/app/setup-api/coding-agent/enable/route.ts` wrote `setCodingAgentEnabled(true)`, logged it, and returned a freshly-read status to the browser — and told the running agent nothing. The panel says ready; the agent has no `coding_agent_*` at all until something unrelated respawns the child.

This is the **third** instance of one shape, and the only call site that was left out:

| # | Trigger | Refresh helper |
|---|---|---|
| #486 | mailbox becomes readable | `src/lib/email-mcp-refresh.ts` |
| #503 | the box becomes able to draw | `src/lib/hermes-image-refresh.ts` |
| **this** | the coding agent becomes usable | `src/lib/coding-agent-mcp-refresh.ts` |

`grep -rn reloadMcpServers src mcp` returned exactly those two callers before this change.

## The fix

`src/lib/coding-agent-mcp-refresh.ts`, modelled on both siblings: compare the verdict before and after the write, call `reloadMcpServers()` only on a **transition**, never throw, never turn the owner's save into an error.

The verdict compared is `CodingAgentStatus.ready` — `enabled` AND the harness installed AND ClawBox AI connected — because that is the same fact `probeCodingAgent` reads. Comparing the raw switch would fire on a box that can run nothing.

The guard is the load-bearing half. A reload kills and respawns every MCP child and invalidates the model's prompt cache. The same route also carries `effort`, `maxTurns`, `tokenLimit` and `defaultDirectory`; none of those change *which* tools exist, and neither does the switch on a box with no harness installed. All of those now cost nothing, and there are tests for each.

The route reads the status twice on the switch branch, deliberately: `ready` is three facts and only one of them is this request's to know. It is a handful of `stat()`s plus one `readdir`, on the one branch that flips a switch.

## Tests — RED first, against unmodified `origin/beta`

```
src/tests/routes/coding-agent/enable.test.ts    2 failed | 12 passed
  x reloads the MCP servers when the switch makes the family available
      AssertionError: expected "vi.fn()" to be called 1 times, but got 0 times
  x reloads them when the switch takes the family away
      AssertionError: expected "vi.fn()" to be called 1 times, but got 0 times
src/tests/unit/coding-agent-mcp-refresh.test.ts  suite failed to load (7 tests unrun)
      Error: Cannot find package '@/lib/coding-agent-mcp-refresh'
```

GREEN after the fix: **21 passed (2 files)**.

The route test mocks `reloadMcpServers` at its own seam rather than mocking the new helper, so it pins the behaviour the owner is owed — "the running agent is told" — and not the name of the module that happens to tell it.

## The recurring shapes this was checked against

- **A probe taken once at startup that never refreshes** — this *is* that bug; both directions of the flip are tested, and so is the no-op case that must not fire.
- **Reporting success from a call that returned, without checking what it returned** — `reloadMcpServers()` returning `false` must read as a refusal: `does not claim success when the dashboard refused` asserts nothing is logged as success.
- **An error path reporting failure over something that succeeded** — the write has already happened when the refresh runs, and an OpenClaw box has no dashboard at all. Two route tests assert the save still answers 200 when the reload is refused and when it throws.

## Deliberately not in scope

- **No honest-refusal `coding_agent_run`.** The image family registers one because an agent with no image tool *improvises* a picture the chat cannot display. There is no equivalent failure here — an agent with no `coding_agent_run` uses its own file tools, which is a reasonable answer, not a broken one — and `src/tests/unit/mcp-coding-agent-tools.test.ts` already pins "the family is absent when the device says the coding agent is off (circuit-breaker rule)". Adding a tool would contradict a decision this repo already made on purpose.
- **Linking ClawBox AI also moves `ready`** (via `clawaiConnected`), and that path already reloads for the image tools. Piggy-backing a second reload there would need coordination with `refreshHermesImageTools` to avoid two reloads on one save; worth its own change, not this one.

## Baselines

`bun run lint` and `tsc --noEmit`: unchanged from `origin/beta` (22 errors / 81 warnings, 24 TS errors — all pre-existing, none in any file this touches).

## Summary by CodeRabbit

* **Improvements**
  * Coding-agent tools now refresh automatically when readiness changes.
  * Unchanged readiness and unrelated setting updates no longer trigger unnecessary refreshes.
  * Settings continue to save successfully even if a tool refresh cannot be completed.
  * Updated status information is returned immediately after configuration changes.

* **Reliability**
  * Tool refresh failures are handled gracefully without interrupting coding-agent configuration.
  * Refresh requests are only reported as successful when accepted.

### DIFF
--- a/src/app/setup-api/coding-agent/enable/route.ts
+++ b/src/app/setup-api/coding-agent/enable/route.ts
@@ -1,4 +1,5 @@
 import { NextResponse } from "next/server";
+import { refreshCodingAgentToolsIfReadinessChanged } from "@/lib/coding-agent-mcp-refresh";
 import { hasOwnerSession } from "@/lib/owner-session";
 import {
   CodingAgentError,
@@ -87,6 +88,19 @@ export async function POST(request: Request) {
   }
 
   try {
+    // The family's availability BEFORE the write, read only on requests that
+    // can actually move it. `ready` — not the raw switch — because `ready` is
+    // the fact `probeCodingAgent` reads to decide whether the coding_agent_*
+    // tools exist at all, and it is the field this route already answers with.
+    //
+    // It is a SECOND status read on the switch branch, deliberately, rather than
+    // deriving the old verdict from the new one: `ready` is `enabled AND the
+    // harness is installed AND ClawBox AI is connected`, and only the first of
+    // those three is this request's to know. The reads are a handful of stat()s
+    // and one readdir, on the one branch that flips a switch — and null here
+    // means "this request cannot move the family", which is every other setting
+    // the route carries.
+    const readyBefore = hasEnabled ? (await getCodingAgentStatus()).ready : null;
     if (hasDirectory) {
       const saved = await setDefaultDirectory(fields.defaultDirectory as string | null);
       console.error(`[coding-agent] default folder ${saved ? "set" : "cleared"} by the owner`);
@@ -107,7 +121,22 @@ export async function POST(request: Request) {
       await setCodingAgentEnabled(fields.enabled as boolean);
       console.error(`[coding-agent] switched ${fields.enabled ? "on" : "off"} by the owner`);
     }
-    return NextResponse.json(await getCodingAgentStatus());
+    const status = await getCodingAgentStatus();
+    // Tell the RUNNING agent, not just the browser. The coding_agent_* tools are
+    // registered behind a probe the MCP server takes ONCE while it boots, and
+    // that server is a long-lived stdio child — so without this the panel says
+    // "ready" while the agent still has no way to start a run. Same shape and
+    // same mechanism as #486 (email) and #503 (images); see the helper for the
+    // rule about when a reload is worth its cost.
+    //
+    // AWAITED, and it cannot fail the save: the helper swallows everything and
+    // returns void, so the worst case is a logged line and a tool list that
+    // catches up at the next restart. A floating promise here would outlive the
+    // response with nothing watching it.
+    if (readyBefore !== null) {
+      await refreshCodingAgentToolsIfReadinessChanged(readyBefore, status.ready);
+    }
+    return NextResponse.json(status);
   } catch (err) {
     // The folder rules answer in the owner's words ("that folder holds
     // credentials…"); pass them through as a 400 rather than a 500, because
--- a/src/app/setup-api/coding-agent/enable/route.ts
+++ b/src/app/setup-api/coding-agent/enable/route.ts
@@ -15,14 +15,7 @@ import {
 export const dynamic = "force-dynamic";
 
 /**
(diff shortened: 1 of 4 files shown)


KrasimirKralev:
## Answering the three pre-merge checks

The review itself found **no actionable comments**, so this covers the three advisory checks.

### Docstring Coverage (66.67% < 80%) — fixed in 959d01a4

The uncovered function was `POST` in `src/app/setup-api/coding-agent/enable/route.ts`, and the check caught something real. The block describing every accepted body was attached to `forbidden()` — a helper that answers none of them — so `POST` had no doc at all and the refusal helper carried a contract it does not implement.

Split, rather than papered over: the owner-only rationale (why middleware is not trusted on this route) stays with the refusal it explains, the request contract moves onto the handler, and the switch branch's new duty — telling the running agent, not just the browser — is now named where a reader first meets it.

### Linked Issues check and Out of Scope Changes — misattribution, no change warranted

Both warnings are derived from linked issue `#39`, and `#39` is not an issue and is not linked to this PR:

- `gh pr view 514 --json closingIssuesReferences` → `[]`. This PR closes nothing and references no issue.
- `#39` is a **merged pull request** from long ago, *"Stabilize Local AI state and simplify provider setup"*.

So the reasoning is sound but the premise is not: this PR is correctly scoped, and there is nothing to bring into line with `#39` because nothing here was ever claimed to be about Local AI, provider setup, or llama.cpp. Both warnings resolve the moment the phantom link does; neither is asking for a code change, and making one to satisfy them would mean adding unrelated Local AI changes to a single-defect PR.

The PR numbers in the description — #486, #503 — are deliberate references to the two **sibling fixes of this same defect shape**, not issue links. This is the third and last call site of that shape.

### Scope, for the record

Four files, one behaviour: flipping the coding agent switch now reaches the running agent instead of only the browser. `bun run lint` and `tsc --noEmit` are unchanged from `origin/beta` (22 errors / 81 warnings, 24 TS errors, none in any file this touches), and the regression tests went 2-failed-plus-a-suite-that-could-not-load on unmodified `origin/beta` to 21 passed here.

PR #518, opened 2026-08-27
fix: a gh probe that timed out is not a missing gh

KrasimirKralev:
## The defect

`gh auth status` is a **network call** — it validates the stored token against api.github.com. On a box whose uplink is down or behind a captive portal it hangs, our timer SIGKILLs it at 60 s, and a signalled child closes with `code === null`.

That is the **same null** the `error` handler resolves with when the binary could not be started at all, because the close handler destructured only `(code)` and threw the `signal` away:

```ts
const timer = setTimeout(() => child.kill("SIGKILL"), opts.timeoutMs ?? GH_TIMEOUT_MS);
child.on("error", () => resolve({ code: null, ... }));   // never started
child.on("close", (code) => resolve({ code, ... }));     // killed -> also null
```

`githubStatus()` read that null as `installed: false`. So a box with **gh 2.4.0 at `/usr/bin/gh`** reported the GitHub CLI as absent, `backupToGitHub` returned `no_gh`, and `/setup-api/coding-agent/git` answered **409 "no_gh"** — telling the owner to install software they already have instead of that the network is down. `disconnectGitHub` carried the identical conflation and answered *"gh is not installed on this ClawBox."* for a logout that merely timed out.

## Reproduced on hardware (192.168.1.45)

| | result |
|---|---|
| `command -v gh` | `/usr/bin/gh` — `gh version 2.4.0+dfsg1` |
| `gh auth status`, uplink OK | reports the token *"is no longer valid"* — it can only know that by **asking github.com**, confirming it is a network call |
| `gh auth status`, uplink blackholed | **hangs, killed at the timeout (exit 124)** — exactly the SIGKILL → `code === null` path |

So the trigger is the ordinary condition of a device on a customer LAN, and the box misreports a present binary as missing.

## The fix

`run()` now keeps three facts apart instead of collapsing them into one null:

- the `close` **signal**,
- a **`timedOut`** flag set by our own timer before it kills,
- **`startFailed`**, set *only* by the `error` event — the sole evidence the binary is genuinely absent.

`githubStatus()` gains a third state: `installed: true, connected: false, reason: "unreachable"`. It is probed **fresh on every call and never cached** — a cached "unreachable" would outlive the outage that produced it and go on refusing backups after the uplink came back.

### The same null reached four more branches

Once the probe was past, four later calls read a non-zero/null code as a *fact about the folder*. The worst one is irreversible:

> `git remote get-url origin` exiting non-zero means **"no remote yet"**, and the branch below it **creates a repository on GitHub**. A killed probe returns the same null — so a transient network fault could create a **second repo** for a folder that already had one.

The regression test pins that no `gh repo create` is issued in that case. A killed `--show-toplevel` no longer tells the owner their repository *"is not its own git repository"*, and a killed push no longer reports a **blank** message (a SIGKILLed child writes no stderr, so the old `(stderr || stdout)` detail was `""`).

### Surface

- `/setup-api/coding-agent/git` answers a network fault **503** (transient, retry) rather than 409/500, and no longer advises an install.
- The card reads **"GitHub unreachable"** rather than "not connected" — we do not know whether an account is connected, only that github.com could not be asked. Translated across all 10 locales.

## Tests

New `src/tests/unit/coding-github-unreachable.test.ts`, written red-first against unmodified `beta`:

- **RED — 8 failed / 6 passed (14)**
- **GREEN — 14 passed (14)**

The failures included `expected false to be true` (installed on a killed probe), `'gh is not installed on this ClawBox.' not to match /not installed/i`, `expected true to be false` (a **second repository actually created**), and `expected '' not to be ''` (the blank push message).

Deliberately covers the three recurring shapes called out for this codebase: a probe cached instead of refreshed, an outcome inferred from an exit code, and an error path firing over something that succeeded (a logout that exited 0 is reported as the success it was).

`lint` and `tsc --noEmit` are byte-identical to the `beta` baseline (105 problems / 24 pre-existing TS errors, none in the touched files).

## Summary by CodeRabbit

* **Bug Fixes**
  * GitHub connectivity issues are now distinguished from disconnected accounts.
  * The coding agent displays a clear “GitHub unreachable” status when access cannot be verified.
  * Backup, disconnect, and repository operations provide more accurate errors and avoid unsafe actions during temporary connectivity failures.
  * Timeout and unavailable-service errors include clearer feedback.

* **Localization**
  * Added “GitHub unreachable” messages across supported languages.

### DIFF
--- a/src/app/setup-api/coding-agent/git/route.ts
+++ b/src/app/setup-api/coding-agent/git/route.ts
@@ -46,7 +46,15 @@ export async function DELETE(request: Request) {
     );
   }
   const out = await disconnectGitHub();
-  if (!out.ok) return NextResponse.json({ error: out.detail ?? "Could not disconnect" }, { status: 500 });
+  if (!out.ok) {
+    // A dead uplink is not a broken box: 503 says "transient, try again",
+    // which is what the owner should do, and the detail no longer claims gh
+    // is missing when the logout simply timed out.
+    return NextResponse.json(
+      { error: out.detail, kind: out.kind },
+      { status: out.kind === "gh_unreachable" ? 503 : 500 },
+    );
+  }
   console.error("[coding-agent] GitHub disconnected by the owner");
   return NextResponse.json(await githubStatus());
 }
@@ -81,7 +89,13 @@ export async function POST(request: Request) {
     });
     const outcome = await backupToGitHub(directory);
     if (!outcome.pushed) {
-      return NextResponse.json({ error: outcome.detail ?? outcome.reason, kind: outcome.reason }, { status: 409 });
+      // 409 means "the request cannot be satisfied as it stands" — true of a
+      // folder with no commits, false of a network that is merely down. That
+      // one gets 503, so nothing tells the owner to install a gh they have.
+      return NextResponse.json(
+        { error: outcome.detail ?? outcome.reason, kind: outcome.reason },
+        { status: outcome.reason === "gh_unreachable" ? 503 : 409 },
+      );
     }
     console.error(`[coding-agent] backed up ${directory} to ${outcome.repo}${outcome.created ? " (new repo)" : ""}`);
     return NextResponse.json(outcome);
--- a/src/components/CodingAgentApp.tsx
+++ b/src/components/CodingAgentApp.tsx
@@ -43,7 +43,7 @@ interface GitHubState {
   loginCommand: string;
   /** "unreachable" means gh is here but could not reach github.com — a
    *  network fault. The card must not read like a missing install. */
-  reason?: "not_installed" | "unreachable";
+  reason?: "not_installed" | "unreachable" | "not_runnable";
 }
 
 interface AgentStatus {
--- a/src/components/CodingAgentApp.tsx
+++ b/src/components/CodingAgentApp.tsx
@@ -41,6 +41,9 @@ interface GitHubState {
   connected: boolean;
   login: string | null;
   loginCommand: string;
+  /** "unreachable" means gh is here but could not reach github.com — a
+   *  network fault. The card must not read like a missing install. */
+  reason?: "not_installed" | "unreachable";
 }
 
 interface AgentStatus {
@@ -557,6 +560,12 @@ export default function CodingAgentApp() {
               {github.connected ? (
                 <span className="text-[11px] text-emerald-400 truncate" data-testid="coding-agent-github-login">
                   {github.login}
+                </span>
+              ) : github.reason === "unreachable" ? (
+                // Not "not connected": we do not know whether an account is
+                // connected, only that github.com could not be asked.
+                <span className="text-[11px] text-amber-400" data-testid="coding-agent-github-unreachable">
+                  {t("codingAgent.githubUnreachable")}
                 </span>
               ) : (
                 <span className="text-[11px] text-[var(--text-muted)]">{t("codingAgent.githubOff")}</span>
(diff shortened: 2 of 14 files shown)


KrasimirKralev:
Good catch — verified against the code and both are real, so both are fixed in e302d450.

**`rev-parse --verify HEAD`** — a non-zero code means "no commits yet"; a killed probe returns the same null. So a dead uplink told an owner with a full history that their folder was empty. Now guarded, and it returns the killed detail instead.

**`rev-parse --abbrev-ref HEAD`** — this one is worse than the report suggests, because the value is *used* rather than merely reported. The name feeds `push --set-upstream origin <branch>`, so the `"main"` fallback would push a `develop` checkout to `main` **and bind it there** — a durable wrong answer manufactured from a transient fault. Guarded too.

Both tests written red-first against the previous commit:

```
× does not call a repository EMPTY when the commit probe is killed
    AssertionError: expected 'nothing_to_push' not to be 'nothing_to_push'
× does not push to 'main' when the branch probe is killed
    AssertionError: expected true to be false      <- a push WAS issued to main
```

The second asserts on the recorded argv that no `push` was spawned at all, not just on the returned reason. Suite is now 16/16 (was 14), and lint/typecheck stay level with the `beta` baseline.

KrasimirKralev:
Also real, and the sharpest finding on this PR — thank you. Fixed in 671960f0.

You are right that `error` is not proof the binary is absent: Node emits it on an already-spawned child whose `kill()` cannot deliver its signal. My own guard therefore reintroduced the exact lie this PR exists to remove, and did it **on the timeout path** — gh runs, the timer fires, the kill fails, and the box says gh is not installed. The fix would have re-created the bug it was fixing.

`spawn` fires only when the process really started, so that is what now separates "never ran" from "ran, then something went wrong":

```ts
let spawned = false;
child.on("spawn", () => { spawned = true; });
child.on("error", () => resolve({ ..., startFailed: !spawned }));
```

A spawned child's real stderr is kept as well, instead of being overwritten with `"could not start"`.

Regression test `does not report a missing gh when the KILL fails on a running process`, red against the previous commit:

```
× does not report a missing gh when the KILL fails on a running process
    AssertionError: expected false to be true     <- installed: false, the lie back again
```

Suite 17/17. `missingBinary()` deliberately still never emits `spawn`, so the genuine not-installed case stays pinned and the two paths cannot collapse back into one.

KrasimirKralev:
Real, and the same shape a third time — fixed in 39cf5f7e.

`ENOENT` is the only errno that means *there is no such file*. `EACCES` means the opposite: gh is right there and would not execute. Both arrived as `startFailed`, so a binary somebody chmod'ed was reported as **not installed** — offering, again, the one remedy that cannot possibly help.

The spawn errno is now kept on the `Result` and read:

- `githubStatus()` returns a new `not_runnable` reason with `installed: true` — the file exists, which is *why* the errno was not ENOENT.
- `backupToGitHub` answers it as `failed` with the permissions, never `no_gh`.
- `disconnectGitHub` likewise stops saying "gh is not installed on this ClawBox."

I kept it off the `BackupFailure` enum and out of the card copy deliberately: the actionable detail already reaches the owner through the existing error path, and a fourth state plus ten locale strings would be more surface than a chmod'ed binary warrants. Happy to promote it if you disagree.

Three tests, red against the previous commit:

```
× does not say 'not installed' for a gh that is present but not executable
× a backup over an unrunnable gh never answers no_gh
× a disconnect over an unrunnable gh never claims it is not installed
```

Suite 20/20; lint and `tsc` still level with the `beta` baseline.

PR #522, opened 2026-08-27
fix: tell an edition-guard 404 apart from a bad id

KrasimirKralev:
Follow-up to #513. One residual, MCP-02a, plus the sibling the class check turned up.

## The residual

#513 decoded `skill_uninstall`'s two **labelled** refusals (404 `not_installed`, 409 `builtin_skill`) and deliberately let an **unlabelled** 404 fall through to the generic mapping, calling that "the honest outcome when we cannot tell which failure it was".

But the unlabelled 404 from these routes was never unknowable. It has one source: `hermesSkillsGuard()` (`src/lib/hermes-skills-server.ts:39`), which answers

```json
{ "error": "Not found" }
```

That is a JSON body with a non-empty `error` string, which is exactly the test `hasJsonErrorBody()` uses to call a 404 **resource-level**. So `fromApiError()` returned:

> The ClawBox could not find what this tool was pointed at.
> Check the id you passed. List what actually exists first (`ui_list_apps`, `code_project_list` or `skill_list`), then call this tool again with one of those ids.

The real condition is an **edition** mismatch. `skill_list` sits behind the same guard and 404s too, so every recovery that sentence names walks back into the guard. `mcp/lib/errors.ts` carries a comment at that exact branch naming this confusion — in the other direction — as the thing it must never do.

It was never only `skill_uninstall`: all six `/setup-api/hermes/skills/*` routes open with the same guard.

## Reproduced first

Against merged beta, before any source change — 12 of 15 new assertions red:

```
Tests  12 failed | 3 passed (15)
```

The 3 that passed on unmodified beta are the anti-regression ones: #513's `not_installed` decode, the store's genuine resource-404, and the guards returning `null` on their own harness. After the fix: **15 passed**.

## The fix

**Label the refusal.** Both guards now send `code: 'not_hermes'` / `code: 'not_openclaw'`. Status and the human-readable `error` string are unchanged, so the browser and `src/middleware.ts` see exactly what they saw before; only a machine caller can now tell the two 404s apart.

**Decode it everywhere, not at one call site.** `EDITION_RULE` is prepended by `skillsGet()` / `skillsPost()`, and all seven calls in `mcp/tools/skills.ts` go through them.

`skill_info` was the one that mattered most: its status-only 404 rule was already answering *"No skill with that id / call skill_search and use an id from its results"* for an off-Hermes device — a confident wrong answer, not merely a vague one. The edition rule is prepended so it precedes that rule; every other 404 rule in the module is `match`-gated on its own `code`, so none of them can shadow each other.

## How I know I found every sibling

```
grep -rn "hermesSkillsGuard" --include=*.ts --include=*.tsx . | grep -v node_modules
grep -rn "hermes/skills" mcp/ | grep -v node_modules
grep -n "apiGet\|apiPost\|rules:" mcp/tools/skills.ts
```

The first names all six guarded routes. The second names every MCP caller of them: seven in `mcp/tools/skills.ts`, one in `desktop.ts` (`ui_list_apps`, which `.catch(() => null)`s the read and reports an empty skill list — swallowed, never mis-decoded), one in `orientation.ts` (a raw `fetch` that reports `HTTP <status>` verbatim and interprets nothing). The third showed that only 3 of the 7 skills call sites passed any rules at all; `installedSkills()`, `skill_list` and the `inspect` docs phase passed none, so a per-call-site fix would have left three holes.

Rather than trust that reading, `mcp/tools/skills.ts` now reaches the API only through the two wrappers, and a test asserts `apiGet`/`apiPost` appear nowhere else in the module — the same defect cannot come back by someone adding an eighth call.

## The sibling: openclawAppsGuard

Widening from "the uninstall route" to "an edition guard's 404" put `openclawAppsGuard()` in scope, and it had the identical defect. `mcp/tools/desktop.ts`'s own header comment already recorded the consequence:

> On a Hermes device `/setup-api/apps/*` answers 404 "Not found", which a small model reads as "no such app" and retries forever — and a chronically-404ing tool is exactly what trips Hermes' per-server circuit breaker and takes EVERY ClawBox tool offline.

`app_search` and `app_install` now carry `STORE_EDITION_RULE`. Those are the only two MCP calls behind that guard: `apps/uninstall` and `apps/icon` are deliberately unguarded (the file says why), and `apps/settings` / `apps/skill-info` have no MCP caller. A wrapper would be the wrong shape there, so a test asserts the rule is present at both guarded call sites instead.

## Tests

`src/tests/unit/mcp-skills-edition-guard.test.ts`, 15 assertions:

- each guard labels its 404, and neither answers at all on its own harness
- all seven guarded tools — `skill_search`, `skill_list`, `skill_info`, `skill_install`, `skill_uninstall`, `app_search`, `app_install` — answer `NOT_SUPPORTED_HERE`, name the harness, say "do not retry", and never name a recovery that goes back through the guard
- the guard's **real** response body is taken off the wire and fed to the tools, rather than a hand-typed copy — the route lives under `src/`, the rule under `mcp/`, and nothing but a string literal joins them
- #513's `not_installed` decode still works (anti-shadowing), and a genuine unlabelled resource-404 from a store route is still `NOT_FOUND`
- the two anti-drift structural checks described above

## Severity

Low, and not customer-facing. The window is a device whose harness changed since the MCP child spawned; a dual-edition switch normally restarts the gateway and respawns it. What it costs when it opens is a small model looping on a recovery that cannot work — which is the failure mode the circuit-breaker comment was already worried about.

## Checks

- `eslint` on all five changed files: clean
- `tsc -p mcp/tsconfig.json`: 3 pre-existing errors in `src/lib/chat-history-cache.ts` and `src/lib/harness/transport.ts`, unchanged, none in touched files
- targeted suites (`mcp*`, `routes/hermes`, `routes/apps`): no new failures

## Summary by CodeRabbit

* **New Features**
  * Added clearer support handling for app-store and skill-store requests in unsupported app editions.
  * Unsupported requests now provide actionable, non-retry guidance instead of appearing as missing apps or skills.
  * Preserved existing handling for genuine “not found” results.

* **Bug Fixes**
  * Improved consistency across browsing, searching, installing, inspecting, and uninstalling store items.
  * Added clearer error details for unsupported editions.
  * Prevented unsupported store requests from being misclassified as missing items.

### DIFF
--- a/mcp/tools/desktop.ts
+++ b/mcp/tools/desktop.ts
@@ -4,6 +4,7 @@
 // answers 404 "Not found", which a small model reads as "no such app" and
 // retries forever — and a chronically-404ing tool is exactly what trips
 // Hermes' per-server circuit breaker and takes EVERY ClawBox tool offline.
+// STORE_EDITION_RULE below is what turns that 404 into "stop".
 
 import { apiGet, apiPost, CLAWBOX_ROOT } from "../lib/api";
 import { ToolError, type ErrorRule } from "../lib/errors";
@@ -53,6 +54,21 @@ const CODE_RULES: ErrorRule[] = [
     next: "Call code_project_list for the ids that exist here, or create one with code_project_init.",
   },
 ];
+
+// openclawAppsGuard() answers 404 `{"error":"Not found","code":"not_openclaw"}`
+// from every store route. Matched on the CODE, not the status: the store routes
+// also 404 for an app id that does not exist, and the two need opposite advice.
+// The tools are registered off an edition probe taken once when the MCP child
+// spawned, so the window is a device whose harness changed since then.
+const STORE_EDITION_RULE: ErrorRule = {
+  status: 404,
+  match: /"code"\s*:\s*"not_openclaw"/,
+  code: "NOT_SUPPORTED_HERE",
+  message: "This ClawBox is not running the OpenClaw harness, so it has no app store.",
+  next:
+    "Do not retry and do not call the app store tools again this session. "
+    + "Call device_status and tell the user which harness the device is on.",
+};
 
 const WEBAPP_RULES: ErrorRule[] = [
   {
@@ -164,6 +180,7 @@ export function registerDesktopTools(reg: Registrar, ctx: McpContext): void {
       const body = await apiGet("/setup-api/apps/store", {
         query: { ...(query ? { q: query } : {}), limit },
         timeoutMs: 20_000,
+        rules: [STORE_EDITION_RULE],
       });
       return json(body);
     },
@@ -175,7 +192,11 @@ export function registerDesktopTools(reg: Registrar, ctx: McpContext): void {
     { app_id: zSlug("App id from app_search") },
     { editions: ["openclaw"], readOnly: false },
     async ({ app_id }: { app_id: string }) => {
-      await apiPost("/setup-api/apps/install", { appId: app_id }, { timeoutMs: 120_000 });
+      await apiPost(
+        "/setup-api/apps/install",
+        { appId: app_id },
+        { timeoutMs: 120_000, rules: [STORE_EDITION_RULE] },
+      );
       return text(`Installed "${app_id}". Open it with ui_open_app using "installed-${app_id}".`);
     },
   );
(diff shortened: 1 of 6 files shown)


KrasimirKralev:
@coderabbitai full review

PR #526, opened 2026-08-27
Guard the wizard save against the Claude subscription surface too

KrasimirKralev:
Follow-up to #520. One residual, closed.

## RES-520-A — `/setup-api/ai-models/configure` is the second write path to `agents.defaults.model.primary` and applied no subscription-surface guard

**Reproduced against merged beta** (`cd90857a`), by the two tests below failing on unmodified `origin/beta`.

#520 added `refuseOffSurfaceClaudeModel` at both guard sites inside `src/app/setup-api/chat/model/route.ts`, explicitly *"for ids that arrive some other way"*. The other way is a different route.

`configure/route.ts` validates a typed custom id for **shape only** — `isValidModelId(targetProvider, requestedModel)`, under the comment *"We don't check against the curated list"* — then writes `config.defaultModel = anthropic/<requestedModel>` and pushes it to `agents.defaults.model.primary`. Nothing in that file imported `@/lib/subscription-surface`.

The client-side rule does not close it either. `AIModelsStep`'s `isModelUsable` only greys out **catalogue** rows; the custom-id path is deliberately exempt (*"A typed custom id is always allowed through"*), and `allowCustom` is true for anthropic. `saveOAuthToken` sends `getRequestedCatalogModelId(true)` verbatim on the subscription path.

So on a Claude-subscription box a customer who flips to "custom model ID", types `claude-fable-5`, and signs in pins the box to a model the `claude-cli` surface cannot route — precisely the failure #520 removed from the chat header. Arch-independent, so it bites on shipped aarch64 hardware.

## The fix: one rule, not two copies

`refuseOffSurfaceClaudeModel` is lifted out of `chat/model/route.ts` into `@/lib/subscription-surface` as `offSurfaceClaudeModelMessage` (plus `isClaudeSubscriptionOnly`, the "is this box Claude-subscription-only" question both routes ask). Each route keeps a thin wrapper that turns the message into its own `NextResponse`. The chat route's behaviour and wording are byte-for-byte unchanged — its 17 existing tests pass untouched.

Two things about the new call site are deliberate:

**It judges the settled `config.defaultModel`, not the typed id.** The typed-id field is the door that was reported, but the same save writes the `PROVIDERS`-table default when nothing is typed. Checking after every branch has had its say is one check for every value this save can put in `primary`.

**It asks whether the box will be subscription-only *after* this save, not before.** Reading the profiles on disk alone gets both directions backwards. A first Claude sign-in is *what writes* the OAuth profile that makes the box subscription-only, so disk alone would wave the very first sign-in straight through; and an API-key save writes the key that makes the API-only models routable, so refusing there would block the very switch the message recommends ("switch Anthropic to API-key mode"). So the API-key half is the enclosing branch — only a subscription save is checked at all — and the predicate overlays the OAuth profile entry `baseOps` is about to write onto the ones already on disk before asking. A key held under some *other* profile key still counts, and still allows.

**Placement.** After the Hermes branch — a Hermes box has no `openclaw.json` to read, and that branch refuses a subscription save outright — and before `writeAuthProfiles`, so a refusal leaves no half-applied save behind. `UNKNOWN` (unreadable or empty surface cache) still means *allow*, the same rule the pickers and the chat route obey.

## How I know I found every sibling

The class is: a write to `agents.defaults.model.primary` that accepts a model id without asking the subscription surface about it. The grep:

```
grep -rn "agents\.defaults\.model\.primary" src/ scripts/ | grep -v "__tests__\|\.test\."
```

Four write sites, all accounted for:

| site | status |
|---|---|
| `chat/model/route.ts:634` | guarded by #520 (both doors) |
| `ai-models/configure/route.ts:1717` | **guarded by this PR** |
| `local-ai/exclusive/route.ts:274` | writes the local (`llamacpp/`/`ollama/`) model — never a Claude id |
| `local-ai/exclusive/route.ts:295` | restores the primary it saved at toggle-on. It accepts no new id, and refusing it would strand the box in Local-only with nothing to restore to — deliberately out of scope |
| `scripts/gateway-pre-start.sh:1254` | writes a `codex/` id, not anthropic |

And a second grep for anything else that reaches the same write indirectly:

```
grep -rn "refuseOffSurfaceClaudeModel\|subscription-surface\|readSubscriptionSurfaceIds\|subscriptionOnlyProviders" -r src/ --include=*.ts --include=*.tsx
```

That confirmed `chat/model/route.ts` was the only importer before this PR. `providers/default/route.ts` ("make this provider the default") writes nothing itself — it calls the `chat/model` POST handler as a plain function and so inherits the guard, which is exactly why it delegates.

## Tests — red first

New: `src/tests/routes/ai-models/configure-subscription-surface.test.ts`, 8 tests.

Against unmodified `origin/beta`: **2 failed, 6 passed** — the two that assert the refusal (`expected 200 to be 400`). The other six are the guard's negative space (API-key saves, both-credentials boxes, unreadable cache, empty cache, non-Claude providers) and had to pass before and after, or the guard would be refusing things that work.

After the fix: **8 passed**, and #520's own 17 in `chat-model-subscription-surface.test.ts` still pass. `src/tests/routes/ai-models/` + `chat-model*` + `providers/` together: 269 passed, 0 failed.

`eslint` on the four touched files: 0 errors, 2 warnings, both pre-existing (`_api`, `OPENAI_PRO_MODEL_RE`). `tsc --noEmit`: no new errors — the ones that remain are the same pre-existing test-file errors present on `origin/beta`.

### DIFF
--- a/src/app/setup-api/ai-models/configure/route.ts
+++ b/src/app/setup-api/ai-models/configure/route.ts
@@ -61,6 +61,10 @@ import { resolveEntitledCodexModel } from "@/lib/codex-model-probe";
 import { fetchPortalTier } from "@/lib/clawbox-ai-portal-tier";
 import { isValidModelId, isCatalogProvider, GOOGLE_MODELS, ANTHROPIC_MODELS, extractProviderModelId } from "@/lib/provider-models";
 import { refreshInBackground as refreshCatalogInBackground } from "@/app/setup-api/ai-models/catalog/route";
+import {
+  isClaudeSubscriptionOnly,
+  offSurfaceClaudeModelMessage,
+} from "@/lib/subscription-surface";
 // The model name on this route arrives in the request body. For a local
 // provider it is the whole of `apiKey`, which nothing further constrains, and
 // it reaches the lines below both directly and inside a subprocess error that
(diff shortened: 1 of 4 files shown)


KrasimirKralev:
The first push drew one new CodeQL alert (`js/user-controlled-bypass`, high): the refusal condition was controlled by `authMode` from the request body, via a `mode: authMode === "subscription" ? "oauth" : "api_key"` ternary inside the predicate.

Fixed properly rather than dismissed, in eaa72063. Only a subscription save can create the hazard at all — an API-key save writes the anthropic key that makes the API-only models routable, so after it lands there is nothing to refuse. So `authMode` becomes the enclosing branch and the projected profile is unconditionally `mode: "oauth"`. Same behaviour for every reachable case (the old ternary's `api_key` arm could never produce a refusal — a keyed provider is never subscription-only), one fewer conditional, and CodeQL now reports "No new alerts in code changed by this pull request".

CodeRabbit's docstring-coverage warning is addressed in the same commit: the three undocumented test helpers now carry one-line docstrings. 5/5 pre-merge checks pass.

PR #533, opened 2026-08-28
fix: close the #519 residuals — every caller reads the verdict, and the update path carries it

KrasimirKralev:
Follow-up to #519. Two independent adversarial verifiers came back with the same shape: **the core fix is real, but a sibling call site was left unguarded.** This closes all four named residuals plus the third sibling the class grep turned up.

## RES-519-A — `--piper-only` reported success from a return code (medium)

`scripts/install-voice.sh` line 747. `install_piper` returns **0** for "there is no pinned artifact for this board" — that is the whole reason #519 taught `--tts-only` to read `TTS_PIPER_VERDICT` instead. `--piper-only` still did `install_piper || PIPER_ONLY_RC=1` and never looked at the verdict, so on a non-aarch64 board `install_piper_engine` took its `return 0` skip path, published `PIPER=skipped:arch-<arch>`, and the mode printed `=== Piper fallback ready ===` and exited 0 with **no engine installed** — in the mode `scripts/openclaw/clawbox-tts.sh:596` tells operators to run when it reports "Piper not installed".

It now decides the way `--tts-only` does: `ready` announces the fallback, `skipped:*` prints "no Piper artifact applies to this board" and exits 0 (the board declines — nothing was asked for and nothing is missing, the same rule `skipped:*` carries everywhere else), and `failed:*` / unreported / anything unparseable exits 1. The `PIPER_ONLY_RC` check stays in front of it, so a `deploy_voice_scripts` failure behind a ready engine still fails as before.

## RES-519-B — `VOICE_RC=11` was told it has a Piper fallback (low)

`install.sh` line 2794 bucketed `10|11|12` and printed `On-device TTS configured (Piper CPU only — …)`. 11 is `install_kokoro_tts`'s "no Jetson CUDA build for `$arch`" — i.e. `uname -m != aarch64`, which is **exactly** the condition under which `install_piper_engine` skips too. Both halves publish `skipped:arch-*`, `--tts-only` correctly prints "=== No on-device TTS engine applies to this board ===" and exits 11, and install.sh then named a fallback that does not exist one line later.

11 gets its own branch. 10 and 12 are genuinely Piper-backed — a Piper failure with Kokoro ready exits 1, with Kokoro failed exits 13 — so they keep the "Piper CPU only" line, and a test pins that. 11 stays a **clean** provision (no `record_provision_failure`, no probe failure): a board neither engine ships for is the case `skipped:*` was defined for, and failing every `install-x64.sh` run would only teach everyone to ignore the check. The stale "10 / 11 … the box speaks on Piper" contract comment in install-voice.sh is corrected with it.

## RES-519-C — the probe scored garbage above silence (low)

`install.sh` line 4911. The chain fails on empty, on `failed:*` and on "neither ready and one failed", but a non-empty value in none of those shapes matched no arm and scored a **silent pass** — so `PIPER=redy` passed while the strictly less informative `PIPER=` correctly failed. Reachable for real: `tts_status_publish` truncates with `>` rather than write-then-rename, so a box that lost power mid-publish can leave a torn value.

**Deviation from the proposed fix, deliberately:** the report suggested closing the chain with a trailing `else`. That would fail every healthy box — `KOKORO=ready` + `PIPER=ready` matches none of the arms either; the fall-through is where the good box lands. So the guard is a final `elif` on an explicit "outside the vocabulary" flag, placed last so the more specific arms above still win (a garbled Piper verdict next to a failed Kokoro is still best reported as "no working engine"). A test pins the healthy box passing.

Both readers of the file (`install.sh:4874-4875`, `install-voice.sh:550-551`) now strip CR. Without it a CRLF `KOKORO=ready` + `PIPER=failed:*` reported **"this box has NO working TTS engine"** over a box whose GPU engine is running — a failure report over something that succeeded.

## RES-519-D — the update path threw the verdict away (medium, customer-facing)

`install.sh` line 3512. #519 rewrote the first caller (`step_openclaw_setup`) with a tolerance table that keeps 12, 13 and 14 apart on purpose. The second caller was still `step_openclaw_tts || echo "  Warning: openclaw_tts step failed (non-fatal)"` — one line indistinguishable from the fourteen `|| echo` lines around it, reporting "this box answers speech with SILENCE" in the same words as a skipped VNC refresh. It now carries the same table, with `*)` naming an out-of-contract code.

And the record: `--step` dispatch ran `"step_${local_step}"; exit 0`, so the `record_provision_failure` #519 added inside `step_openclaw_tts` never reached the `PROVISION_FAILURES` summary or `write_provision_status` (install.sh:5268-5306) — an in-app update that left a shipped aarch64 box mute finished as a **successful update**. The dispatch block now publishes the verdict.

Implementation notes worth reviewing:

- An **EXIT trap**, not `"step_x" || rc=$?`. The OR-list form switches `set -e` off for the whole body of the dispatched step, so every guard inside it that relies on errexit to stop would run on instead. The trap also catches a step that dies mid-way, which previously reported nothing at all.
- Only `incomplete` is ever written there. One dispatched step finishing cleanly is not evidence that the whole box provisioned, so a clean run publishes nothing and leaves the marker to the full install that owns it. A test pins exit 0 + no marker for a clean step, and a failing step keeping its own exit code.

## Fixing the class, and how I know I found every sibling

The class is *"a caller that reports from the return code rather than from the published verdict."* Two greps, both run against the merged head:

```
rg -n 'install_piper\b'      # -> defn 263, and callers 678 (--tts-only), 747 (--piper-only), 870 (full pipeline)
rg -n 'step_openclaw_tts'    # -> defn 2584, callers 1643 (step_openclaw_setup), 3512 (step_post_update), + the dispatch whitelist
```

`install_piper` has **three** callers, not two. The third — the full-pipeline path at line 870 — was the least honest of them: it printed `TTS: Kokoro-82M …, Piper CPU fallback` unconditionally, on the x86 runs where the engine declines and on the runs where its download failed. Its summary now names the engine from `TTS_PIPER_VERDICT` too. Its exit contract is deliberately unchanged; only the claim it makes is.

A third grep covers the verdict readers, so the CR fix is complete rather than half-applied:

```
rg -n 'TTS_STATUS_FILE|tts-status'   # -> exactly two parsers: install.sh:4874-4875, install-voice.sh:550-551
```

## Tests — red first

New file `src/tests/unit/install-tts-verdict-siblings.test.ts`, 21 tests, executing the shipped artifacts against stubs like the #519 suite next to it (the real `--piper-only` dispatch, the real `step_openclaw_tts`, the real `step_validate_services`, the real `step_post_update`, and the real `--step` dispatch block lifted out of install.sh).

- Against **unmodified `origin/beta`** (4709b4f2, sources stashed, test file in place): **10 failed / 11 passed**.
- With the fix: **21 passed**.
- `install-tts-no-engine.test.ts` (#519's own suite): 23/23 still green.
- `install-provision-status-marker`, `install-post-update-units`, `root-steps`, `observability-install`, `desktop-power-wiring`: 55 passed / 8 skipped, unchanged.

The 11 that pass red are the over-correction guards, and they are the point of half of this diff: a healthy box still passes the probe, a board that declines both engines still provisions clean, 10 and 12 still name Piper, a clean dispatched step still exits 0 and writes no marker.

`bash -n` clean on both scripts. `tsc`: 24 pre-existing errors before and after, none from the new file. `eslint` clean on the new file. (`clawbox-tts.test.ts` and `install-kokoro-tts.test.ts` fail identically with and without this change on the Windows dev host — spawn-path artifacts, 46/23 and 37/9 both sides; Linux CI is the gate.)

## Honest severity

A, B and C are all gated on `uname -m != aarch64`, and shipped ClawBox hardware **is** aarch64, so none of them can mute a Jetson today. D is reachable on shipped hardware: `KOKORO_STAMP_VERSION` was bumped to 2 to force every already-stamped box to redo the Kokoro install, so a flaky network during that redo gives `KOKORO=failed:torch`, and a box whose `$PIPER_DIR/piper` is missing then gives exit 13 rather than 12.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved voice-engine installation reporting for unsupported architectures and unavailable engines.
  - Correctly distinguishes ready, skipped, failed, missing, and invalid text-to-speech statuses.
  - Provides accurate Piper fallback summaries instead of incorrectly reporting availability.
  - Reports post-update speech failures individually without interrupting the remaining update process.
  - Installer failures now return appropriate error statuses and clearly mark incomplete provisioning.

- **Tests**
  - Added comprehensive coverage for voice-engine verdicts, installation outcomes, updates, and status reporting.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -2791,8 +2791,28 @@ step_openclaw_tts() {
     # there — which is where an unhandled 13 would have landed — would put the
     # name of a working fallback on the box this whole fix exists to catch.
     case "$VOICE_RC" in
-      10|11|12)
+      10|12)
         echo "  On-device TTS configured (Piper CPU only — $KOKORO_REASON)"
+        ;;
+      11)
+        # 11 was bucketed with 10 and 12 and is not like them. It is
+        # install_kokoro_tts's "no Jetson CUDA build for this architecture",
+        # i.e. `uname -m` is not aarch64 — which is exactly when
+        # install_piper_engine has no pinned artifact either and publishes
+        # `skipped:arch-*` of its own. install-voice.sh has just printed "=== No
+        # on-device TTS engine applies to this board ===" for that pair, so
+        # naming a Piper CPU fallback one line later put the name of an engine
+        # this box does not have on the summary line an operator reads.
+        #
+        # Deliberately NOT a recorded provisioning failure: nothing was asked
+        # for and nothing is missing on a board neither engine ships for, which
+        # is the same rule `skipped:*` carries in step_validate_services.
+        #
+        # On stdout, not stderr, for the same reason: install-voice.sh prints
+        # its "no engine applies to this board" line on stdout too, and painting
+        # every install-x64.sh run red would only teach everyone to ignore it.
+        echo "  On-device TTS configured, but no speech engine applies to this CPU architecture"
+        echo "  (no Jetson CUDA build and no pinned Piper artifact) — this box answers speech with SILENCE"
         ;;
       1)
         # An engine SURVIVES here — install-voice.sh returns 13, not 1, when
@@ -3509,7 +3529,21 @@ step_post_update() {
   # tts-local-cli provider. Without this call the whole of TASK-383 would be
   # fresh-install-only, and every already-shipped box would keep answering a
   # spoken request with silence.
-  step_openclaw_tts || echo "  Warning: openclaw_tts step failed (non-fatal)"
+  # The SAME tolerance table step_openclaw_setup applies to this step, and for
+  # the same reason: 12, 13 and 14 are three different facts about a box's
+  # speech and the update path has to keep them apart too. A single
+  # `|| echo "(non-fatal)"` — indistinguishable from the fourteen lines around
+  # it — reported "this box answers speech with SILENCE" in the same words as a
+  # skipped VNC refresh, on the very path that reaches ALREADY-SHIPPED boxes.
+  local TTS_UPDATE_RC=0
+  step_openclaw_tts || TTS_UPDATE_RC=$?
+  case "$TTS_UPDATE_RC" in
+    0) ;;
+    12) echo "  Warning: Kokoro GPU TTS did not install (recorded above; the update continues)" ;;
+    13) echo "  Warning: this box has NO working TTS engine — it answers speech with SILENCE (recorded above; the update continues)" ;;
+    14) echo "  Warning: the TTS install did not complete (recorded above; the update continues)" ;;
+    *)  echo "  Warning: openclaw_tts returned $TTS_UPDATE_RC, which is not in its contract (non-fatal)" ;;
+  esac
   # Re-assert the gateway service after an in-app update. The full update
   # syncs repo files and rebuilds before this continuation runs; older devices
   # can therefore reach the new UI while the gateway is still using stale
@@ -4871,8 +4905,14 @@ step_validate_services() {
       # spoken request with silence.
       local tts_state="" piper_state=""
       if [ -r "$TTS_STATUS_FILE" ]; then
-        tts_state=$(sed -n 's/^KOKORO=//p' "$TTS_STATUS_FILE" 2>/dev/null | tail -1)
-        piper_state=$(sed -n 's/^PIPER=//p' "$TTS_STATUS_FILE" 2>/dev/null | tail -1)
+        # `tr -d '\r'`: the file is written by a shell on the device, but it is
+        # also restored from tarballs and edited by hand, and a CRLF line ends
+        # the verdict as `ready\r` — a value that is neither `ready` nor any
+        # other word in the vocabulary. Parse the line rather than merely
+        # refusing it; what a garbled value must NOT do is score a pass, and
+        # that is what the guard below is for.
+        tts_state=$(sed -n 's/^KOKORO=//p' "$TTS_STATUS_FILE" 2>/dev/null | tr -d '\r' | tail -1)
+        piper_state=$(sed -n 's/^PIPER=//p' "$TTS_STATUS_FILE" 2>/dev/null | tr -d '\r' | tail -1)
       fi
       # `ready` is the only verdict that means "this engine can speak".
       # `skipped:*` is a board that was never going to run it, `failed:*` is one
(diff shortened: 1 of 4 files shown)


KrasimirKralev:
@coderabbitai both halves of that finding are right, and both are now fixed in 5b7499b6.

**Empty suffix.** `skipped:` does match `skipped:*`, and a torn write is exactly how a reasonless verdict appears — which is the failure mode this arm exists for, so accepting one as a legitimate board decline would have left a hole in the middle of the fix. Both patterns now require a reason (`skipped:?*` / `failed:?*`), in the probe and in the two new verdict switches in `scripts/install-voice.sh`. The real publishers only ever emit `skipped:no-cuda`, `skipped:arch-<arch>` and `failed:<stage>`, so nothing legitimate is affected.

**Ordering.** Right, and it is the better argument. My comment justified putting the arm last on the grounds that "a garbled Piper verdict next to a failed Kokoro is still best reported as no working engine" — but that reasoning does not survive being turned around: `KOKORO=redy` + `PIPER=failed:download` reported "this box has NO working TTS engine — it will answer every spoken request with SILENCE" over a GPU engine that may be running perfectly. That is a failure report over something that succeeded, which is the same defect this PR is about, one register down. The arm now sits second, ahead of every arm that names an engine state and behind only "no verdict at all" (which is strictly more specific: both keys absent).

It stays an `elif` rather than the trailing `else` your original outside-diff note suggested, for a reason worth recording: the chain's fall-through is also where the **healthy** box lands — `KOKORO=ready` + `PIPER=ready` matches no arm either — so an `else` there would fail every good box on the shelf. `src/tests/unit/install-tts-verdict-siblings.test.ts` pins that.

Both regression cases you asked for are in the same commit:
- `KOKORO=redy` + `PIPER=failed:download` → fails as *unrecognised*, and asserts the output does **not** say "NO working TTS engine".
- `KOKORO=skipped:no-cuda` + `PIPER=skipped:` → fails rather than passing as a board decline.

New suite is 24 tests, all green, plus #519's own 23 still green.

KrasimirKralev:
@coderabbitai both halves of the outside-diff finding on `install.sh:4927-4970` were right, and both are fixed in 5b7499b6:

1. **`skipped:` / `failed:` with the reason cut off.** `skipped:*` matches a bare `skipped:` in bash, so a truncated write — the exact torn-write case the guard exists for — could still pass as a legitimate board decline. Both `case` patterns now require a reason: `skipped:?*` / `failed:?*`.
2. **Ordering.** The unreadable arm was placed last, so `KOKORO=redy` next to `PIPER=failed:download` was reported as "this box has NO working TTS engine" — a claim about a GPU engine that may be running perfectly, i.e. the same failure-report-over-a-success this PR exists to remove. It is now second, ahead of every arm that names an engine state and behind only "no verdict at all".

It stays an `elif` rather than the trailing `else` the proposed diff closes with, deliberately: the chain's fall-through is also where the **healthy** box lands (`KOKORO=ready` + `PIPER=ready` matches no arm), so an `else` there would fail every good box on the shelf. `install-tts-verdict-siblings.test.ts` pins that healthy case alongside the two new ones.

Regression coverage added with the fix, both red against unmodified `beta`:
- `does not accept a skip with its reason truncated away` (empty reason)
- `does not read an engine state out of a verdict it could not parse` (malformed Kokoro next to a failed Piper)

Suite is 24 tests: 13 failed / 11 passed against unmodified `origin/beta`, 24 passed with the fix.

PR #551, opened 2026-08-28
fix: a box with no working TTS engine must fail provisioning, not pass it

KrasimirKralev:
## Summary

Follow-up to #519, #533 and #544. All three removed a false success in this chain and all three kept the same carve-out, in the same words: when **both** engines report `skipped:*` the run is graded clean. The ruling from the product owner is that the carve-out goes — *"a silent box is not a healthy box."*

**A box that finishes provisioning with no working speech engine now FAILS, naming every engine it tried and the concrete reason each one is not there.**

### The carve-out, and why nothing legitimate was behind it

Three layers each let it through:

| layer | what it did with two `skipped:*` verdicts |
| --- | --- |
| `install-voice.sh` `--tts-only` | printed `=== No on-device TTS engine applies to this board ===` on **stdout** and `exit "$KOKORO_RC"` (10 or 11) |
| `install.sh` `step_openclaw_tts` | `11)` set `TTS_RC=0` and called no `record_provision_failure` |
| `install.sh` `step_validate_services` | the mute-box arm carried `&& { kokoro_failed || piper_failed }`, so two skips matched **no arm** and fell out of the chain as a silent PASS |

The summary line that same step printed for `VOICE_RC=11`, verbatim from `beta`:

```
  On-device TTS configured, but no speech engine applies to this CPU architecture
  (no Jetson CUDA build and no pinned Piper artifact) — this box answers speech with SILENCE
```

on stdout, `TTS_RC=0`, `PROVISION_FAILURES` empty, and then `=== ClawBox Setup Complete ===`. A run that describes a mute box in its own words and grades it healthy is not a check anyone can act on.

**The justification does not hold up.** The comment defending it — repeated in #519, #533 and #544 — is that failing a board neither engine ships for *"would only teach everyone to ignore this check"*, and it names `install-x64.sh` as the run it would fail. That run does not exist:

```
$ grep -c 'voice\|tts\|TTS\|VOICE\|piper\|kokoro' install-x64.sh
0
$ grep -rn 'install-voice.sh' --include='*.sh' . | grep -v '^./scripts/bench'
./install.sh:2602:    bash "$PROJECT_DIR/scripts/install-voice.sh" --tts-only
```

`install-x64.sh` never calls the script that publishes these verdicts, so **no shipped run lands in the carve-out**. What does land in it is `install.sh` run where `uname -m` is not aarch64 — the only way both halves skip, since `install_piper_engine`'s single non-failure skip is the architecture test (`install-voice.sh:199`) and `install_kokoro_tts` returns 11 on that same board (`install-voice.sh:616`). The outcome is a box with no speech, reported as a healthy one.

### The fix

**One question, asked at every call site: is either engine `ready`?**

`skipped:*` still says *why* an engine is absent — it is the sentence the operator needs, and a new `tts_verdict_explain` writes it — but it no longer decides whether the run passed. `skipped:*` means "nothing is missing" only while the *other* half is carrying the box.

`scripts/install-voice.sh`

- **`--tts-only`**: the two-outcome guard collapses to one. No `ready` engine is `exit 13`, on stderr, naming both engines and the concrete reason each is absent. `11` is no longer reachable as an exit status of this mode.
- **`--piper-only`** *(sibling)*: `skipped:?*` printed "No Piper artifact applies to this board" and **exited 0** — in the mode `clawbox-tts.sh:596` tells an operator to run *because* the box will not speak. The mute-box question is now asked first, of **both** halves (`tts_status_load` has already read the Kokoro verdict off disk for exactly this reason), and placed ahead of the `PIPER_ONLY_RC` guard because "no engine at all" outranks "the CPU half specifically did not land".
- **The full pipeline** *(sibling)*: #544 taught its two engine lines to read the verdicts and left the status alone, so a run that printed *"no Kokoro GPU engine applies to this board"* **and** *"no Piper fallback applies to this board"* still exited 0 — the script ends on an `echo`. Same guard, same code.
- `tts_verdict_is_failure` is **removed**: it existed only to separate "both failed" from "both skipped", and that separation is the defect.

`install.sh`

- **`11)`** is a recorded provisioning failure returning `TTS_RC=13`. It is unreachable from the fixed `--tts-only`, and it stays and fails anyway — a code whose meaning is "no engine applies to this board" must never be the one that grades clean if some future path revives it. Leaving the lenient branch behind is exactly how the siblings in #519/#533/#544 stayed unguarded after each fix.
- **`13)`**'s banner now names `KOKORO=` and `PIPER=` from the verdict file. 13 covers the declining board as well as the failing one, and "neither engine is here" is not actionable on its own — a skip and a failed download are different repairs.
- The `11` summary line moves to **stderr**, with the other failures rather than inside the success summary.
- **`step_validate_services`**: the mute-box arm drops the `failed:*` precondition.

#### One thing got *narrower*, deliberately

The mute-box arm asserts SILENCE only when **both** verdicts are present (`-n` on each). One half non-`ready` and the other **unreported** still fails, but under a separate arm that says *"no engine on this box is confirmed ready"* — an unreported half is unknown, not absent, and asserting a mute box from a missing line is the failure-report-over-a-success this chain keeps producing.

That also tightens the arms below it. Past that point exactly one engine is `ready`, so `kokoro_failed` is reached only with a ready Piper and `piper_failed` only with a ready Kokoro — each finally states a fallback it can prove. `KOKORO=failed:model` with no `PIPER=` line used to match the Kokoro arm, whose sentence ends *"— this box answers speech on the Piper CPU fallback"*: a fallback claimed from an **unreported** verdict. A test pins both halves of that.

### Fixing the class, and how I know I found every site

The class is *"a verdict that explains an engine's absence being read as permission to pass."*

```
rg -n 'skipped:'  -g '*.sh' -g '*.ts' -g '*.py'   # -> only install.sh + install-voice.sh decide anything on it
rg -n 'TTS_STATUS_FILE|tts-status'                # -> 3 readers in install.sh, 1 in install-voice.sh; all strip CR
rg -n 'install_piper\b'                           # -> defn + 3 callers: --tts-only, --piper-only, full pipeline
rg -n 'step_openclaw_tts'                         # -> defn + step_openclaw_setup:1643, step_post_update:3589,
                                                  #    and the --step whitelist
```

| site | verdict |
| --- | --- |
| `install-voice.sh` `--tts-only` both-declined arm | **fixed** — exit 13 |
| `install-voice.sh` `--piper-only` `skipped:?*` arm | **fixed** — sibling |
| `install-voice.sh` full-pipeline tail | **fixed** — sibling |
| `install.sh` `step_openclaw_tts` `11)` | **fixed** — recorded, TTS_RC=13 |
| `install.sh` `step_openclaw_tts` `13)` banner | **fixed** — names both verdicts |
| `install.sh` summary `11)` arm | **fixed** — stderr |
| `install.sh` probe mute-box arm | **fixed** — drops the `failed:*` precondition |
| `install.sh` `step_openclaw_setup:1643` tolerance table | **already correct** — 13 is tolerated, warned and recorded; it now simply receives more cases |
| `install.sh` `step_post_update:3589` tolerance table | **already correct** — same table, same reason |
| `install.sh` `--step` dispatch EXIT trap | **already correct** — publishes `incomplete` from `PROVISION_FAILURES`, which this PR now populates for 11 |
| `install.sh` summary `10\|12` "Piper CPU only" | **sound, left alone** — reachable only with a ready Piper (`--tts-only` returns 13 or 1 otherwise); pinned by a test |
| `install-voice.sh:738-740` "(Kokoro GPU, Piper fallback)" | **sound, left alone** — #544's proof still holds: reached only with `PIPER_RC=0`/`DEPLOY_RC=0` past the new guard |
| `src/lib/local-models.ts`, dashboard, API | **no consumer** — nothing outside these two scripts reads the verdict |

## Type of change

- [x] Bug fix (non-breaking)
- [ ] New feature (non-breaking)
- [ ] Breaking change
- [ ] Documentation / tooling
- [ ] Other: ___

Non-breaking in the sense that matters: a box that could not install its speech engines still finishes provisioning and comes up reachable, which is how it gets fixed. What changes is that the run now says so, in the exit status, the provisioning marker and the health check.

## How was this tested?

- [x] `bun run lint` passes — `eslint` clean on all four test files
- [x] `bun run test` passes — see the counts below
- [ ] `bun run build` succeeds — not run; no application code in this diff (two shell scripts and four test files). CI's `build-identity`, `test`, `e2e` and `e2e-install` are green.
- [x] Manually verified on device — Jetson Orin, `aarch64`, `R36 (release), REVISION: 5.0`

### Unit tests — red first

New `src/tests/unit/install-tts-mute-box-fails.test.ts`, 30 tests, executing the shipped artifacts against stubs: the real `--tts-only` and `--piper-only` dispatches, the real engine-report block at the end of the full pipeline, the real `step_openclaw_tts`, the real `step_validate_services`.

- Against unmodified `origin/beta` (`3694ae72`, sources restored, final test files in place): **18 failed / 107 passed** across the four touched suites.
- With the fix: **247 passed / 0 failed** across those four plus `install-tts-survivor-verdict`, `install-tts-provider-registry`, `clawbox-tts`, `install-provision-status-marker`, `install-post-update-units`, `root-steps`, `observability-install`.
- `install-edition-switch-refusal`, `install-foreign-edition-teardown`, `desktop-power-wiring` and that group: **139 passed**, unchanged.
- The new suite run four consecutive times: 30/30 each time.

Roughly half the new suite are **over-correction guards**, and they are the point of half the diff — a box with one engine is not a mute box:

- a no-CUDA Orin that speaks on Piper still exits 10 and provisions clean
- a healthy box still exits 0
- a lost fallback behind a working Kokoro is still 1 -> 14, never 13
- `--piper-only` on a board with no artifact still exits 0 **when a ready Kokoro is on record**
- `KOKORO=ready` + `PIPER=skipped:arch-*` still passes the probe
- an unreadable verdict still outranks the engine-naming arms (#533's guard)
- an unreported half is still reported as unassertable, not as proven silence

#### Assertions deliberately reversed

Four existing tests pinned the behaviour the ruling overturns. Each is reversed in place with the reason inline, rather than deleted:

- `install-tts-no-engine`: *"passes a board that was never going to run either engine"* -> **FAILS**; *"does not cry 'no engine'..."* -> exit 11 -> **13**
- `install-tts-verdict-siblings`: *"still exits 0 there..."* -> **13** (plus a new guard for the seeded-Kokoro case); *"still records nothing for a board neither engine ships for (11)"* -> **RECORDS**
- `install-kokoro-tts`: `it.each([[10],[11]])` "exit %i is a SKIP" -> 10 only, with 11's move explained; the exit-11 step test now asserts the recorded failure and reads the two lines off **stderr**

`bash -n` clean on both scripts. `tsc --noEmit`: **24 errors before and after**, the same pre-existing set #533 and #544 both reported; the one hit inside a file this PR touches is `install-kokoro-tts.test.ts:202`, in the harness's `ProcessEnv` cast, untouched by this diff.

*Environment note:* the Windows dev host cannot run these spawn-heavy suites. All counts above were taken on Linux (WSL Ubuntu 22.04, node 22, `npm ci` from the committed lockfile). Linux CI is the gate.

### Before / after, on a real Orin

`aarch64`, `R36 (release), REVISION: 5.0`. Both versions of `scripts/install-voice.sh` fetched onto the board from this repo and run there, with `PIPER_DIR` and the verdict file redirected to scratch paths so no box state was mutated. `uname` is shimmed to `x86_64` for A and B — that is the only way to make a real Orin take the both-declined path, and everything else about the run is the shipped script.

**A — `beta` (`a1021dee`), board declines both engines:**

```
=== On-device TTS (Kokoro GPU + Piper fallback) ===
  Skipping Piper: no pinned artifact for x86_64 (ClawBox ships aarch64)
CLAWBOX_TTS_PIPER=skipped:arch-x86_64
  Skipping Kokoro: no Jetson CUDA build for x86_64 (ClawBox ships aarch64)
CLAWBOX_TTS_KOKORO=skipped:arch-x86_64
=== No on-device TTS engine applies to this board (...) ===     <- stdout
### exit status: 11                                             <- install.sh grades this clean
```

**B — this branch, same board, same shim:**

```
CLAWBOX_TTS_PIPER=skipped:arch-x86_64
CLAWBOX_TTS_KOKORO=skipped:arch-x86_64
=== NO WORKING TTS ENGINE - this box will answer every spoken request with SILENCE ===
===   Kokoro (GPU): SKIPPED (arch-x86_64) - this board does not run it
===   Piper (CPU):  SKIPPED (arch-x86_64) - this board does not run it
===   Verdicts recorded in /tmp/tts-b/status
### exit status: 13
```

**C — this branch, REAL aarch64 board, nothing shimmed:** no regression.

```
  Piper binary installed at /tmp/tts-c2/piper/piper
  Piper voice ready: en_US-lessac-medium
CLAWBOX_TTS_PIPER=ready
  CUDA detected: Build cuda_12.6.r12.6/compiler.34714021_0
CLAWBOX_TTS_KOKORO=ready
=== On-device TTS ready (Kokoro GPU, Piper fallback) ===
### exit status: 0
KOKORO=ready
PIPER=ready
```

**E — `--piper-only` on the real board, this branch:** `=== Piper fallback ready ===`, exit 0, `PIPER=ready`. **F — the box still speaks:** the shipped entrypoint returns a 24 kHz Kokoro WAV, exit 0.

The board's `/etc/clawbox/tts-status`, its installed Piper, and its `kokoro-server.service` (md5 identical before and after) were left untouched.

**And the full-install harness agrees:** `e2e-install` — a real systemd container running `install.sh` end to end — is **green on this branch**. It runs arm64 (native runner, or qemu on x86), so `uname -m` is `aarch64` there and Piper is always asked for; the both-declined path is not reachable in it, which is exactly the point made below about blast radius.

### Does voice actually work on our hardware?

This was the more important half of the brief, and the answer is **yes** — the "both engines skip on our own boards" hypothesis does **not** reproduce. Measured on a live Orin (`aarch64`, `R36 (release), REVISION: 5.0`) running the real shipped script from `beta` HEAD, with `PIPER_DIR` and the status file redirected to scratch paths so no box state was mutated:

```
$ bash /home/clawbox/clawbox/scripts/install-voice.sh --tts-only
=== On-device TTS (Kokoro GPU + Piper fallback) ===
  Downloading Piper 2023.11.14-2...
  Piper binary installed at /tmp/ttsprobe/piper/piper
  Piper voice ready: en_US-lessac-medium
CLAWBOX_TTS_PIPER=ready
  CUDA detected: Build cuda_12.6.r12.6/compiler.34714021_0
  Kokoro already installed by a previous run — skipping the GPU install
CLAWBOX_TTS_KOKORO=ready
=== On-device TTS ready (Kokoro GPU, Piper fallback) ===
### exit status: 0

$ cat /tmp/tts-probe-status
KOKORO=ready
PIPER=ready
```

Both engines reach `ready`, and both genuinely speak:

```
$ ~/.openclaw/workspace/scripts/openclaw/clawbox-tts.sh "ClawBox voice check one two three." /tmp/tts-probe.wav
$ file /tmp/tts-probe.wav
/tmp/tts-probe.wav: RIFF (little-endian) data, WAVE audio, Microsoft PCM, 16 bit, mono 24000 Hz   # Kokoro

$ /tmp/ttsprobe/piper/piper --model .../en_US-lessac-medium.onnx --output_file /tmp/piper-probe.wav
[piper] [info] Real-time factor: 0.126 (infer=0.18 sec, audio=1.44 sec)
$ file /tmp/piper-probe.wav
/tmp/piper-probe.wav: RIFF (little-endian) data, WAVE audio, Microsoft PCM, 16 bit, mono 22050 Hz  # Piper
```

`detect_cuda` resolves `nvcc` through its `$CUDA_HOME_DIR/bin/nvcc` fallback (it is **not** on the default PATH on JetPack, and that fallback is what saves it), `import kokoro, torch` succeeds under the unit's `LD_LIBRARY_PATH` with `torch.cuda.is_available() == True`, and the pinned Piper tarball and voice still download and verify against their sha256.

So `skipped:*` for **both** engines is not something aarch64 hardware can reach: on aarch64 Piper is always *asked* for, and its outcome is `ready` or `failed:*` — the latter already exits 13. **This PR's honest blast radius is therefore `install.sh` run on a non-aarch64 host**, plus every future path that reaches "no engine" by a route nobody has thought of yet, which is what the three sibling PRs before this one were each about.

## Checklist

- [x] Branch is up to date with the target branch (rebased on `3694ae72`)
- [x] Follows the conventions in CLAUDE.md and CONTRIBUTING.md
- [x] No secrets, tokens, or credentials committed
- [x] Added/updated tests where it makes sense
- [ ] Updated docs where user-facing behavior changed — no doc change: the user-facing behaviour that changes is a provisioning run failing where it used to pass, and the contract it belongs to is documented in the scripts' own comments, which this diff updates

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -2631,7 +2631,32 @@ step_openclaw_tts() {
   case "$VOICE_RC" in
     0)  KOKORO_READY=true ;;
     10) KOKORO_REASON="no CUDA toolkit on this board" ;;
-    11) KOKORO_REASON="no Jetson CUDA build for this CPU architecture" ;;
+    11) KOKORO_REASON="no Jetson CUDA build for this CPU architecture, and no pinned Piper artifact either"
+        # ── "Nothing applies to this board" is still a box that cannot speak ──
+        # 11 was deliberately left clean: no TTS_RC, no recorded failure, on the
+        # grounds that a board neither engine ships for was never asked for one
+        # and that failing every such run "would only teach everyone to ignore
+        # the check". The summary this arm then printed said, in its own words,
+        # "this box answers speech with SILENCE" — and the run finished as a
+        # healthy provision. A run that describes a mute box and grades it
+        # healthy is not a check anyone can act on.
+        #
+        # install-voice.sh now answers that board with 13, so 11 should never
+        # arrive here again. The branch stays, and it fails, because a code
+        # whose meaning is "no engine applies to this board" must never be the
+        # one that grades clean if any future path revives it — leaving the
+        # lenient branch behind is exactly how the sibling call sites in #519,
+        # #533 and #544 stayed unguarded after each fix.
+        TTS_RC=13
+        echo "  ############################################################" >&2
+        echo "  # This box has NO working TTS engine — no Kokoro GPU build and" >&2
+        echo "  # no pinned Piper artifact apply to this CPU architecture." >&2
+        echo "  # It will answer every spoken request with SILENCE." >&2
+        echo "  # Engines and reasons: $TTS_STATUS_FILE" >&2
+        echo "  # Re-run:  sudo bash $PROJECT_DIR/install.sh --step openclaw_tts" >&2
+        echo "  ############################################################" >&2
+        record_provision_failure openclaw_tts
+        ;;
     12) KOKORO_REASON="the Kokoro GPU install failed, see the log above"
         # ── A hard failure must stop arriving as a soft fallback ─────────────
         # This is the branch that made a shipped defect invisible. Kokoro's
@@ -2674,6 +2699,14 @@ step_openclaw_tts() {
         echo "  ############################################################" >&2
         echo "  # This box has NO working TTS engine — neither Kokoro nor Piper." >&2
         echo "  # It will answer every spoken request with SILENCE." >&2
+        # Both engines named with the reason each one is absent. 13 now covers
+        # the board that DECLINED both halves as well as the one whose install
+        # failed, and "neither engine is here" is not an actionable report on
+        # its own — which half declined for want of CUDA and which one failed
+        # its download lead to different fixes. The verdicts are what the run
+        # published, so they are what gets printed.
+        echo "  # Kokoro (GPU): ${KOKORO_VERDICT:-no verdict published}" >&2
+        echo "  # Piper (CPU):  ${PIPER_VERDICT:-no verdict published}" >&2
         echo "  # Re-run:  sudo bash $PROJECT_DIR/install.sh --step openclaw_tts" >&2
         echo "  ############################################################" >&2
         record_provision_failure openclaw_tts
@@ -2840,20 +2873,16 @@ step_openclaw_tts() {
         # install_kokoro_tts's "no Jetson CUDA build for this architecture",
         # i.e. `uname -m` is not aarch64 — which is exactly when
         # install_piper_engine has no pinned artifact either and publishes
-        # `skipped:arch-*` of its own. install-voice.sh has just printed "=== No
-        # on-device TTS engine applies to this board ===" for that pair, so
-        # naming a Piper CPU fallback one line later put the name of an engine
-        # this box does not have on the summary line an operator reads.
+        # `skipped:arch-*` of its own. So 11 is a box with NO engine.
         #
-        # Deliberately NOT a recorded provisioning failure: nothing was asked
-        # for and nothing is missing on a board neither engine ships for, which
-        # is the same rule `skipped:*` carries in step_validate_services.
-        #
-        # On stdout, not stderr, for the same reason: install-voice.sh prints
-        # its "no engine applies to this board" line on stdout too, and painting
-        # every install-x64.sh run red would only teach everyone to ignore it.
-        echo "  On-device TTS configured, but no speech engine applies to this CPU architecture"
-        echo "  (no Jetson CUDA build and no pinned Piper artifact) — this box answers speech with SILENCE"
+        # It used to print these two lines on stdout and record nothing, which
+        # meant the run said "this box answers speech with SILENCE" and then
+        # graded itself healthy. It is a recorded provisioning failure now (see
+        # the 11 arm above), so this line reports the same fact the rest of the
+        # step reports, on stderr with the other failures rather than in the
+        # middle of the success summary.
+        echo "  On-device TTS configured, but no speech engine applies to this CPU architecture" >&2
+        echo "  (no Jetson CUDA build and no pinned Piper artifact) — this box answers speech with SILENCE" >&2
         ;;
       1)
         # An engine SURVIVES here — install-voice.sh returns 13, not 1, when
@@ -4932,11 +4961,17 @@ step_validate_services() {
     # Probe: on-device TTS delivered the engine it was asked for.
     #
     # scripts/install-voice.sh publishes its verdict to $TTS_STATUS_FILE for
-    # exactly this check. The distinction the file carries is the whole point:
-    # `skipped:*` means this board was never going to run Kokoro (no CUDA, no
-    # Jetson build for its architecture) and is a PASS, while `failed:*` means
-    # the GPU engine was requested and did not arrive — the box is on the Piper
-    # CPU fallback and someone has to fix it.
+    # exactly this check. The distinction the file carries tells an operator
+    # WHAT to fix: `skipped:*` means this board was never going to run that
+    # engine (no CUDA, no Jetson build for its architecture), `failed:*` means
+    # it was requested and did not arrive.
+    #
+    # What the distinction does NOT decide is whether the check passes. That is
+    # decided by one question asked of both engines together — is either one
+    # `ready`? A `skipped:*` engine is a PASS only while the other half is
+    # carrying the box; two skipped engines is a box with no voice, and it used
+    # to score healthy here because the arm below asked for a `failed:*` before
+    # it would report one.
     #
     # An ABSENT verdict fails too. The TTS step runs before this check on both
     # the install and the update path, so nothing here can assert "Kokoro is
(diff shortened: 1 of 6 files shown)


KrasimirKralev:
Note on the one red `test` run (commit `153fb43`, since re-run):

```
FAIL  src/tests/routes/clawkeep-memory.test.ts > POST /setup-api/clawkeep/memory/index
      > never returns the pid it is running under
Tests  1 failed | 7791 passed (7792)
```

Unrelated to this diff, and pre-existing. The test writes `startedAtMs: Date.now()` into the run state and then asserts the response JSON does not contain `String(process.pid)`:

```ts
await fs.writeFile(..., JSON.stringify({ ..., startedAtMs: Date.now(), childPid: process.pid }));
const res = await indexPOST(...);
expect(JSON.stringify(await res.json())).not.toContain(String(process.pid));
```

`index/route.ts` returns `{ accepted, run }`, so that 13-digit timestamp is echoed back. A short Linux pid is a substring of a 13-digit epoch-millisecond number often enough to matter, so the assertion fails on an unlucky pid and passes on every other one — it ran green 8/8 locally and green on the previous commit of this same branch. Worth tightening (assert on the `childPid` field rather than on a substring of the whole document), but out of scope here: different subsystem, and nothing in this PR touches it.

This branch changes only `install.sh`, `scripts/install-voice.sh` and four `install-*` test files. `e2e-install` — a real systemd container running `install.sh` end to end on arm64 — is green.

PR #561, opened 2026-08-30
A box whose password was set outside the wizard can be logged into again

yalexx:
Found during a QA sweep of a freshly-flashed unit on `beta` @ `cac44028`. The box could not be claimed at all.

## The deadlock

`/etc/shadow` is the authority for "does this box have an owner"; the `password_configured` flag in `data/config.json` is only a cache of it. `system/credentials` already knows this (TASK-444a). `/login-api` did not — it gated on the flag alone. When the two disagreed, both onboarding gates refused:

| Endpoint | Consults | Answer |
|---|---|---|
| `/login-api` | `password_configured` flag → falsy | `400 Password not configured. Complete setup first.` |
| `/setup-api/system/credentials` | `/etc/shadow` → owned | `401 Authentication required` |

The wizard cannot claim the box, and the owner cannot log in. Nothing in the UI recovers it — only an SSH session that resets the account back to the factory default.

## Observed on hardware

```
$ passwd -S clawbox
clawbox P 08/30/2026 0 99999 7 -1          # a usable password is set

$ printf "clawbox\0" | /usr/sbin/unix_chkpwd clawbox nonull; echo $?
7                                           # PAM_AUTH_ERR — not the factory default

$ cat data/config.json
{"setup_progress_step": 3, "hostname": "clawbox"}   # no password_configured key

$ curl -s -o /dev/null -w "%{http_code}\n" -X POST http://127.0.0.1/login-api \
    -H "Content-Type: application/json" -d "{\"password\":\"<the owner password>\"}"
400          {"error":"Password not configured. Complete setup first."}

$ curl -s -o /dev/null -w "%{http_code}\n" -X POST http://127.0.0.1/setup-api/system/credentials \
    -H "Content-Type: application/json" -d "{\"password\":\"<a new password>\"}"
401          {"error":"Authentication required"}
```

Restoring the account to the factory default is the only thing that breaks the tie — after `chpasswd` puts `clawbox` back, the unmodified wizard call succeeds immediately, which isolates the account state as the sole cause:

```
$ echo clawbox:clawbox | sudo chpasswd
$ curl -s -w "%{http_code}\n" -X POST http://127.0.0.1/setup-api/system/credentials \
    -H "Content-Type: application/json" -d "{\"password\":\"<a new password>\"}"
200          {"success":true,"authenticated":true}
```

The state needs nothing exotic to reach: an imaging rig that pre-sets the account password, a restore that brings back the OS account but not `data/config.json`, or an installer who ran `passwd` over SSH before opening the wizard.

## The fix

When the flag is falsy and setup is not complete, ask `hasOwnerPassword()` before refusing. If `/etc/shadow` says the account carries a password somebody set, let the login proceed and heal the stale flag.

## Why this does not weaken the gate

`hasOwnerPassword()` answers:

- `false` for **no password** and for the **published factory default** → an as-flashed box still gets the `400` and still cannot be logged into with the default that is in this repository
- `null` when `/etc/shadow` cannot be read at all → stays closed
- `true` only for a password somebody deliberately set → and the caller must still prove it via `verifyPassword`

## Tests

`src/tests/routes/login-api.test.ts` — 5 new cases, 3 of which fail on `beta` and pass here:

- ✅ lets the owner log in when shadow says the account is owned *(fails before)*
- ✅ heals the stale flag so the next request takes the normal path *(fails before)*
- ✅ still rejects the wrong password on such a box *(fails before)*
- ✅ keeps refusing an as-flashed box holding only the factory default *(guards the security property)*
- ✅ refuses when the shadow state cannot be read *(guards the `null` case)*

Full suite: **8423 passed / 579 files**, no regressions.

## Summary by CodeRabbit

- **Bug Fixes**
  - Login now recognizes an existing configured system password even when the setup status is stale.
  - Successful authentication repairs the password configuration status automatically.
  - Systems without a configured owner password continue to require initial setup.
  - Invalid credentials, factory-default accounts, and unreadable password states remain blocked.

### DIFF
--- a/src/app/login-api/route.ts
+++ b/src/app/login-api/route.ts
@@ -1,6 +1,7 @@
 import { NextResponse } from "next/server";
 import { get, set } from "@/lib/config-store";
 import { verifyPassword, createSessionCookie, getSessionSigningSecret, getSessionGeneration } from "@/lib/auth";
+import { hasOwnerPassword } from "@/lib/system-password";
 import {
   checkLockout,
   recordFailure,
@@ -85,6 +86,28 @@ export async function POST(request: Request) {
     const setupComplete = await get("setup_complete");
     if (setupComplete) {
       // Auto-migrate: user completed setup before auth was added
+      await set("password_configured", true);
+      await set("password_configured_at", new Date().toISOString());
+    } else if ((await hasOwnerPassword()) === true) {
+      // /etc/shadow is the authority, the flag is only a cache — the same
+      // argument system/credentials already makes (TASK-444a). When the two
+      // disagree in this direction the box is otherwise UNCLAIMABLE:
+      //
+      //   this route          — flag says "no password"  → 400 "Complete setup first"
+      //   system/credentials  — shadow says "owned"      → 401 "Authentication required"
+      //
+      // Neither gate can be satisfied, so the wizard cannot claim the box and
+      // the owner cannot log in. Nothing in the UI recovers it. That state is
+      // reachable without anything exotic: a box imaged by a rig that pre-sets
+      // the account password, a restore that brings back the OS account but not
+      // data/config.json, or an installer who just ran `passwd` over SSH before
+      // opening the wizard.
+      //
+      // Trusting shadow here does NOT weaken the gate. hasOwnerPassword() is
+      // false for both "no password" and the published factory default, so an
+      // as-flashed box still gets the 400 and still cannot be logged into with
+      // the default everyone knows — only a password somebody deliberately set
+      // opens this path, and the caller must still prove it below.
       await set("password_configured", true);
       await set("password_configured_at", new Date().toISOString());
     } else {
--- a/src/tests/routes/login-api.test.ts
+++ b/src/tests/routes/login-api.test.ts
@@ -13,6 +13,10 @@ vi.mock("@/lib/auth", () => ({
   getSessionGeneration: vi.fn().mockResolvedValue(0),
 }));
 
+vi.mock("@/lib/system-password", () => ({
+  hasOwnerPassword: vi.fn(),
+}));
+
 vi.mock("@/lib/login-rate-limit", () => ({
   checkLockout: vi.fn(),
   recordFailure: vi.fn(),
@@ -25,6 +29,7 @@ vi.mock("@/lib/login-rate-limit", () => ({
 import * as config from "@/lib/config-store";
 import { verifyPassword, createSessionCookie, getSessionSigningSecret } from "@/lib/auth";
 import { checkLockout, recordFailure, recordSuccess, padResponseTime } from "@/lib/login-rate-limit";
+import { hasOwnerPassword } from "@/lib/system-password";
 
 const mockGet = vi.mocked(config.get);
 const mockSet = vi.mocked(config.set);
@@ -35,6 +40,7 @@ const mockCheckLockout = vi.mocked(checkLockout);
 const mockRecordFailure = vi.mocked(recordFailure);
 const mockRecordSuccess = vi.mocked(recordSuccess);
 const mockPadResponseTime = vi.mocked(padResponseTime);
+const mockHasOwnerPassword = vi.mocked(hasOwnerPassword);
 
 describe("/login-api", () => {
   let POST: (req: Request) => Promise<Response>;
@@ -49,6 +55,9 @@ describe("/login-api", () => {
     mockCheckLockout.mockResolvedValue({ locked: false, retryAfterSeconds: 0 });
     mockRecordFailure.mockResolvedValue({ locked: false, retryAfterSeconds: 0 });
     mockRecordSuccess.mockResolvedValue(undefined);
+    // Default: no owner password on the account, so the flag alone decides and
+    // the pre-existing cases below are unaffected by the shadow fallback.
+    mockHasOwnerPassword.mockResolvedValue(false);
     const mod = await import("@/app/login-api/route");
     POST = mod.POST;
   });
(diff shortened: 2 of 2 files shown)


yalexx:
## Hardware verification on a real box

Verified on a live ClawBox (Jetson, `CLAWBOX_EDITION=openclaw`, claimed +
`setup_complete`). Branch deployed to the device, built **on the box** with
`bun run build` (postbuild ran — `build-info.json: commit=17302d8
branch=fix/unclaimable-box-login-deadlock buildId=0UbC20bpmTnZdeeqGSTHN`), then
`systemctl restart clawbox-setup.service`.

The deadlock state was induced by editing `data/config.json` **only**. The OS
account was never touched — `passwd -S clawbox` reported `clawbox P 08/30/2026 0
99999 7 -1` before, during and after every case below.

---

### Before the fix — the deadlock reproduces

Same box, previous build (`beta` @ `cac44028`), same induced state:

```
--- /login-api with the REAL owner password ---
HTTP/1.1 400 Bad Request
{"error":"Password not configured. Complete setup first."}

--- /setup-api/system/credentials, unauthenticated, as the wizard would ---
HTTP/1.1 401 Unauthorized
{"error":"Authentication required"}

--- /setup-api/setup/status ---
{"setup_complete":false,"password_configured":false,"update_completed":false,
 "wifi_configured":false,"setup_progress_step":null}
```

Neither gate can be satisfied. Confirmed unrecoverable from the UI.

---

### (a) The deadlock state now logs in, and the flag heals — PASS

Measured atomically (induce → request → re-read, one on-box script) with
timestamps, so the write is attributable to this request:

```
### T0  14:33:58.750  induce deadlock state (config.json ONLY)
{ "setup_complete": false, "setup_completed_at": "2026-08-30T14:18:20.088Z" }
/etc/shadow ground truth: clawbox P 08/30/2026 0 99999 7 -1

### T1  14:33:58.862  POST /login-api with the real owner password
HTTP/1.1 200 OK
set-cookie: clawbox_session=<REDACTED>; Path=/; Max-Age=1200; HttpOnly; SameSite=lax
{"success":true}

### T2  14:33:59.207  config.json AFTER
{
  "setup_complete": false,
  "setup_completed_at": "2026-08-30T14:18:20.088Z",
  "password_configured": true,
  "password_configured_at": "2026-08-30T14:33:58.967Z"
}
```

`password_configured_at` (…58.967Z) falls inside the request window
(…58.862 → …59.207).

---

### (b) An as-flashed box is still refused, and stays claimable — PASS

The owner account's password is deliberately not the factory default and could
not be changed for this test, so an as-flashed device was simulated with a
**throwaway local account** carrying `FACTORY_DEFAULT_PASSWORD`:

```
# useradd -m -d /home/qafactory -s /bin/bash qafactory
# echo "qafactory:clawbox" | chpasswd
# passwd -S qafactory
qafactory P 08/30/2026 0 99999 7 -1
```

**Library level** — the unmodified `src/lib/system-password.ts` executed under
both account states:

```
PROBE A — real owner account (password is NOT the factory default)
{ "runningAs": 1000, "CLAWBOX_USER": "clawbox",
  "hasSystemPassword": true, "isFactoryDefaultPassword": false,
  "hasOwnerPassword": true }      <-- fix opens the path

PROBE B — as-flashed account (password IS the factory default)
{ "runningAs": 1001, "CLAWBOX_USER": "qafactory",
  "hasSystemPassword": true, "isFactoryDefaultPassword": true,
  "hasOwnerPassword": false }     <-- fix does NOT open the path
```

**HTTP level** — a second standalone instance of the same build running *as*
that as-flashed account (`CLAWBOX_USER=qafactory
CLAWBOX_ROOT=/tmp/qa-factory-root PORT=3100`), scratch config
`{"hostname":"clawbox","setup_complete":false}`:

```
(b1) POST /login-api {"password":"clawbox","duration":43200}    # the factory default
HTTP/1.1 400 Bad Request
{"error":"Password not configured. Complete setup first."}

(b2) POST /login-api {"password":"anything-at-all","duration":43200}
HTTP/1.1 400 Bad Request
{"error":"Password not configured. Complete setup first."}

(b3) config.json AFTER — the flag was NOT wrongly healed:
{ "hostname": "clawbox", "setup_complete": false }
```

That last line is the point: the published default still logs nobody in, and the
new branch leaves no `password_configured` residue on an unclaimed box.

The *claimability* half is derived rather than executed.
`system/credentials` computes
`passwordAlreadyConfigured = flagSaysConfigured || (await hasOwnerPassword()) === true`
(`src/app/setup-api/system/credentials/route.ts:106-107`) and only calls
`requireSession()` when true; on the as-flashed box both disjuncts are false
(b3 + Probe B), so the session gate is skipped. The request was not fired live
because past that gate the route writes a chpasswd record for
`getSystemUsername()` and starts the root `clawbox-root-update@chpasswd`
service — with a throwaway `CLAWBOX_USER` that is a real password mutation, and
the risk of it resolving to the wrong account was not worth taking.

---

### (c) Normal login on a correctly-claimed box — PASS

```
POST /login-api {"password":"<owner>","duration":43200}
HTTP/1.1 200 OK
{"success":true}

GET /setup-api/system/info   (with that cookie)
HTTP/1.1 200 OK
{"hostname":"clawbox","platform":"Linux 5.15.185-tegra","arch":"arm64","cpus":6,...}

GET /setup-api/system/info   (no cookie)
HTTP/1.1 401 Unauthorized
```

---

### (d) Wrong password still rejected, rate limiting still fires — PASS

```
--- state before: {"byKey":{}}
attempt 1: 401 {"error":"Incorrect password"}
attempt 2: 401
attempt 3: 401
attempt 4: 401
attempt 5: 429  retry-after: 300  {"error":"Too many failed attempts. Try again later."}
attempt 6: 429  retry-after: 300

--- state after:
{"byKey":{
  "cf:198.51.100.99":{"failures":5,"firstFailureAtMs":1788100623626,"lockedUntilMs":1788100924941},
  "global":          {"failures":5,"firstFailureAtMs":1788100623630,"lockedUntilMs":1788100924944}}}

--- CORRECT password, same CF ip:        429  retry-after: 289
--- CORRECT password, DIFFERENT CF ip:   429  retry-after: 289
```

Header rotation buys nothing — the always-on `global` bucket catches the fresh
`CF-Connecting-IP` (TASK-444c), confirmed on hardware.

---

### Cleanup

`data/config.json` restored byte-for-byte
(`sha256 7f8c6212098bd039c1e31832a431ca54b487823fc9286c5d16c54f0e666e3491`);
throwaway account and scratch root removed (`id qafactory` → no such user);
main service `active`. The owner password and the Linux password were never
changed.

`gh pr checks 561`: all 11 checks green.

PR #562, opened 2026-08-30
A box whose primary model is llamacpp can answer again

yalexx:
## Symptom

On a freshly-provisioned Orin Nano, chat is completely non-functional. Every
turn fails before the agent can reply:

```
[gateway] agent model: llamacpp/gemma4-e2b-it-q4_0 (thinking=medium, fast=off)
[diagnostic] lane task error: lane=main
    error="FailoverError: Unknown model: llamacpp/gemma4-e2b-it-q4_0"
[model-fallback/decision] decision=candidate_failed
    requested=llamacpp/gemma4-e2b-it-q4_0 candidate=llamacpp/gemma4-e2b-it-q4_0
    reason=model_not_found next=none
Embedded agent failed before reply: Unknown model: llamacpp/gemma4-e2b-it-q4_0
```

Driven through the real gateway WebSocket (`connect` → `chat.send`, exactly what
`ChatApp.tsx:410-420` sends), the user sees:

```
chat state=final
"⚠️ Agent failed before reply: Unknown model: llamacpp/gemma4-e2b-it-q4_0."
```

## Root cause

`~/.openclaw/openclaw.json` on the affected box:

```json
"models":  { "providers": {} },
"agents":  { "defaults": { "model": { "primary": "llamacpp/gemma4-e2b-it-q4_0" } } }
```

OpenClaw ships an `ollama` plugin but **no llamacpp one** — its own `health`
event lists the loaded plugins as `browser, canvas, device-pair, file-transfer,
memory-core, ollama, phone-control, talk-voice, tts-local-cli`. So `llamacpp/*`
resolves **only** through an explicit `models.providers.llamacpp` entry, and
there isn't one. `openclaw-config.ts:387` states the same rule from the other
side: "a present-but-empty `models.providers.<p>` is still read by the gateway
as a provider definition."

The config backups show how the box got there:

```
openclaw.json.bak.4   12:51:39  providers=[] primary='anthropic/claude-sonnet-4-20250514'
openclaw.json.bak     13:48:09  providers=[] primary='anthropic/claude-sonnet-4-20250514'
openclaw.json         14:06:01  providers=[] primary='llamacpp/gemma4-e2b-it-q4_0'
```

`models.providers` was `{}` from the install onward. At 14:06 the primary flipped
to the local model — that is the "Model migration" block in
`scripts/gateway-pre-start.sh`, which rescues boxes stuck on the retired May-2025
Sonnet id. It writes `primary` and **nothing else**, so it moves the box from one
unresolvable model to another and leaves it just as mute. Its own comment says it
exists because "every chat turn fails before the agent can reply" — which remained
true afterwards.

`ai-models/configure` writes the provider def on the local-AI path
(`src/app/setup-api/ai-models/configure/route.ts:2127-2146`), and
`src/app/setup-api/llamacpp/install/route.ts:210-214` invokes it — but a box that
arrives at llamacpp *through the migration*, or from an image that pre-set
`primary`, never runs that route. The runtime is installed and fine; only the
OpenClaw-side registration is missing.

Two things make this hard to notice: the failure is only visible in the gateway
log, and ClawBox actively reports the broken model as usable —
`/setup-api/chat/model` returns `"available": true` for it, derived from the
llamacpp runtime being present and never cross-checked against
`models.providers`.

## Fix

Repair the config at gateway start, in the same script and immediately after the
migration that creates the state: when the primary or any fallback names
`llamacpp/<model>` and `models.providers.llamacpp` is absent, write the provider
def the local-AI path would have written.

This is the same bug shape — and the same remedy — as the OpenRouter
provider-def migration already in this script ("`auth.profiles.openrouter:default`
set but no `models.providers.openrouter` entry, so OpenClaw's runtime has no
baseUrl to call and every chat turn silently returns `usage: 0/0/0`").

Deliberate refusals:

- **No token, no write.** If `data/.local-ai-token` is missing or shorter than 16
  chars, it writes nothing and logs why. The proxy authenticates
  openclaw → Next.js with a per-install bearer (`src/lib/local-ai-token.ts`);
  writing the entry without it would trade "Unknown model" for a 401 on every
  turn, which is not an improvement.
- **Never overwrites** an existing `models.providers.llamacpp`.
- **Uses the configured model id**, not a hardcoded one.
- Honours `LLAMACPP_CONTEXT_WINDOW`, with the same ≥16384 floor as
  `src/lib/llamacpp.ts`.

## Evidence

Verified on real hardware (Jetson Orin Nano, `CLAWBOX_EDITION=openclaw`) that the
defect reproduces exactly as described above, and that the llamacpp runtime
itself is healthy — through the very proxy OpenClaw is meant to use:

```
POST /setup-api/local-ai/llamacpp/v1/chat/completions   (Bearer <local-ai token>)
{"model":"gemma4-e2b-it-q4_0","messages":[{"role":"user","content":"Reply with exactly: PONG"}]}

HTTP=200  TIME=15.91s      # includes the standby cold start
{"choices":[{"message":{"role":"assistant","content":"PONG"}}],"usage":{"total_tokens":18}}
$ ss -ltn | grep 8080
LISTEN 127.0.0.1:8080      # llama-server woke on demand, as designed
```

So the model answers fine; only the registration was missing.

## Test

`src/tests/unit/gateway-pre-start-llamacpp-provider.test.ts` — 9 cases. It
extracts the migration region out of the **shipped** `.sh` rather than a copy, so
it fails if the real script drifts (same approach as
`gateway-pre-start-clawai-images.test.ts`).

```
# script reverted to pre-fix
 Test Files  1 failed (1)
      Tests  7 failed | 2 passed (9)

# with the fix
 Test Files  1 passed (1)
      Tests  9 passed (9)
```

The 2 that pass either way are the negative cases ("leaves an existing provider
untouched", "does nothing on a non-llamacpp box"), which should hold regardless.

One case reproduces the exact hardware sequence: primary on the retired Anthropic
id with `models.providers: {}` → the migration moves it to
`llamacpp/gemma4-e2b-it-q4_0` **and** the destination is now resolvable.

## Not covered here

Nothing asserts that `agents.defaults.model.primary` names a provider that
actually exists in `models.providers`, and `/setup-api/chat/model` still reports
`available: true` without consulting it. This PR repairs the state at boot; it
does not add that invariant. Worth a follow-up.

## Summary by CodeRabbit

* **New Features**
  * Added automatic setup for Llama.cpp model connections configured as the primary model or a fallback.
  * Supports configurable proxy URLs, model identifiers, and context-window settings.
  * Preserves existing provider configurations, including empty entries, and handles legacy provider setups.

* **Bug Fixes**
  * Repairs missing or invalid Llama.cpp provider configuration during gateway startup.
  * Ignores malformed configuration values and skips changes when authentication credentials are unavailable or invalid.

### DIFF
--- a/scripts/gateway-pre-start.sh
+++ b/scripts/gateway-pre-start.sh
@@ -254,10 +254,14 @@ if isinstance(_primary_now, str):
 _fallbacks_now = model_defaults.get("fallbacks")
 if isinstance(_fallbacks_now, list):
     _llamacpp_refs.extend([f for f in _fallbacks_now if isinstance(f, str)])
-_wants_llamacpp = [r for r in _llamacpp_refs if r.startswith("llamacpp/")]
+# The runtime trims the ref before it checks the prefix, so " llamacpp/x " starts
+# the local runtime but would skip this repair and still fail model resolution.
+_wants_llamacpp = [r.strip() for r in _llamacpp_refs if r.strip().startswith("llamacpp/")]
 
 _mp = cfg.setdefault("models", {}).setdefault("providers", {})
-if _wants_llamacpp and not _mp.get("llamacpp"):
+# Key presence, not truthiness: an existing but empty {} entry is a deliberate
+# operator choice and must be preserved, which .get() would silently overwrite.
+if _wants_llamacpp and "llamacpp" not in _mp:
     # The proxy authenticates openclaw -> Next.js with a per-install bearer
     # (src/lib/local-ai-token.ts). Writing the entry WITHOUT it would trade
     # "Unknown model" for a 401 on every turn, which is not an improvement, so
@@ -281,7 +285,13 @@ if _wants_llamacpp and not _mp.get("llamacpp"):
         )
     else:
         _model_id = _wants_llamacpp[0][len("llamacpp/"):]
-        _ctx = int(os.environ.get("LLAMACPP_CONTEXT_WINDOW") or 0)
+        # A non-numeric override must not abort gateway pre-start: this migration
+        # runs on the path that repairs a mute box, so raising here would turn a
+        # bad env var into a box that never starts at all.
+        try:
+            _ctx = int(os.environ.get("LLAMACPP_CONTEXT_WINDOW") or 0)
+        except ValueError:
+            _ctx = 0
         if _ctx < 16384:
             _ctx = 131072
         _proxy_root = (
--- a/scripts/gateway-pre-start.sh
+++ b/scripts/gateway-pre-start.sh
@@ -223,6 +223,87 @@ if isinstance(primary_model, str) and primary_model.lower() in (
     model_defaults["primary"] = "llamacpp/gemma4-e2b-it-q4_0"
     changed = True
 
+# Migration: a primary (or fallback) that names `llamacpp/<model>` while
+# `models.providers.llamacpp` is absent. OpenClaw ships an `ollama` plugin but
+# NO llamacpp one, so `llamacpp/*` resolves ONLY through an explicit provider
+# entry. Without it every chat turn dies before the agent can reply:
+#
+#   [model-fallback/decision] decision=candidate_failed
+#       requested=llamacpp/gemma4-e2b-it-q4_0 reason=model_not_found next=none
+#   Embedded agent failed before reply: Unknown model: llamacpp/gemma4-e2b-it-q4_0
+#
+# Measured on a freshly-provisioned Orin Nano: `models.providers` was `{}` from
+# the install onward, and the dead-Anthropic migration directly above then moved
+# `primary` onto the local model — swapping one unresolvable id for another and
+# leaving the box just as mute. The device's OWN status route reports the model
+# `available: true` the whole time (it reads the llamacpp runtime, which is
+# genuinely installed and answers on its proxy), so nothing surfaces the gap.
+#
+# `ai-models/configure` writes this entry on the local-AI path, but a box that
+# reaches llamacpp through the migration above — or through an image that
+# pre-set `primary` — never runs that route. Same shape, and the same fix, as
+# the OpenRouter provider-def migration further down.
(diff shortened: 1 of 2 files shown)


yalexx:
### Integration verification (in addition to the unit test)

The unit test runs an *extracted* region of the script. To also check the block
in place — with every surrounding binding real, which would catch an undefined
name or scope error the extracted test cannot — the entire Python heredoc from
the shipped `.sh` was compiled and then run against a synthetic config in exactly
the state measured on hardware:

```
heredoc body lines: 1155 ends at file line 1297
PYTHON HEREDOC COMPILES OK
```

Input — the observed broken state (empty provider map, primary on the retired
Anthropic id):

```json
{"models":{"providers":{}},
 "agents":{"defaults":{"model":{"primary":"anthropic/claude-sonnet-4-20250514"}}}}
```

Result:

```
exit: 0
stdout: Repaired models.providers.llamacpp for llamacpp/gemma4-e2b-it-q4_0
        Updated gateway config

primary  = llamacpp/gemma4-e2b-it-q4_0
provider = {
  "baseUrl": "http://127.0.0.1/setup-api/local-ai/llamacpp/v1",
  "api": "openai-completions",
  "apiKey": "ffff…ffff",
  "models": [{ "id": "gemma4-e2b-it-q4_0", "name": "gemma4-e2b-it-q4_0",
               "reasoning": false, "input": ["text"],
               "cost": {"input":0,"output":0,"cacheRead":0,"cacheWrite":0},
               "contextWindow": 131072, "maxTokens": 131072 }]
}
```

So the migration and the repair compose correctly: the box ends up on the local
model **and** that model is now resolvable.

**What this does not prove.** The fix has not been run on the affected device.
The box used for this investigation became inaccessible before the fix was
written (its owner credential stopped verifying, unrelated to this change), so
deploying #562 and confirming a live chat turn recovers is still outstanding and
should happen before anyone relies on it.

PR #531, opened 2026-08-28
fix: make Jetson Python CLI installs PEP 668-safe

jamesachurchill:
## Summary

First focused implementation step toward #390.

The Jetson installer currently installs the ClawKeep and Hugging Face command-line applications with `pip --user`. Ubuntu 24.04 enforces PEP 668's externally-managed-environment policy and refuses those installs.

This change:

- provisions `pipx` during the normal Jetson dependency step;
- defensively refreshes APT metadata and provisions `pipx` when the ClawKeep or local-model step is dispatched independently;
- installs and refreshes ClawKeep in an isolated pipx environment;
- migrates the Hugging Face CLI to pipx even when an older user-pip `hf` command already exists;
- removes stale pre-pipx command shims before pipx writes its own;
- retains the existing user-scoped pip fallback for offline JetPack 6.2 / Ubuntu 22.04 systems; and
- adds focused installer-contract tests for the PEP 668-safe control flow.

This does **not** claim full JetPack 7 support or change the published support matrix. CUDA, voice, browser, service-lifecycle, and physical-device acceptance remain separate phases of #390.

## Type of change

- [x] Bug fix (non-breaking)
- [ ] New feature (non-breaking)
- [ ] Breaking change
- [ ] Documentation / tooling
- [ ] Other: ___

## How was this tested?

Focused local checks:

- `bash -n install.sh`
- `git diff --check upstream/beta..HEAD`
- independent final review of the amended diff: no actionable findings

Contained aarch64 Ubuntu 24.04 Testbench proof against the exact commit:

- confirmed the host is ARM64 Ubuntu 24.04 with Python's `EXTERNALLY-MANAGED` marker present;
- confirmed an ARM64 `pipx` package candidate is available from the configured Ubuntu repositories;
- passed 17 source/control-flow assertions covering noninteractive APT refresh-before-install, pipx provisioning and idempotence, ClawKeep migration/fallback ordering, existing-`hf` migration, stale-shim removal, and absence of `--break-system-packages`;
- used no credentials, delivery routes, service changes, dependency installation, or retained remote state.

The full project checks were not run locally; fork CI remains the gate:

- [ ] `bun run lint` passes — not run locally
- [ ] `bun run test` passes — focused source-contract proof passed; full Vitest suite not run locally
- [ ] `bun run build` succeeds — not run locally
- [ ] Manually verified on device / dev install — physical JetPack 7 acceptance remains a later #390 phase

## Checklist

- [x] Branch is up to date with the target branch
[...]

PR #569, opened 2026-09-01
fix: make Jetson Python CLI installs PEP 668-safe (rebase of #531)

yalexx:
Rebase of #531 by @jamesachurchill onto current `beta`, plus real-hardware verification.

The original PR could not be merged: it was ~110 commits behind `beta` and `CONFLICTING`, and
it comes from a fork, so the rebase could not be pushed to the contributor's branch. The commit
here is theirs, unchanged — only the base moved. It rebased cleanly with no content conflicts.

## What it does

Installs the ClawKeep and Hugging Face CLIs through `pipx` instead of `pip --user`, so they keep
working under PEP 668 (`externally-managed-environment`, enforced from Ubuntu 24.04 / JetPack 7).
Keeps the `pip --user` path as a fallback for offline JetPack 6.2 / Ubuntu 22.04 hosts, and removes
stale pre-pipx command shims before pipx writes its own.

## Device verification

Flashed a clean Orin Nano 8GB (serial `1791526043297`) from `beta`, then ran this branch on it.

Baseline before the change, on JetPack 6.2 / Ubuntu 22.04 / Python 3.10.12:

```
pipx: NONE
-rwxrwxr-x 1 clawbox clawbox 177 /home/clawbox/.local/bin/clawkeep      # pip --user script
-rwxrwxr-x 1 clawbox clawbox 179 /home/clawbox/.local/bin/clawkeepd
```

This is exactly the "upgrade from a pre-pipx install" case the change handles.

Ran `sudo bash install.sh --step post_update` (the real in-app-update path that reaches
`step_clawkeep_install`):

```
installing clawkeep from spec '/home/clawbox/clawbox/clawkeep'...
done! 
  installed package clawkeep 0.1.0, installed using Python 3.10.12
  ClawKeep CLI installed: usage: clawkeep <command> [args...]
```

After:

```
pipx: /usr/bin/pipx
clawkeep  -> /home/clawbox/.local/pipx/venvs/clawkeep/bin/clawkeep
clawkeepd -> /home/clawbox/.local/pipx/venvs/clawkeep/bin/clawkeepd
$ clawkeep --help
usage: clawkeep <command> [args...]
```

The stale pip scripts were replaced by pipx symlinks, `pipx list` reports `clawkeep 0.1.0`, and
both commands resolve and run from a login shell. `clawbox-gateway` and `clawbox-setup` are active
afterwards (gateway `200` on 18789).

Note this box is Ubuntu 22.04, where PEP 668 does not apply — so this run verifies the change does
not regress the currently shipping platform, and that the pipx migration path works on real
hardware. It does not exercise the 24.04 enforcement itself.

Closes #531.

## Summary by CodeRabbit

* **New Features**
  * Improved installation reliability by preferring isolated `pipx` environments for ClawKeep and Hugging Face CLI tools.
  * Added fallback behavior for offline systems or environments where `pipx` cannot be installed.
  * Automatically removes outdated launchers and verifies installed commands are executable.
  * Supports migrating existing Hugging Face CLI installations to `pipx` when available.

* **Bug Fixes**
  * Improved compatibility with systems enforcing externally managed Python environments.
  * Prevented unsafe package installation methods during setup.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1001,6 +1001,33 @@ wait_for_apt() {
   recover_dpkg
 }
 
+# Best-effort pipx bootstrap. step_apt_update installs pipx as part of the
+# full-install/update sequence, but step_clawkeep_install and (the Hugging
+# Face CLI install inside) step_llamacpp_install can also run standalone via
+# a root-owned standalone step service, outside that sequence — so each calls
+# this first rather than assuming apt_update already ran on this boot.
+#
+# pipx is what keeps these installs working under PEP 668 (the
+# externally-managed-environment policy Ubuntu enforces from 24.04 / JetPack
+# 7 onward); on JetPack 6.2 / Ubuntu 22.04 it is optional polish; pip --user
+# still works there, so callers fall back to it when this returns non-zero.
+ensure_pipx() {
+  as_clawbox_login "command -v pipx" &>/dev/null && return 0
+  wait_for_apt
+  # Refresh package metadata first — a standalone dispatch (see comment
+  # above) may run on a boot where step_apt_update never ran, so the local
+  # apt cache can be stale or empty and "apt-get install pipx" would 404 on
+  # a fresh image. Tolerate failure here (offline JetPack 6.2 hosts still
+  # need to fall through to the pip --user path below) but don't hide it.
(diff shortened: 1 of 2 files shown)


yalexx:
@jamesachurchill  thanks for contributing !

PR #590, opened 2026-09-02
fix: name the provider hermes billed on an unpinned chat turn

KrasimirKralev:
Closes the defect on **TASK-633** and the coverage gap on **TASK-667**.

## The symptom, on hardware

The ordinary new chat. Press "+", type, send — the request names neither a model
nor a provider, because the client has no per-provider pick to send yet. The
reply comes back naming the model and unable to say who served it:

```
POST /setup-api/hermes/chat  {"message":"which model are you","sessionKey":"desktop-…"}
done → {"model":"gpt-5.6-sol","sessionId":"20260902_205829_99ac3f","harness":"hermes", …}
        keys: ['harness','model','reasoning','sessionId','text']      ← no "provider"
transcript[1]  model='gpt-5.6-sol'  provider=None
```

Three independent runs, three fresh sessions, three times the same. Not a cold
box and not a resumed turn — the catalogue was live and warm
(`source:'dashboard'`, `stale:false`), and none of the three sent a `sessionId`.
This is the default path, and on it the bubble shows a model with nothing beside
it while the tools tell the agent the label is there.

**The harness knew the answer and was not asked.** For the same three session
ids, Hermes' own store held it exactly:

```
session_model_usage
  20260902_205829_99ac3f  gpt-5.6-sol  billing_provider='openai-codex'  subscription_included
  20260902_205959_a3457b  gpt-5.6-sol  billing_provider='openai-codex'  subscription_included
  20260902_210003_9d1a74  gpt-5.6-sol  billing_provider='openai-codex'  subscription_included
```

## The cause

These turns take the dashboard transport. `servedProviderSlug` returns `""`
whenever the request named no provider — `if (!reported || !onRequestedModel ||
!requested) return ""` — and `servedPair` then records the frame whole, model
without provider, deliberately: *a turn that reported a model owns both halves of
the record, including the half it left empty.*

**That refusal is correct and this PR does not touch it.** #581 exists because
the alternative was worse: two independent `||`s put the session's provider back
beside a model it never served and wrote that invented pairing into the
customer's durable transcript. The guard stays exactly as it is.

What was missing is a **source**, not a looser guard. The turn refuses to invent
a provider; nobody was asking the one party that actually knows.

The gap survived a review that was looking for it because the served-model tests
all sit under `describe("… what the CLI path records …")`. The CLI path fills
both halves from `config.yaml`; the dashboard transport — which is what answers
on a Hermes box — fills neither. That is TASK-667, and it is why the tests in
this PR are written against the dashboard transport.

## Harness first — what Hermes natively provides, and what it does not

`session_model_usage(session_id, model, billing_provider, billing_base_url,
billing_mode, task, api_call_count, …, first_seen, last_seen)` in
`~/.hermes/state.db` is the harness's own record of which provider billed which
model in which session. It is read **read-only**, keyed on the session id the
turn returned **and** the model the frame named.

The other candidates were checked and each fails on a citable ground:

| Native surface | Why it cannot answer this |
|---|---|
| `messages` table | `PRAGMA table_info(messages)` → 23 columns, none naming a model, provider or billing mode. Asked on the box; the answer now sits in the code where the question used to be. |
| `sessions` table | Carries `model` and `billing_provider`, but only the **last** ones the thread used. Reading them per bubble relabels every older reply in a conversation that switched — the exact defect this field exists to prevent. |
| `session.usage` gateway **RPC** | Exists by name, but its params and result payload are undocumented in every public source. Adopting it means guessing a wire shape. |
| `GET /api/analytics/usage?days=N` | Aggregated by day and by model over N days, not addressable by session id, and carries no billing provider. |
| `GET /api/sessions/{id}` | Serves the `sessions` row — the thread's last model. Same limitation as above. |
| `/usage` via `slash.exec` | Human-formatted prose, no `--json`, and it names the *active* provider: the present-tense value that is precisely what is ambiguous here. |
| `hermes -z --usage-file f.json` | Writes exactly the right fields, but upstream states it has no effect outside `-z`, and `-z` starts a new session per invocation and ignores `--resume`. Unusable for a dashboard turn. |
| `hermes sessions list` | Workspace-scoped by cwd; answers "No sessions found" for this route (already recorded in `transcript-store.ts`). Its columns carry no provider. |

**So the finding is: there is no non-SQLite native surface for this, and reading
the store directly is the available mechanism.** That is not a novel liberty
here — `hermes-turn-record.ts` already reads `state.db` read-only for the same
turn's reasoning and tool calls, by the same guarded `node:sqlite` idiom, and the
established third-party Hermes usage tool reads the same file the same way. This
PR adds one more query to that existing reader rather than a second door.

**One candidate is still open, and it is named in the code.** The transport's own
comment says a `session.usage` **EVENT** already arrives on the socket this turn
is running on, and the frame loop's `default` case drops it. Its payload has
never been captured — nothing in this repo has ever looked. If it carries the
provider it beats this read on every axis (per turn, pushed, no second store, no
ambiguity to refuse) and retires `billedProviderFor` outright. That capture is
one read-only box run and it belongs to the next lane; the pointer is in the
function's docstring so it cannot be lost.

Worth stating plainly: `session_model_usage` is **not** publicly schema-
documented (unlike `sessions`). The module's degrade-to-nothing contract is what
makes that acceptable — a renamed table or column costs the label and nothing
else.

## What the fix does *not* do

TASK-633's originally proposed fix — mapping a `custom`/null provider back to
`config.yaml` — **is wrong and is not what this is.** What the box is configured
to run is not evidence about what a session ran; #581 removed exactly that
reading and it stays removed. Nothing here reads `config.yaml`, the device
default, or the catalogue's `current` pair.

## Bound to the turn that just ran

A usage row **aggregates** a session's calls for its key, so `last_seen` says
when that combination last billed, never which turn it billed for — and
comparing it against a clock this process holds would mean knowing its epoch and
units, which nothing here does. An unbounded `WHERE session_id AND model` would
therefore answer with any row that session ever wrote, including an earlier
turn's: a conversation that moved to another provider serving the same model id
would get one bubble confidently naming the wrong one, written durably. That is
the #581 defect class re-sourced, and it is not acceptable.

So the rows are compared against **themselves**. The session's billing rows are
marked (`model | provider | api_call_count | last_seen`) *before* the turn runs;
afterwards, only a row not carrying one of those marks can speak for it. No
absolute time, no unit, no epoch — only "did this change". It also turns the
open timing question below from a wrong answer into a blank.

## The rules it refuses on

`billedProviderFor` answers nothing rather than guessing, in six cases — each one
mutation-tested, i.e. deleting it makes a named test fail:

- **the request named a provider.** Then a blank came from the contradiction
  guard doing its job (a canonical slug asked of a session the dashboard reports
  as the kind `custom`), and the billing record must not talk over it. Gated on
  `wantsProvider`, so `auto` — ClawBox's word for "let Hermes decide" — counts as
  *unpinned*, which is right: the record is precisely the answer to what Hermes
  decided.
- **no usage row** for that (session, model). Unchanged behaviour.
- **no row this turn wrote** — the baseline above.
- **two providers** billed the same model during this turn.
- **the recorded value is the literal `custom`**, which is a kind as well as a
  real slug; the same refusal `servedProviderSlug` makes about the same word.
- **the catalogue says that provider cannot serve that model** — but only where
  the catalogue is an *enumeration*. `normalizeRow` seeds the box's own provider
  and the on-device one from a single configured id, so vetoing the harness's own
  record against a list this repo padded would blank the label on the commonest
  provider on the fleet. Only `isUserDefined === false` says "Hermes ships this
  one and listed its models".

The catalogue is read **from cache only** (`cachedModelOptions`, new export).
This runs after the answer has streamed, and the veto is waived anyway whenever
the catalogue is unavailable or stale, so awaiting a live fetch would hold the
`done` frame — and the durable transcript write with it — for a check that a
failed fetch waives. The test file makes `getModelOptions` throw, so the settle
path fetching again is a test failure rather than a slow turn.

## RED → GREEN

RED, on unmodified `origin/beta` `e8ba1e1a`:

```
 FAIL  src/tests/routes/hermes/chat-served-model-dashboard.test.ts >
       what the DASHBOARD transport records as the served provider >
       names the provider Hermes itself billed, when the frame gave a model and the request pinned nothing
AssertionError: expected { text: 'I am a model.', …(3) } to match object { model: 'gpt-5.6-sol', …(1) }

- Expected
+ Received

  {
    "model": "gpt-5.6-sol",
-   "provider": "openai-codex",
  }

 Test Files  1 failed (1)
      Tests  1 failed | 3 passed (4)
```

The cases that assert an *absence* pass on beta by construction — beta shows no
provider at all — so each is mutation-tested instead. Deleting any one of the six
refusals above fails a named case; deleting the baseline filter fails two.

GREEN: **638 test files / 9252 tests**, one pre-existing flake
(`chat-spoken-reply.test.tsx`, a `waitFor` timeout under full-suite load — 18/18
three times in isolation, a different case each full run, and this branch touches
none of the modules it imports). `npx tsc --noEmit` → exit 0. `eslint` on the
changed files → clean. `bun run check:mcp-tools` → Tool contract OK.

## Sibling call sites

- **Both settle sites** in `hermes/chat/route.ts` go through `servedPair`, so both
  are covered by one change: the `done` path and the quiet-stream recovery
  (which passes `null` and takes the session's settled pair — the same defect,
  the same answer, and the same baseline).
- **The transport's two sites** in `hermes-dashboard-turn.ts` — the session site
  and the completion site — are deliberately untouched. They keep refusing to
  invent; `DASHBOARD_PROVIDER_KIND` is now exported so the route makes the same
  refusal about the same word from one definition instead of two.
- **The CLI path** (`cliServedPair`) is deliberately unchanged, and the docstring
  now says why — for both of its blanking cases, not just the resumed one.
- **Every test that `vi.mock`s `@/lib/harness/hermes-turn-record`** carries both
  new exports. Two of the three failed loudly when they did not; the third passed
  only because its cases never reach the branch, and was completed anyway, since
  a partial mock of a module the route imports fails as a missing-export error
  rather than as the assertion it was written for.
- `readHermesTurn`'s sqlite open moved into a shared `withStateDb` helper so the
  new readers do not duplicate the guard. Its behaviour and its journal text are
  unchanged (`what = "the agent's record"` reproduces the old string byte for
  byte), and its existing tests — including the degradation cases — pass
  untouched.

## Stale claims corrected

`settleTurn`'s "UNVERIFIED … one read-only command settles it" note asked
precisely the question the device lane then ran; the answer replaces the
question. `servedPair`'s headline ("taken WHOLE, never half from one source and
half from the other") and `settleTurn`'s lead ("from the transport rather than
from the agent's own record") both stated a rule this change narrows, and both
now state the rule that actually holds.

## Not proven here

- **The device leg was not run** — the brief for this fix ruled out touching a
  box. Two things therefore rest on the earlier lane's capture rather than on a
  run of this code: that Hermes has committed the usage row by the time
  `message.complete` arrives (if it has not, the first turn of a new chat finds
  no row and the label stays blank — the change is inert there, never wrong), and
  that `payload.model` and `session_model_usage.model` are the same string. The
  second is evidenced: the lane's capture shows `done → model: 'gpt-5.6-sol'`
  beside `session_model_usage … gpt-5.6-sol` for all three sessions, and
  `gpt-5.6-terra` / `claude-sonnet-5` matching their bubbles in the switch test.
  The first is not, and it is one read-only command on a box — run it beside a
  live turn and it settles both, and the `session.usage` frame capture above at
  the same time.
- **The browser.** Every assertion here is at the route, transcript-store and
  harness-store level; whether the rendered bubble *displays* the provider it now
  carries is owed a browser pass, as it was for #581.

## Summary by CodeRabbit

* **Bug Fixes**
  * Dashboard chat now shows the provider that billed a turn when no provider was explicitly selected.
  * Provider and model labels remain accurate for automatic routing, custom providers, and confirmed catalogue pairings.
  * Billing labels are omitted when provider information is missing, ambiguous, invalid, or unrelated to the served model.
  * Corrected provider labels are also saved to the conversation transcript.

### DIFF
--- a/src/app/setup-api/hermes/chat/route.ts
+++ b/src/app/setup-api/hermes/chat/route.ts
@@ -41,14 +41,19 @@ import { DESKTOP_TRANSCRIPT_KEY, transcriptKeyIsSafe } from "@/lib/harness/trans
 import { resolveInMediaRoot } from "@/lib/harness/media-root";
 import { mediaUrl, splitAssistantMedia } from "@/lib/chat-media";
 import { extractReasoningPanels, stripAgentStatusFrames } from "@/lib/hermes-reasoning-panel";
-import { readHermesTurn } from "@/lib/harness/hermes-turn-record";
+import { readHermesBillingProvider, readHermesTurn } from "@/lib/harness/hermes-turn-record";
 import {
   adoptHermesGeneratedImages,
   reclaimImageMentions,
   servableMediaRoot,
 } from "@/lib/harness/hermes-generated-media";
 import { capabilitiesFor, UNKNOWN_FACTS } from "@/lib/harness/capabilities";
-import { isQuietStreamError, openDashboardTurn, type DashboardTurn } from "@/lib/hermes-dashboard-turn";
+import {
+  DASHBOARD_PROVIDER_KIND,
+  isQuietStreamError,
+  openDashboardTurn,
+  type DashboardTurn,
+} from "@/lib/hermes-dashboard-turn";
 import {
   errorFromStderr,
   hermesFailureMessage,
@@ -410,16 +415,18 @@ async function settleTurn(
     // moment the turn was submitted, after any switch had been applied and
     // acknowledged.
     //
-    // UNVERIFIED, and worth stating as such: this rests on `state.db`'s
-    // `messages` table having no model column, which nothing here has ever
-    // asked. `readHermesTurn` selects an explicit column list
-    // (`hermes-turn-record.ts`), so the code neither proves nor disproves it.
-    // If that table — or `sessions` — does carry the model, the harness knows
-    // the answer natively and this reconstruction should be replaced by reading
-    // the row `readHermesTurn` already opens for reasoning and tool calls,
-    // which would be strictly more accurate and would retire `cliServedPair`
-    // with it. One read-only command settles it:
-    // `sqlite3 ~/.hermes/state.db 'PRAGMA table_info(messages)'`.
+    // ASKED, on the box, and the answer is on the record now.
+    // `PRAGMA table_info(messages)` lists 23 columns and not one of them names
+    // a model, a provider or a billing mode — so the row `readHermesTurn`
+    // already opens genuinely cannot supply this, and the reconstruction
+    // stands. `sessions` DOES carry `model` and `billing_provider`, but only
+    // the LAST ones the thread used: reading them per bubble would relabel
+    // every older reply in a conversation that switched, which is the defect
+    // this field exists to prevent.
+    //
+    // What the harness can answer per (session, model) is
+    // `session_model_usage.billing_provider` — see `billedProviderFor`, which
+    // asks it for the one case nothing else can speak for.
     ...(served.model ? { model: served.model } : {}),
     ...(served.provider ? { provider: served.provider } : {}),
   }, sessionKey);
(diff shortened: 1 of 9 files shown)


KrasimirKralev:
Addressing the two flags on the automated pass, and recording the one CI failure.

**🟡 "large PR (1000 lines changed) — consider splitting".** Taken, and it does not
split cleanly here. 593 of those lines are tests, and they are the other half of
the work rather than padding: TASK-667 is the finding that the served-model tests
covered only the CLI path while the dashboard transport — the one that actually
answers on this hardware — was untested for the unpinned turn. Landing the fix
without them would recreate exactly the gap that let this defect survive a review
looking for it. Of the source lines, most are the comment blocks that were already
there and are now wrong (`servedPair`'s headline, `settleTurn`'s lead and its
UNVERIFIED note, `cliServedPair`'s justification) — rewriting them is not optional
when the rule they state is the rule being narrowed. What is left is one new reader,
one exported constant, one cached-only catalogue accessor, and six refusals, and
each refusal is only meaningful beside the read it guards.

**ℹ️ "opens `~/.hermes/state.db` read-only twice per unpinned dashboard turn".**
Correct, deliberate, and worth stating precisely so the cost is on the record:

- Only on the **unpinned** turn. A request that named a provider takes neither read
  (`requestNamedProvider ? null : …`), so the pinned path is unchanged.
- The route already opens that store once per turn regardless — `readHermesTurn` in
  `settleTurn` — so the marginal cost is one extra open plus a single-row query on a
  six-column primary key, both read-only.
- The pre-turn read is what makes the answer honest rather than merely present. A
  usage row aggregates a session's calls for its key, so `last_seen` cannot say
  which turn it billed for; without a baseline an earlier turn's row would label
  this bubble whenever a conversation moved to another provider serving the same
  model id. That is a wrong provider in a durable transcript, which is the defect
  class this whole area exists to remove. Comparing the rows against themselves
  needs no clock and no epoch — and it also turns "Hermes has not written the row
  yet" into a blank instead of a stale answer.
- If the `session.usage` frame the transport already receives turns out to carry the
  provider, both reads go away with the function. The pointer is in the docstring.

**CI: the `test` job failed on `coding-agent.test.ts` and it is not this branch.**
The failure is `ENOTEMPTY: directory not empty, rmdir '/tmp/coding-agent-…/site/.git'`
— a filesystem race in that suite's own temp-directory teardown, not an assertion.
That file passes 83/83 three times in a row locally, this branch changes no file
under `coding-agent`, and the whole suite is 9252/9253 here with the one unrelated
`chat-spoken-reply` `waitFor` flake. Re-running the job.

PR #598, opened 2026-09-03
fix: the Skills app name was a literal, and /app/<id> kept a second English name table

KrasimirKralev:
Finding **F-34** (board TASK-636, under TASK-604).

## Symptom

On a box in any language but English, one desktop app kept its English name: the
Hermes **Skills** store. The icon label, the shelf tooltip, the launcher entry
and the window title all said "Skills" while every neighbour translated around
them. The same window opened with "Open in new tab" was worse — its whole title
bar was English for *every* app, so a Bulgarian owner saw "Hermes Skills" above
a Bulgarian store ("Филтри", "Няма инсталирани умения").

## Cause

A desktop app's `name` is a **translation key**, not copy: `resolveAppName` is
`t(app.name) || app.name`, and `t()` returns the key itself when it is missing.
`{ id: "hermes-skills", name: "Skills", … }` was a literal, so it resolved to
itself in all ten locales — silently, because the fallback is designed to never
throw. The setup wizard's window (`name: "Setup"`, appended in `getAllApps()`)
had the same defect.

The second call site: `/app/<id>` kept its own `APP_TITLES` map of English
names. That is why the standalone window never translated any app's title.

## The mechanism this uses

Nothing new. The repo's own i18n harness — `useT()`/`t()` over
`mergeTranslations(setupTranslations, desktopTranslations, …)` keyed `app.*` —
already does this for the other thirteen apps; the fix just uses it.

Considered and rejected: reusing `skills.title`, which already exists translated
in all ten locales. It means "**Hermes** Skills" (bg "Умения на Hermes"), so it
would rename the English tile and put a Hermes-only string under a 110px icon
cell; it also lives in `hermesTranslations`, and the desktop registry is
edition-neutral. The new `app.skills`/`app.setup` values are consistent with the
existing `skills.title` wording in every locale (bg "Умения" vs "Умения на
Hermes", fr "Compétences" vs "Compétences Hermes", de/nl/sv keep "Skills").

## The fix

- `app.skills` and `app.setup` in all ten locales.
- The registry entry and the `getAllApps()` setup entry name those keys.
- `src/lib/desktop-apps.ts`: the app registry moved out of `src/app/page.tsx` so
  the standalone window derives `APP_TITLES` from it instead of duplicating it.
  A drift is now a compile error rather than a convention.
- The desktop context menu's last hardcoded `Hermes Skills` now reads `t("app.skills")`.
- `/app/<id>` resolves the key only for built-ins — an installed app's name is
  catalogue copy and an unknown id is the raw URL segment, so neither goes
  through `t()` (a webapp registered as `search` would have been titled with the
  "Search..." button label).

## RED → GREEN

RED, on unmodified beta `00aea577`, `src/tests/components/standalone-app-title-i18n.test.tsx`:

```
 FAIL  |components| standalone-app-title-i18n.test.tsx > /app/<id> title bar > names the app in the owner's language
TestingLibraryElementError: Unable to find an element with the text: Умения.
        <span class="text-xs font-medium text-white/70">
          Hermes Skills
        </span>
        …
                  <p class="text-xs text-[var(--text-secondary)]">
                    Все още няма по какво да се филтрира.
```

RED, on beta `bc39e24f`, the unit test in its pre-refactor form (it scraped both
tables out of source; after the refactor it imports the shared registry, so this
exact form no longer applies):

```
 FAIL  desktop-app-names-i18n.test.ts > gives every app a name key that exists in every locale
AssertionError: app "hermes-skills" should name a translation key, not copy: expected 'Skills' to match /^app\./

 FAIL  desktop-app-names-i18n.test.ts > titles the standalone window with the same name the desktop shows
AssertionError: /app/settings should title itself with the desktop's name for "settings": expected 'Settings' to be 'app.settings'
```

GREEN: `bun run test:coverage:ci` — see the CI run on this PR. Locally
`tsc --noEmit` and `eslint` are clean.

## Sibling call sites

Handled: `/app/<id>`'s `APP_TITLES`; the desktop context-menu literal
(page.tsx); the `getAllApps()` setup entry; `desktop-default-icons.test.ts`,
which pinned a registry line to `page.tsx` and now reads the new module.

Cleared, with reasons:

- `ChromeShelf`, `ChromeLauncher`, `ChromeWindow` and the launcher's search
  filter all receive an already-resolved name from `resolveAppName` — nothing to change.
- The installed-app grid renders store-catalogue names, which are copy, not keys.
- `mcp/lib/context.ts` `builtInApps()` keeps its own English names and
  descriptions for the agent (`browser` is "Browser Setup" there, not "Browser").
  It is a tool catalogue, not a UI surface, and the agent answers in the owner's
  language itself.
- No e2e spec asserts an app title; `e2e/standalone-builtins.spec.ts` checks only
  "App not found" and "Back to Desktop".

## Two English strings change

The standalone window title for the App Store is now "Store" (the desktop's name
for it) and for the skills store "Skills". Both now say what the desktop window
for the same app says.

## Known, not fixed here

`I18nProvider` loads its catalogue in an effect, so any surface using `t()`
paints the raw key for the first frame and keeps painting it if that chunk load
fails. The standalone title bar now shares that behaviour with every desktop
icon label. Seeding the provider with a static English pack is an `i18n.tsx`
change affecting every surface and belongs in its own PR.

## Summary by CodeRabbit

* **New Features**
  * Added localized names for the Skills and Setup apps across supported languages.
  * Standardized desktop app titles so built-in app names appear in the selected language, including standalone app windows.

* **Bug Fixes**
  * Corrected standalone app title display to show translated names instead of English labels.

* **Tests**
  * Added coverage for localized app names and standalone title rendering.


KrasimirKralev:
CodeRabbit pre-merge warning — Docstring Coverage 50% — refuted, not fixed.

Both named functions this PR adds already carry doc comments: `StandaloneTitle` in `src/app/app/[id]/page.tsx` (why it has to render inside the provider, and why only a built-in name goes through `t()`), and `expectNameIsResolvable` in the new unit test (why only `en` is checked per key). The other two functions the check counted are inline arrow callbacks inside data expressions — `apps.map((a) => [a.id, a.name])` and the test-utils render mock — where a docstring would be noise, not documentation.

The new module and both new test files also open with a comment explaining the defect they exist for, which is where the real explanation belongs.

PR #634, opened 2026-09-04
fix: let an entitled ClawBox AI Max model run instead of vetoing it on the device-tier badge (TASK-691)

KrasimirKralev:
**TASK-691** — owner ruling 2026-09-04: "WE have it DeepSeek Pro. Fix it and make it work." The earlier "hide the Pro id below the Max plan" line is revoked; a Pro turn picked from the Telegram `/model` keyboard, and from the ClawBox chat picker, has to actually run.

## Customer-visible symptom

A ClawBox AI **Max** owner picks the Max model (`deepseek-v4-pro`) — in the Telegram `/model` keyboard, or in the chat header. He then opens the dashboard chat and is told

> Max Tier needs a Max subscription. Upgrade in the ClawBox portal to unlock it — switching you to Pro Tier so chat keeps working.

…and the box is moved off the model he chose: the picker POSTs `deepseek/deepseek-v4-flash` as the new primary. He is being sold the subscription he already pays for, and the Max model he picked is gone.

## Cause

`ChatPopup` decided **entitlement** from the two-value device-tier badge (`clawboxLogin.tier === 'pro'`), in two places: the pick-time refusal, and a boot guard that *writes* — it rewrites `agents.defaults.model.primary` back to Flash.

The badge is a device **default**, not an entitlement. `mapPortalTier()` prefers the portal's `deviceTier`, and the configure route documents that on purpose: "a Max subscriber who deliberately runs Flash on this device" is a supported state (TASK-481). A default is the right thing to *write* as the primary model. It is the wrong thing to *veto* an explicit pick with — and the old condition vetoed on two values that are not refusals:

* **`null` because nobody answered.** Portal unreachable, or a failed `/setup-api/ai-models/status` poll: `useClawboxLogin`'s transient handler leaves `tier: null, loading: false`, and `null !== 'pro'` fired the guard. **One failed poll rewrote a Max box's primary model to Flash.** That is the *false failure* class, with a destructive write behind it, and it is what the RED below reproduces.
* **`"flash"` because the device default is Flash** while the account may run Pro — the case the gate now gets right.

## Harness/portal-first: the mechanism already exists

The portal already publishes the answer, in the very response `fetchPortalTier()` fetches for the badge. Measured read-only on the box (2026-09-04):

```
GET https://clawbox.com/api/clawbox-ai/device-info   →  200
{"tier":"max","deviceTier":"pro",
 "allowedModels":["deepseek-v4-flash","deepseek-v4-pro","deepseek-v4-flash-vision-exp","gpt-5.6-luna","gpt-5-nano"],
 "defaultModel":"deepseek-v4-pro", …}
```

ClawBox was throwing `allowedModels` away and re-deriving entitlement from a badge. No new list, no new probe, no extra request: the fix carries the portal's own list through `fetchPortalTier` (same cache, same round trip), surfaces it as `clawaiAllowedModels` on `/setup-api/ai-models/status`, and gates on it.

## The rule now

`portalDeniesClawboxAiModel(model, allowedModels)` is **false on any doubt**:

* refuses only when the portal named a list and that list excludes the model's bare id;
* an absent / non-array / empty list is "not answered" and never a refusal;
* only ClawBox AI refs (`clawai/…`, `deepseek/…`) are judged — the list says nothing about a provider the owner brought their own key for;
* the boot guard — the one that writes — additionally requires the refused model to be the **Max tier** (the only one with a tier below it to fall back on, which is what its message says) and requires Flash itself to be allowed, so it can never switch to a second refused model;
* it latches only on a switch that happened; a failed one releases the latch instead of leaving the box on a refused model under a message saying it moved.

**Nothing that used to be refused becomes allowed on an older portal.** A portal that ANSWERS but publishes no `allowedModels` is not "unknown" — there the badge is all the entitlement there has ever been, so the status route fills the list from it (`allowedModelsForTier`). Only an unreachable portal yields `null`. This repo's own pre-existing test fixture for a Free device-info body already carries `allowedModels: ["deepseek-v4-flash"]` (`src/tests/routes/ai-models/status.test.ts` on `beta`), i.e. the field is published on unpaid plans too — recorded there before this PR, not measured by me.

A genuinely unentitled pick keeps the upgrade prompt and the drop to Flash — that protection is unchanged and pinned by a test.

## Device evidence (read-only, OpenClaw box, beta `08e7057d`)

* portal: `tier=max`, `deviceTier=pro`, `allowedModels` contains `deepseek-v4-pro` (same answer on the Hermes box).
* `agents.defaults.model.primary` = `deepseek/deepseek-v4-pro`; one turn, no override:
  `executionTrace: {winnerProvider:"deepseek", winnerModel:"deepseek-v4-pro", result:"success", fallbackUsed:false}`.
* **What the proxy does with a model outside the list** — it refuses in the open, it does not silently downgrade. `POST <proxy>/chat/completions {"model":"gpt-5.5"}` with the paired token:
  `400 {"error":{"message":"Model not allowed: gpt-5.5","type":"invalid_request_error","code":"model_not_allowed"}}`, while `deepseek-v4-pro` on the same token answers `200` with `model: "deepseek-v4-pro"`. Three comments in this tree claimed a silent live-tier downgrade; they are corrected to the measured contract.

So the proxy serves Pro for this plan — the model the picker was refusing runs. No mutex taken, nothing on the box changed, no message sent.

## RED → GREEN

`src/tests/components/chat-clawai-pro-entitlement.test.tsx`, on unmodified `beta`:

```
FAIL  a ClawBox AI Pro model the account is entitled to
      > is left running when the portal lists it, whatever the device-tier badge says
AssertionError: expected [ Array(1) ] to deeply equal []
- []
+ [ "{\"model\":\"deepseek/deepseek-v4-flash\"}" ]

FAIL  a ClawBox AI Pro model the account is entitled to
      > is left running when the entitlement could not be read at all
AssertionError: expected [ Array(1) ] to deeply equal []
- []
+ [ "{\"model\":\"deepseek/deepseek-v4-flash\"}" ]

Test Files  1 failed (1)   Tests  2 failed (2)
```

i.e. beta rewrites the box's primary model to Flash in both cases. After the fix, with the third case (a portal that really does refuse the id, which must still downgrade):

```
Test Files  1 passed (1)   Tests  3 passed (3)
components suite:  133 files, 1207 tests passed
unit suite:        567 files, 8958 tests passed
tsc --noEmit: clean   eslint on the changed files: clean (no new warnings)
```

## Sibling call sites

* **`ChatPopup` is the shared chat on BOTH editions** — one component, one fix. On Hermes the model pill (`changeHermesModel`) is a per-provider preference and never carried a tier gate, and `/setup-api/chat/model` is OpenClaw-shaped, so the Hermes Telegram pick was never undone by the picker; the change strictly narrows what can refuse on either edition.
* `src/lib/chat-reasoning.ts:114` — `isClawboxAiProModel` there selects reasoning levels, not entitlement. Left alone.
* `SettingsApp.tsx`, `TierUpgradeCelebration.tsx`, `page.tsx` — read the badge for paid-feature gating and the plan badge, which is what the badge is for. Unchanged.
* `scripts/claude-ds:117-123` picks the coding-agent model from the stored badge, `src/app/setup-api/ai-models/configure/route.ts:1950` writes the primary from it, and `src/lib/hermes-clawai.ts:56,121` writes Hermes' `model.default` from it. All three set a **default**, which is what the badge is for and what TASK-481 decided; none of them vetoes a pick. Left alone deliberately — changing what the device *defaults* to is a product decision, not this defect.
* `src/lib/hermes-tts.ts:478` and `scripts/gateway-pre-start.sh:1454` gate whether cloud VOICE runs on `clawai_tier === "pro"` — that one is an entitlement decision taken from the badge, the same class one surface over. This PR cannot fix it: the portal's `allowedModels` carries chat and vision ids only (measured above), no speech id. Recorded as a gap.

## Review

One Opus `clawbox-review` pass: 4 high, 5 medium, 5 low. Applied: the "older portal answers without a list" compatibility leg (the pass was right that the first cut turned a fail-closed gate into a fail-open one for anyone the portal answers differently), the write-guard re-narrowed to the Max tier, message accuracy, the latch-on-success, one shared normaliser, the corrected proxy comments, and every test finding (constructed fixtures now say so, the negative assertions sit behind the same window the positive control lands in, the hook's parsing and same-array promise are covered, and a sibling fixture that had stopped meaning what its comment said). Refuted with the reason in the thread above: the three "still derives the model from the badge" findings (S2/S3/S4) — those are defaults, not vetoes, and TASK-481 decided them.

## Declared gaps

* **Device leg unproven for the UI change.** The ClawBox web server is down on both dev boxes right now (`clawbox-setup.service` inactive; `clawbox-root-update@rebuild_reboot` failed on the OpenClaw box and activating on the Hermes box — an update someone else is running), so the fixed picker and route were not exercised on hardware. The portal input the fix reads was verified read-only on both boxes. After this merges and deploys: open the chat on a Max box whose primary is `deepseek/deepseek-v4-pro` — no "needs a Max subscription" message, and the model stays Pro.
* **Not fixed here, and it is the other half of the card's title:** the OpenClaw box carries an orphaned `agents.defaults.modelPolicy.allow = ["openai/gpt-5.5"]` that ClawBox never writes, reads or surfaces. Measured today: `--model deepseek/deepseek-v4-pro` (== the primary) serves, while `--model deepseek/deepseek-v4-flash` and `--model anthropic/claude-opus-5` are refused by that list; in the installed core 2026.8.1 `dist/apply-session-model-selection-*.js` refuses a picker selection through the same policy and the Telegram ingress drain is one of its callers. That is TASK-654 D-08/D-09, an owner decision with its own branch.

---

## Finishing round

Two review passes were owed on this branch: the one that BLOCKED it (1 high, 2 medium, 3 low) and a fresh Opus pass over the result, which found two more HIGHs the earlier rounds had missed. Both are applied. One of the mediums is **refuted with evidence** rather than applied — see below.

### 1. The blocker: the guard retried a failed switch, forever

`switchChatModel` is a `useCallback` whose deps include `switchingModel` — state it writes itself — and the boot guard is a `useEffect` that depends on that callback. A switch that FAILED flipped `switchingModel` true→false, gave the callback a new identity, and re-ran the very effect whose `.then(switched => { if (!switched) tierGuardRef.current = false })` had just cleared the latch. Nothing about the box had changed, so the guard fired again. On any repeatable failure of `/setup-api/chat/model` — 400 "Selected AI provider is not configured", 400 "Invalid model identifier", 409, 500, or a proxy answering non-JSON into the same catch — a box whose primary is `deepseek/deepseek-v4-pro` hammered its own setup server and filled its own transcript for as long as the chat window was open.

**It does not reproduce with an ordinary mock, and that is the part worth keeping.** A `fetch` stub that settles inside the same microtask never lets React commit the `switchingModel = true` render (the one that raises the "Switching AI provider…" overlay), so the callback identity never changes and the loop is invisible — the first version of this test passed on the broken head. Holding the POST open until that render commits, the way a real round trip does, gives **251–475 POSTs in a 200 ms window**, measured across four runs.

**Never releasing the latch is the mirror defect, and the second pass caught it.** `src/app/page.tsx:2401` keeps `ChatPopup` mounted for the whole session and only toggles `isOpen` (the component's own comment at `:2843` says so), so a boolean latch that is never released leaves a box on a refused model until a page reload — probe-once, one of the three classes this codebase keeps producing.

So the latch is no longer a boolean: **it remembers the inputs the last attempt was made for** (`{ model, allowed }`). Re-entry on unchanged box state buys nothing — no loop. A changed active model or a changed entitlement list buys exactly one more attempt, and so does closing and re-opening the chat, which is the retry an owner actually reaches for. `allowedModels` keeps its array identity across polls that bring the same ids back (`sameIds` in `use-clawbox-login.ts`), so `===` is a sound "the answer has not changed" test. The guard also waits rather than recording an attempt while another switch is in flight, since `switchChatModel` early-returns there without doing anything.

The **"switched you to Pro Tier" message moved to the success path** with it: a failed switch now leaves only the error, which is true.

### 2. Second HIGH: the compatibility list was built from the device badge and handed to the guard that WRITES

Found by the second pass, in the compatibility leg the *first* pass asked for. `allowedModelsForTier(lookup.tier)` derived the fallback entitlement from `mapPortalTier`'s output — which prefers the portal's `deviceTier` **stamp**, not the plan. On a portal that answers without `allowedModels`, a Max subscriber whose box is stamped `deviceTier: "flash"` (the TASK-481 state this PR exists to respect) got `["deepseek-v4-flash"]`, which the boot guard reads as a positive refusal and writes his primary model down on, under a message telling him to buy the plan he already has. **TASK-691 verbatim, through the compatibility door.**

Root cause: `PortalLookup` collapsed the PLAN (`body.tier`) and the DEVICE DEFAULT (`body.deviceTier`) into one field, so by the time the compat path ran the plan had been thrown away. `PortalLookup` now carries both — `mapPortalPlanTier` is the sibling of `mapPortalTier` that reads the plan and ignores the stamp — and the fallback list reaches the Max id when **either** reading does. The "either" is load-bearing in both directions: the plan alone would start refusing a box the portal stamped `"pro"`, which used to work, and a compatibility path must not narrow what it was added to preserve.

### 3. Refuted, not applied: the click-time decline (`M2`)

The blocking review's second medium asked the click gate to refuse the Max pick while the portal is unreachable and the badge is not Max. I implemented it, the second pass argued it should be dropped, and on checking the code it is right. Three findings from the tree, not opinions:

* **It fires on exactly the state this PR exists to fix.** A Max subscriber whose box is stamped `deviceTier: "flash"` — preserved on purpose by `mapPortalTier` (`clawbox-ai-portal-tier.ts:142-151`) — is declined the Max pick whenever the portal is silent. Same customer, same refusal, same ticket.
* **It does not cover the case its own docstring justified it with.** The harm named was a revoked token writing the Max model into `agents.defaults.model.primary`. A box that was *on* the Max default carries `deviceTier: "pro"`, so the persisted badge is `"pro"`, the condition is false, and the pick goes through and writes the primary anyway.
* **Its third argument is not the fact the doc claimed.** `clawai_tier` is not "the last portal-confirmed badge": `configure/route.ts:1907-1911` resolves `portalConfirmedTier ?? requestedClawboxAiTier ?? stored ?? default` and writes the result at `:2524`/`:2558`, so on any configure where the portal probe failed the stored badge is the **wizard plan picker's** value. The gate is satisfied by anyone who once clicked "Max" offline, and fires on the genuine Max owner.

The blast-radius concern behind it is real — the Max pick does persist into `agents.defaults.model.primary`, which Telegram and the coding-agent wrapper also run. But the remedy has to be *asking the authority*, not inferring from the one field this PR just finished proving is not an entitlement. The repo already has that pattern: `src/lib/clawbox-ai-vision.ts:50-80` sends one `max_tokens: 1` completion and reads `400 model_not_allowed` as a measured refusal while treating auth failures, timeouts and 5xx as "the question failed". That is the right shape and it is a change of its own, not a line in this PR.

### RED → GREEN

The blocker, on the blocked head `9df09a29`:

```
FAIL  src/tests/components/chat-clawai-pro-entitlement.test.tsx
      > POSTs once, says only what happened, and does not re-arm
AssertionError: expected [ …(2) ] to deeply equal [ Array(1) ]
  [
    "{\"model\":\"deepseek/deepseek-v4-flash\"}",
+   "{\"model\":\"deepseek/deepseek-v4-flash\"}",
  ]
```

An earlier form of the same test, asserting only the false-success half, on the same head:

```
AssertionError: expected '…' not to match /switching you to Pro Tier/i
+ Received: "…Max Tier needs a Max subscription. Upgrade in the ClawBox portal to
unlock it — switching you to Pro Tier so chat keeps working.Error: Selected AI
provider is not configured…"
```

The two defects the second pass added, both RED on the intermediate head `0040e3fd` (checked in a throwaway worktree at that commit):

```
FAIL  chat-clawai-pro-entitlement.test.tsx > tries again when the chat is closed and re-opened
AssertionError: expected [ Array(1) ] to have a length of 2 but got 1
- 2
+ 1

FAIL  status.test.ts > keeps the Max id in that fallback when the PLAN is Max and only
                       the device stamp is Flash
AssertionError: expected [ 'deepseek-v4-flash' ] to deeply equal [ 'deepseek-v4-flash', …(1) ]
  [
    "deepseek-v4-flash",
-   "deepseek-v4-pro",
  ]
```

GREEN:

```
npx vitest run src/tests/components/chat-clawai-pro-entitlement.test.tsx \
  src/tests/components/chat-thinking-default.test.tsx \
  src/tests/routes/ai-models/status.test.ts \
  src/tests/unit/clawbox-ai-entitlement.test.ts \
  src/tests/unit/use-clawbox-login.test.ts
→ Test Files 5 passed (5)   Tests 61 passed (61)

npx tsc --noEmit → clean

npx vitest run --project components → Test Files 138 passed (138)   Tests 1291 passed (1291)
npx vitest run --project unit       → Test Files 597 passed (597)   Tests 9576 passed (9576)
```

Both full suites are green on this head, run one at a time. Recorded because the first attempt was not: running the two projects and `tsc` concurrently on one machine produced a scatter of 5-second timeouts (`chat-spoken-reply`, `chat-email-refs-surfaces`, `chat-header-seed-race`, `run-tunnel`, …). They were checked rather than waved through — each passed alone, and `chat-spoken-reply` failed *worse* on the untouched base head `9df09a29` in a throwaway worktree — and then they all passed in the sequential runs above. Contention, not flake, and not this diff.

### Sibling sweep: every effect that depends on a callback and writes through it

Swept all **67** `useEffect` (and zero `useLayoutEffect`) in `ChatPopup.tsx` against every `useCallback`/`useMemo` in the file. **`switchChatModel` is the only one with the full shape** — the only non-stable callback an effect both depends on and calls, where the call writes one of that callback's own deps. It writes *two* of them: `switchingModel`, and `chatModelState` via `setChatModelState` on success, so even a *successful* switch changes the identity and re-runs the effect. Every other callback appearing in an effect dep array bottoms out in `useCallback([])` (`connect`, `refreshChatModelState`, `scrollToBottom`, `failPending`, `dispatchTurn` and its whole dep chain, `startRun`, `endImageWait`, `stopRecording`, `abandonRecording`, `seedHermesHeader`, …); `adapter`, `caps` and `clawboxLogin.allowedModels` are identity-stable by construction.

Six effects write a dep of *themselves* without a callback in between; each was checked rather than assumed, and all are bounded. The nearest miss is the thinking-level effect (`:1602`), which releases its own latch on failure exactly the way this one did and is held only by a `prev === suggested` React bail-out.

**One defect of the same class found while sweeping, deliberately not fixed here:** the `new WebSocket(wsUrl)` catch (`ChatPopup.tsx:2308-2314`) sets `status: 'error'` without `tearDownReloadOverlay()`, unlike the three other terminal paths. The retry effect at `:4666` therefore stays armed and resets `retryCountRef.current = 0` every 3 s, so the reconnect ladder can never exhaust on a malformed `wsUrl`. Separate path, separate defect — **TASK-712**.

### Declared, and not resolvable from this repo

* **The premise is not measured.** The only `device-info` response anyone has seen is `{"tier":"max","deviceTier":"pro","allowedModels":[…]}`, where badge and entitlement agree — so it cannot discriminate between the portal computing `allowedModels` from the PLAN (what this PR needs) and from the DEVICE STAMP. The discriminating measurement is a Max-plan token whose device is stamped `deviceTier: "flash"`, and it needs the portal plus a specially stamped box. **What can be said without it:** if the list turns out to be stamp-derived, this PR is still never worse than `beta` — `beta` vetoed on the badge, which is the same stamp, and additionally vetoed when nothing answered at all. It would then be an incomplete fix rather than a regression, and the remaining half would be a portal change.
* **Device leg unproven.** The ClawBox web server was down on both dev boxes throughout this work (`clawbox-setup.service` inactive, a root update in flight), so the fixed picker and guard were never exercised on hardware. Nothing in the delta touches the updater, gateway, systemd units, device paths or the edition lock — it is browser-side chat logic CI exercises — but the guard's failure path is exactly what a box run surfaces. After merge and deploy: open the chat on a Max box whose primary is `deepseek/deepseek-v4-pro` (no "needs a Max subscription", model stays Pro), then again with the gateway stopped so the write fails (exactly one error line, no repeat), then close and re-open the chat (exactly one more attempt).
* **Edition scope, and how it composes with #626.** The gate covers the OpenClaw chat path. On Hermes the model pill (`changeHermesModel`) is a per-provider localStorage preference that never calls `switchChatModel` and never carried a tier gate, before or after this PR — so nothing is lost there, but nothing is gained either, and `allowedModels` now makes that half cheap. Follow-up, not a regression.

  Rebased onto `538274b3`, which includes **#626** (`/setup-api/chat/model` is now harness-aware). Checked for semantic conflict, not just textual: on Hermes that route answers `409 WRONG_STORE` to **both** verbs, so `refreshChatModelState` bails on `!res.ok`, `chatModelState` stays null, and this boot guard — which reads `chatModelState?.activeModel` — cannot fire on that edition at all. It can no longer write into a store Hermes does not use, which is the safest possible resolution of the edition question above. #626 also made "nothing pinned" a null/null state; the guard's `if (!active || …) return` reads that as "no guard", which is correct. #626's own ChatPopup hunk (`unselectableActive`, ~`:6077`) and its route changes are present and untouched on this head.
* Other follow-ups recorded from the review rounds: **TASK-713** (should a re-pair rewrite the primary from the badge over an explicit pick — the one path where this PR's invariant does not hold, TASK-481's decided behaviour), the `speechEntitledTier` badge-as-entitlement sibling in `hermes-tts.ts:530-539` + `gateway-pre-start.sh` (must move together), and the `allowedModels: []` contract with the portal (today `[]` reads as "not answered" on both sides, which is the safe default but is asserted nowhere the portal can see it).

## Summary by CodeRabbit

- **New Features**
  - ClawBox AI model availability now reflects portal entitlements rather than relying solely on device tier.
  - Users are notified when a selected model is unavailable, with guidance for subscription or portal settings.
  - Unsupported active Pro models automatically switch to Flash when permitted.

- **Bug Fixes**
  - Preserves model access when entitlement information is missing or temporarily unavailable.
  - Correctly handles provider prefixes, capitalization, and invalid entitlement entries.
  - Reports portal-provided model availability and sensible tier-based fallbacks.

### DIFF
--- a/src/app/setup-api/ai-models/status/route.ts
+++ b/src/app/setup-api/ai-models/status/route.ts
@@ -2,6 +2,8 @@ import { NextResponse } from "next/server";
 import { readConfig } from "@/lib/openclaw-config";
 import { get as getConfigValue, set as setConfigValue } from "@/lib/config-store";
 import {
+  CLAWBOX_AI_FLASH_MODEL_ID,
+  CLAWBOX_AI_PRO_MODEL_ID,
   normalizeClawboxAiTier,
   type ClawboxAiTier,
 } from "@/lib/clawbox-ai-models";
@@ -33,6 +35,20 @@ const CLAWBOX_AI_TIER_CONFIG_KEY = "clawai_tier";
 // `applyClawaiToHermes` writes it here; the OpenClaw path reads the same token
 // from `models.providers.deepseek.apiKey` in openclaw.json instead.
 const CLAWBOX_AI_TOKEN_CONFIG_KEY = "clawai_token";
+
+/**
+ * The entitlement list a device-info answer without `allowedModels` implies:
+ * the old badge rule, written out. `"pro"` (the Max device tier) reaches both
+ * chat tiers, anything else reaches Flash — which every paid plan can run.
+ *
+ * Only the compatibility path uses this. A portal that publishes the real list
+ * always wins, and a portal that could not be reached gets no list at all.
+ */
+function allowedModelsForTier(tier: ClawboxAiTier | null): string[] {
+  return tier === "pro"
+    ? [CLAWBOX_AI_FLASH_MODEL_ID, CLAWBOX_AI_PRO_MODEL_ID]
+    : [CLAWBOX_AI_FLASH_MODEL_ID];
+}
 
 function normalizeProvider(provider: string | null): string | null {
   if (!provider) return null;
@@ -181,7 +197,13 @@ async function buildStatusResponse(state: ResolvedAiState): Promise<NextResponse
       const lookup = await fetchPortalTier(state.clawaiToken);
       if (lookup.source === "portal") {
         clawaiAccountTier = lookup.tier;
-        clawaiAllowedModels = lookup.allowedModels;
+        // A portal that answered but published no list is not "unknown": it is
+        // an older portal build, and there the badge IS all the entitlement
+        // there has ever been. Fill the list from it so nothing that used to
+        // be refused silently becomes allowed on such a deployment. Only the
+        // portal-ANSWERED branch does this — an unreachable portal stays null,
+        // which is the whole point of the change.
+        clawaiAllowedModels = lookup.allowedModels ?? allowedModelsForTier(lookup.tier);
         accountTierSource = "portal";
         // Persist the portal-confirmed tier so the portal-unreachable
         // fallback reflects the last *confirmed* tier, not a stale
--- a/src/app/setup-api/ai-models/status/route.ts
+++ b/src/app/setup-api/ai-models/status/route.ts
@@ -37,15 +37,27 @@ const CLAWBOX_AI_TIER_CONFIG_KEY = "clawai_tier";
 const CLAWBOX_AI_TOKEN_CONFIG_KEY = "clawai_token";
 
 /**
- * The entitlement list a device-info answer without `allowedModels` implies:
- * the old badge rule, written out. `"pro"` (the Max device tier) reaches both
- * chat tiers, anything else reaches Flash — which every paid plan can run.
+ * The entitlement list a device-info answer without `allowedModels` implies.
  *
  * Only the compatibility path uses this. A portal that publishes the real list
  * always wins, and a portal that could not be reached gets no list at all.
+ *
+ * It reaches the Max id when EITHER reading of the response does, and that
+ * "either" is load-bearing in both directions. The device stamp alone is what
+ * the old badge rule used, and deriving the list from it reproduced TASK-691
+ * through this very door: a Max subscriber whose box is stamped
+ * `deviceTier: "flash"` — a state `mapPortalTier` preserves on purpose — would
+ * get `["deepseek-v4-flash"]`, which the boot guard reads as a positive refusal
+ * and WRITES his primary model down on. The plan alone would be the mirror
+ * mistake: a box the portal stamped `"pro"` used to be allowed to run the Max
+ * model under the old rule, and a compatibility path must not start refusing
+ * something that used to work. So: the more permissive of the two.
  */
-function allowedModelsForTier(tier: ClawboxAiTier | null): string[] {
-  return tier === "pro"
+function allowedModelsForCompat(
+  deviceTier: ClawboxAiTier | null,
+  planTier: ClawboxAiTier | null,
+): string[] {
+  return deviceTier === "pro" || planTier === "pro"
     ? [CLAWBOX_AI_FLASH_MODEL_ID, CLAWBOX_AI_PRO_MODEL_ID]
     : [CLAWBOX_AI_FLASH_MODEL_ID];
 }
@@ -203,7 +215,8 @@ async function buildStatusResponse(state: ResolvedAiState): Promise<NextResponse
         // be refused silently becomes allowed on such a deployment. Only the
         // portal-ANSWERED branch does this — an unreachable portal stays null,
         // which is the whole point of the change.
-        clawaiAllowedModels = lookup.allowedModels ?? allowedModelsForTier(lookup.tier);
+        clawaiAllowedModels = lookup.allowedModels
+          ?? allowedModelsForCompat(lookup.tier, lookup.planTier);
         accountTierSource = "portal";
         // Persist the portal-confirmed tier so the portal-unreachable
         // fallback reflects the last *confirmed* tier, not a stale
--- a/src/app/setup-api/ai-models/status/route.ts
+++ b/src/app/setup-api/ai-models/status/route.ts
@@ -163,6 +163,13 @@ async function buildStatusResponse(state: ResolvedAiState): Promise<NextResponse
 
   let clawaiAccountTier: ClawboxAiTier | null = null;
   let accountTierSource: "portal" | "picker" = "picker";
+  // The portal's own list of model ids this token may run. NULL MEANS "NOT
+  // ANSWERED" — no token, portal unreachable, or a portal build that does not
+  // publish the field — and no caller may read that as a refusal. It is
+  // deliberately not backed by a local fallback the way the tier badge is: a
+  // remembered entitlement is a guess, and a guess is what locked a Max owner
+  // out of the model he pays for (TASK-691).
+  let clawaiAllowedModels: string[] | null = null;
   if (state.hasClawaiProfile) {
     clawaiAccountTier = localTier;
     // Ask the portal whenever a clawai token is paired, regardless
@@ -174,6 +181,7 @@ async function buildStatusResponse(state: ResolvedAiState): Promise<NextResponse
       const lookup = await fetchPortalTier(state.clawaiToken);
       if (lookup.source === "portal") {
         clawaiAccountTier = lookup.tier;
+        clawaiAllowedModels = lookup.allowedModels;
         accountTierSource = "portal";
         // Persist the portal-confirmed tier so the portal-unreachable
         // fallback reflects the last *confirmed* tier, not a stale
@@ -202,6 +210,9 @@ async function buildStatusResponse(state: ResolvedAiState): Promise<NextResponse
     model: state.model,
     clawaiTier,
     clawaiAccountTier,
+    // Model ids the portal says this account may run, or null when it did not
+    // say. The picker gates on this, never on the tier badge above.
+    clawaiAllowedModels,
     // Whether *any* clawai profile is configured. Distinguishes
     // "no ClawBox AI account at all" (false) from "Free user with
     // a paired clawai token" (true, clawaiAccountTier=null) — the
@@ -230,7 +241,8 @@ export async function GET() {
     return NextResponse.json(
       {
         connected: false, provider: null, providerLabel: null, mode: null, model: null,
-        clawaiTier: null, clawaiAccountTier: null, clawaiConfigured: false, tierSource: "picker",
+        clawaiTier: null, clawaiAccountTier: null, clawaiAllowedModels: null,
+        clawaiConfigured: false, tierSource: "picker",
       },
       {
         headers: {
--- a/src/components/ChatPopup.tsx
+++ b/src/components/ChatPopup.tsx
@@ -273,7 +273,7 @@ import {
 } from '@/lib/hermes-reasoning'
 import { readHermesChatPrefs, writeHermesChatPrefs } from '@/lib/hermes-chat-prefs'
 import { useClawboxLogin } from '@/lib/use-clawbox-login'
-import { clawboxAiMaxPickUnconfirmed, isClawboxAiProModel, portalDeniesClawboxAiModel, CLAWBOX_AI_MODEL_BY_TIER } from '@/lib/clawbox-ai-models'
+import { isClawboxAiProModel, portalDeniesClawboxAiModel, CLAWBOX_AI_MODEL_BY_TIER } from '@/lib/clawbox-ai-models'
 import { PORTAL_DASHBOARD_URL } from '@/lib/max-subscription'
 import { HeaderDropdown } from '@/components/HeaderDropdown'
 import { buildDeviceConnectParams } from '@/lib/gateway-device-identity'
@@ -4177,25 +4177,6 @@ function ChatPopup({ isOpen, onClose, onOpenFull, onOpenSettingsSection, onThink
       }])
       return false
     }
-    // The Max model is also the one pick the portal may simply not have been
-    // asked about — it is unreachable, or it answers 401/403 to a revoked or
-    // migrated token, both of which arrive here as an unanswered list. This
-    // pick persists into `agents.defaults.model.primary`, which Telegram and
-    // the coding agent run too, so declining it while the answer is missing and
-    // the last confirmed badge is not Max costs the owner a retry, where
-    // letting it through costs every turn until the portal answers. The BOOT
-    // GUARD, which writes, keeps failing open on the same doubt — see
-    // `clawboxAiMaxPickUnconfirmed`.
-    if (!clawboxLogin.loading
-      && clawboxAiMaxPickUnconfirmed(target.model, clawboxLogin.allowedModels, clawboxLogin.tier)) {
-      setMessages(prev => [...prev, {
-        role: 'system',
-        text: `${target.label} could not be checked against your ClawBox AI plan — the ClawBox portal did not answer. Try again in a moment, or [check the plan in the portal](${PORTAL_DASHBOARD_URL}). Staying on the current model.`,
-        timestamp: Date.now(),
-        variant: 'error',
-      }])
-      return false
-    }
     setSwitchingModel(true)
     setErrorMsg('')
     try {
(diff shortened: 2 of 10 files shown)


KrasimirKralev:
Portal measurement for the open caveat, read-only on both dev boxes just now: `GET /api/clawbox-ai/device-info` answers `tier=max`, `deviceTier=pro`, `allowedModels=[deepseek-v4-flash, deepseek-v4-pro, deepseek-v4-flash-vision-exp, gpt-5.6-luna, gpt-5-nano]`, `defaultModel=deepseek-v4-pro` (also `meters`, `proReasonerEligible`). Both readings are high on both devices, so the measurement does not discriminate plan-derived from stamp-derived `allowedModels` — the PR body's caveat stands as written. Settling it needs the portal side to state which field drives the list, or a device stamped below its plan. Recorded in the run queue's Found list.

KrasimirKralev:
Rebased onto beta f0d9cafe (was two merges behind, #628 and #636 — no file overlap; range-diff confirms every commit patch-identical). Head is now 68e9bfb8, same content as the reviewed d78fcef6. Final review verdict for that content: MERGE-OK, 0 high / 0 medium / 4 low (code-review-pr634-final3.md) — the round-2 POST-storm blocker measured dead (52 POSTs per 200 ms on the pre-fix head vs 1 here), the compat list reaches the Max id when either the plan or the stamp says Max, and the M2 refutation's citation checks out.

PR #643, opened 2026-09-04
fix: free memory and stop local models before the rebuild

yalexx:
## The problem

An in-app update rebuilds the box through `do_rebuild`, which ran `bun run build` against whatever memory the device happened to have left.

On an 8 GB Jetson that is not a tuning question:

| Engine | Holds memory for |
|---|---|
| `ollama.service` | a resident model, ten idle minutes after the last chat turn |
| `kokoro-server.service` | its voice on the GPU, five idle minutes |
| `whisper-server.service` | same |
| llama.cpp (`llama-server`) | until something stops it — ClawBox supervises it by pidfile |

So an owner who pressed **Update** a minute after talking to their box was building underneath several gigabytes of resident model. The build is OOM-killed and the update ends on a half-written `.next`.

## The fix

`free_memory_for_build` stops those four engines and drops the page cache, called from `do_rebuild` **after** `clawbox-setup.service` is stopped and before `bun run build`.

That order is the whole mechanism, not tidiness: the gateway reaches ollama and llama.cpp through the web server's own proxy (`src/lib/local-ai-runtime.ts`), so with the web server down nothing can pull a model back in behind it for the length of the build. A free performed earlier — while the proxy is still listening — can be undone by the next chat turn, and the build runs for minutes.

Because `do_rebuild` is the single chokepoint, this covers the in-app update (`updater.ts` → the `rebuild_reboot` root step → `step_rebuild_reboot`), `install.sh --step rebuild`, and a full install.

### Properties it holds to

- **Stop, never disable** — the same rule the idle standby in `local-ai-runtime.ts` follows. Every engine here is meant to come back on demand; an update that quietly un-enabled one would be a box that stopped talking after its next reboot, which is a worse bug than the OOM this prevents.
- **It cannot fail the update.** `install.sh` runs under `set -euo pipefail` and this is called as a bare statement, so every command that can legitimately fail on a healthy box carries its own fallback. The engines are not what the owner pressed Update for.
- **It never signals a pid it has not identified.** The llama.cpp pidfile outlives a crash and pids are recycled, and this runs as root — so the pid is checked against `/proc/<pid>/cmdline` first. The TypeScript sibling (`stopLlamaCppServer`) can go straight from file to signal because it usually still holds the child handle; from bash there is nothing but the file.
- **The log does not lie.** A process that has exited and not yet been reaped still answers `kill -0`, so the state letter is read out of `/proc` before claiming it refused SIGTERM. The three-second grace is the shape of that sibling's 1.5 s.

## What the owner sees

Nothing new — and nothing lost. `do_rebuild`'s first line has always been `systemctl stop clawbox-setup.service`, so the UI is already blind for the whole build: `SystemUpdateApp` polls `/setup-api/update/status`, marks the server down after three failures, holds the last step on screen, and reloads once the box answers again. This work runs inside that existing window and adds a few seconds to it. Its output goes where the rest of the rebuild's does — `journalctl -u clawbox-root-update@rebuild_reboot`:

```
Freeing memory for the build (2140 MB available)...
  Stopping ollama.service...
  Stopping kokoro-server.service...
  Stopping llama.cpp server (pid 12345)...
  Memory available for the build: 6180 MB (was 2140 MB)
```

## Tests

`src/tests/unit/install-free-memory-before-build.test.ts` — 15 tests, 3.8 s.

Text assertions pin the ordering (after the stop, before the build; reached by both `step_rebuild` and `step_rebuild_reboot`) and the stop-never-disable rule. The rest drives the real shell functions against stubbed `systemctl`/`sudo`/`sync` on `PATH`, because the two properties that matter most cannot be read out of a regex:

- a pidfile naming a **`llama-server`** → the process is stopped and the pidfile removed;
- a pidfile naming **anything else** → the process is untouched;
- a box where every stop refuses → still exits 0 and reports the memory.

Neighbouring install.sh suites re-run green (166 tests).

https://claude.ai/code/session_01EzUHLCD8Wkz9mYcKbGkKZz

## Summary by CodeRabbit

* **Bug Fixes**
  * Rebuilds now proactively free memory before starting, helping reduce failures caused by insufficient available memory.
  * Cleanup safely handles unavailable services, failed stop attempts, stale process identifiers, and cache-clearing errors without aborting updates.
  * Process validation prevents unrelated or recycled processes from being stopped.

* **Tests**
  * Added coverage for memory cleanup, service handling, process validation, update rebuilds, and failure scenarios.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1107,10 +1107,135 @@ ensure_node_pty() {
   echo "  node-pty rebuilt and verified"
 }
 
-# Stop the setup service, clear cache, reinstall, and rebuild
+# ── Memory for the build ─────────────────────────────────────────────────────
+
+# MemAvailable in MiB, or 0 when /proc cannot be read.
+#
+# Zero rather than a guess, on the same reasoning as the memory guard in
+# scripts/openclaw/clawbox-tts.sh: on a Jetson an unreadable /proc means
+# something is wrong, and assuming "plenty" is how the OOM killer gets invited
+# in. Here it only makes the log line honest — nothing below is conditional on
+# the number.
+available_mb() {
+  local kb
+  kb=$(awk '/^MemAvailable:/ {print $2; exit}' /proc/meminfo 2>/dev/null)
+  [ -n "$kb" ] || { printf '0'; return 0; }
+  printf '%s' "$((kb / 1024))"
+}
+
+# Print the pid of the llama.cpp server named by ClawBox's own pidfile, and
+# nothing at all unless that process is really it.
+#
+# The pid is checked against /proc/<pid>/cmdline before the caller signals it:
+# the pidfile outlives a crash, pids are recycled, and this runs as root — so
+# an unverified kill from a stale file is a root SIGKILL aimed at whatever
+# inherited the number. The TypeScript sibling (stopLlamaCppServer in
+# src/instrumentation-node.ts) can go straight from the file to the signal
+# because it usually still holds the child handle; from bash there is nothing
+# but the file.
+llamacpp_pid_if_running() {
+  local pidfile="$PROJECT_DIR/data/llamacpp/server.pid" pid
+  [ -f "$pidfile" ] || return 1
+  read -r pid < "$pidfile" 2>/dev/null || return 1
+  [[ "$pid" =~ ^[0-9]+$ ]] || return 1
+  grep -qa 'llama-server' "/proc/$pid/cmdline" 2>/dev/null || return 1
+  printf '%s' "$pid"
+}
+
+# Give `bun run build` the board to itself.
+#
+# The rebuild is the most memory-hungry thing this appliance ever does, and on
+# an 8 GB Jetson it shares that memory with whatever the box was doing a minute
+# earlier: ollama keeps a model resident for ten idle minutes after the last
+# chat turn, Kokoro holds its voice on the GPU for five, and a llama.cpp server
+# stays up until something stops it. A `next build` starting underneath 4 GB of
+# resident model does not build slowly — it is OOM-killed, and the update ends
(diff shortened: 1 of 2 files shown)


yalexx:
Both findings addressed in e7f4414d.

**`sync` under errexit — fixed as suggested.** This was a real defect and it contradicted the one property the function advertises. `free_memory_for_build` is called as a bare statement from `do_rebuild`, so a non-zero `sync` would have aborted the whole update before the build. It now warns like every other step here.

**Signalling a bare pid after the identity check — fixed, but not as proposed.** The race is real: three seconds is long enough for the pid to be freed and reused, and this runs as root. But "keep a stable process handle from verification through both signals" isn't available to us — a handle that survives the wait means a pidfd, and bash can neither open one nor signal through it. `llama-server` is also not a child of `install.sh` (the web server spawns it), so there is no `wait`-able job either.

What it does instead is re-establish the identity immediately before the escalation, with the same check that established it:

```bash
if [ "$(llamacpp_pid_if_running || true)" = "$pid" ]; then
```

That narrows the window from the whole three-second grace to the gap between two lines. It does not close it, and the comment in the code says so rather than implying otherwise. The SIGTERM's own window was already just the two lines.

It also subsumes the zombie check it replaces: a process that has exited and not yet been reaped answers `kill -0` but has an empty `cmdline`, so it cannot pass this check — which means the journal no longer claims a process that went quietly "did not exit on SIGTERM".

Both are pinned by tests that fail against the unfixed code (verified by reverting each fix in turn): a failing `sync` must not abort the run, and a server that ignores SIGTERM is the only path that reaches the escalation at all.

On the shellcheck SC2329 note ("this function is never invoked") — it is, from `do_rebuild`, a few lines below; the checker isn't seeing the call site.

PR #648, opened 2026-09-04
Four fixes: local engines guaranteed, plus two UI cleanups

yalexx:
Four independent fixes on one branch.

---

## 1. Every box gets faster-whisper (`c1463907`)

**faster-whisper was unreachable code on every box we have ever shipped.** `install.sh:3179` is the only caller of `install-voice.sh` and always passes `--tts-only`, whose arm exits *above* the STT steps at `scripts/install-voice.sh:799-869`. "Whisper: Not installed" was the fleet's permanent state, with no route to fixing it short of SSH.

### The hour that wasn't

The stated reason for excluding it was cost — *"roughly an hour on an Orin"* (`install-voice.sh:19`, `:693`, `install.sh:3166`). **That number was never measured** (traced to commit `25c4dbac`, no timing recorded), and it was the cost of letting `nvcc` emit code for every CUDA architecture it knows.

Pinned to the architecture the board actually has, measured on this Orin Nano:

```
clone + submodules:   34s
cmake:                 1s
make -j4:            255s
────────────────────────────
TOTAL:               290s   (4m 50s)
```

Verified a real CUDA build, not a silent fallback: `libctranslate2.so.4.8.2` links `libcudnn.so.9` / `libcublas.so.12` / `libcublasLt.so.12`, and `cuobjdump --list-elf` reports **`sm_87` and nothing else**.

`cuda_arch_pin()` reads the architecture off `/proc/device-tree/compatible` (tegra234 → 87, tegra194 → 72, tegra186 → 62) rather than from a GPU query — Jetson has no `nvidia-smi`, and torch is not importable this early on a fresh flash.

### Placement, and why

`install_whisper_stt` runs from the `--tts-only` arm **after** Kokoro, so a failed pip or a lost network can never cost the box its voice — the exact hazard that kept STT out of the shipped path — and **before** the case that exits, because STT is independent of TTS and a box whose Kokoro is missing should still get its ears. Its status is reported and deliberately does **not** feed the exit code: that contract is Kokoro's, and `install.sh` grades the step by it.

### Two traps this avoids

- **The unit file is written LAST**, only once the import verifies. `src/lib/local-models.ts:451` derives `installed` from the unit's *presence alone*, while `src/lib/stt-local.ts:120` demands a real `import faster_whisper`. A unit written ahead of a failed install makes those two disagree, and Settings advertises an engine the box does not have. Pinned by tests.
- **Stamp-gated**, for the reason `install-voice.sh:60-73` already documents for Kokoro: the wheels land before the CUDA library and before the weights, so gating on the import alone would latch a half-done box in as ready. An installed box pays one import check per update.

The CUDA build is non-fatal (the prebuilt CPU wheel pip already fetches transcribes, and both runtime scripts fall back to `cpu/int8`), and `make -j4` rather than `-j$(nproc)` — six parallel `nvcc` jobs on a 7.4 GB board is how this gets OOM-killed halfway through.

---

## 2. The embedding model is repaired on updates (`6547a923`)

`qwen3-embedding:0.6b` was **fresh-install-only**. `step_ollama_install` pulls it when a box is built and `gateway-pre-start.sh` re-checks it detached at every gateway start, but **no update path ran it** — and the boot check only helps if it happens to find ollama awake.

Two ways a box lost semantic memory permanently: it was offline when flashed, or the helper's single 5-second curl landed while ollama was still binding its port after `systemctl restart ollama`. That branch is a silent no-op — no state written, no retry, no provisioning failure — so the box kept lexical FTS **and looked healthy doing it**.

`step_post_update` now calls `ensure_local_embeddings`, and the fresh-install arm waits for readiness first.

**`ollama_wait_ready` checks `is-enabled`, not `is-active`.** Local AI's off switch is `systemctl disable --now`; the runtime's idle standby is `stop` and deliberately never `disable`. So `is-active` cannot tell "asleep" from "switched off for good", and starting on it would silently reverse the owner's own decision — the one thing an "install what is missing" pass must never do. A test pins this and fails without the guard.

It probes the URL the helper itself probes, read from the same variable: `localhost` and `127.0.0.1` are not the same address on a box whose resolver answers ::1 first.

Additions go *into* the existing `if/elif` chain at `install.sh:4913-5048` — that region is one statement, and removing a branch header would leave the rest dangling.

---

## 3. Coding Agent: the file roster is gone from the run rail (`b0226f63`)

A run that touched 29 files filled the rail with a wall of path chips that told the reader nothing the count above already did — and the count was already there as a StatTile in the figures grid.

The duplication went further than the number: `codingAgent.statFiles` and `codingAgent.filesTitle` were **both literally "Files changed"**, so the same label rendered twice on one screen. `filesTitle` is dropped from all ten catalogues — `hermes-translations.test.ts` fails a locale carrying a key English lacks, so it is all or none.

---

## 4. Memory Shard: the plain diamond (`446c526a`)

`diamond_shine` draws a brilliant-cut gem with sparkle marks laid over its facet lines; at the 48px the desktop draws an app icon, the two line sets merge and it reads as a blob. `diamond` is the same gem without them.

Both ship a ligature in the bundled subset — checked against the font's own GSUB table rather than assumed — so this is a legibility choice, not a repair. Nothing was broken.

---

## Tests

**172 pass** across `install-kokoro-tts`, `install-whisper-stt` (new, 16), `install-ensure-local-embeddings` (new, 16), `clawbox-tts`, `local-models-installer-contract`, `install-x64`; plus 273 across the coding-agent, desktop-icon and translation suites. `tsc --noEmit` clean, eslint 0 errors, `bash -n` clean on both shell files.

Two guards were verified by reverting them and watching the tests go red: the `is-enabled` owner-switch check, and the write-the-unit-last ordering.

Three Kokoro tests that pinned the *exclusion* now pin the *inclusion* — one of them was named for the hour that was never measured.

https://claude.ai/code/session_01T3ZN3UeRKTumzvUzWtXXRi

## Summary by CodeRabbit

- **New Features**
  - Added faster-whisper speech-to-text installation with CUDA support, model setup, validation, and service configuration.
  - Added automatic local embedding recovery during updates with readiness checks and safe fallback behavior.
  - Improved voice installation options, including independent speech-to-text setup.

- **UI Updates**
  - Updated the Memory Shard icon to use the `diamond` glyph.
  - Run statistics now show file counts without displaying detailed file lists.

- **Bug Fixes**
  - Improved installation reliability by waiting for required services before continuing.

- **Tests**
  - Expanded coverage for embedding recovery, speech-to-text installation, service readiness, and run-page behavior.


yalexx:
All three findings addressed in d7fad3e1, plus the CI failure.

**Unreported `write_whisper_unit` failure — fixed, and it was the important one.** `install_whisper_stt` runs under `|| STT_RC=$?`, so errexit is off inside its body; a failed `mkdir` or heredoc redirection let it stamp the install and announce a ready engine whose unit file did not exist. That is exactly the disagreement between `local-models.ts` (which trusts unit presence) and `stt-local.ts` (which demands a real import) that the write-last ordering exists to prevent — arriving from the other direction, and worse, because a stamped box skips the work on every later update and would have carried the defect permanently. Both call sites now return 12 before the stamp and before anything claims readiness, with a test that fails without it.

**Vacuous ordering assertion — fixed as suggested.** `-1` is less than every index, so it passed on a run where Kokoro never happened.

**Unguarded anchor — fixed as suggested.** `slice(-1)` returning the file's last character rather than an empty string is a nasty one: the exit-contract test would have iterated zero lines and passed having inspected no code. There is one `ttsOnlyArm()` helper now and it throws.

Separately, the `test` failure was mine and not a flake. My readiness line said "Ollama did not answer in time", and `ensure-local-embeddings.test.ts` asserts that phrase is *absent* when the failure is a parse error — the point being that a box whose core answered fine and whose installer could not read it must blame `python3`, not the core. Two unrelated "did not answer" sentences in one run send the operator to the wrong box, so the readiness line says "not reachable" now, which is the vocabulary `ensure-local-embeddings.sh` already uses for that condition.

208 tests green across the five affected suites; `tsc --noEmit` clean, eslint clean, `bash -n` clean.

yalexx:
Second pass addressed in 25b25eb9. Both valid.

**Disabled vs unreachable Ollama.** This was a real defect I introduced. `ollama_wait_ready` returned 1 for two different things, and both callers print a connectivity diagnostic on failure — so a box whose Ollama is off on purpose read:

```
  Ollama is switched off on this box - leaving it that way.
  Ollama is not reachable - semantic memory stays on lexical FTS for now
```

The second line contradicts the first and sends whoever reads it after a network fault that does not exist. "Switched off" returns 2 now, and only 1 prints the connectivity line. A test pins the two codes apart.

**Readiness message on the failure path.** Correct — the test asserted the write error but not the absence of the readiness line, so a regression printing `faster-whisper ready` before returning 12 would have satisfied every assertion in it. Added.

On the shellcheck SC2329 note ("this function is never invoked"): `ollama_wait_ready` is invoked, from `step_ollama_install` and from `ensure_local_embeddings`; the checker is not seeing the call sites, the same way it did not see `free_memory_for_build`'s in #643.

142 tests green across the four affected suites, `tsc --noEmit` clean, `bash -n` clean.

yalexx:
Third pass addressed in 7b4d859a, plus the CI failure.

**Disabled-but-running Ollama — valid, and fixed as described.** `systemctl disable` without `--now` leaves the unit up until the next boot, and asking `is-enabled` first refused a reachable Ollama and skipped a repair that would have started nothing. The rule being enforced is *never **start** an engine the owner switched off*, not *never use one* — so activity is checked first, and the boot setting decides only when the service is already down. The regression case you asked for is in: disabled + active must return 0 and must not call `start`.

**The `test` failure is not from this PR.** `coding-agent.test.ts` failed with `ENOTEMPTY: directory not empty, rmdir '.../code-projects/site/.git'` — 1 of 10,936, in a suite this branch does not touch (the diff is `install.sh`, `scripts/install-voice.sh` and their tests). The same failure blocked #643. It passes locally on every run.

Rather than force it through again, it is fixed here. The test never creates that `.git` — the code under test does, through a spawned `git init` — so the directory can still be gaining files at the moment `afterEach` removes the tree. Node's `rm` retries precisely this class of error (`EBUSY`/`EMFILE`/`ENFILE`/`ENOTEMPTY`/`EPERM`) and defaults to not retrying at all, so the cleanup now asks it to.

Ten-odd other suites share the same cleanup line; only this one spawns `git init`, so only this one is changed.

164 tests green across the affected suites, `tsc --noEmit` clean, `bash -n` clean.

Issue #304, opened 2026-08-08
Harden all /setup-api callers against expired-auth redirects (generalize #231)


KrasimirKralev:
Follow-up to #231 / #303, which fixed only the VNC app.

The same latent bug — a raw `fetch("/setup-api/...")` gets the middleware's HTML login redirect on an expired session and throws when `.json()` parses the login page — remains in ~30 other callers across ~15 files (e.g. `i18n.tsx`, `SettingsApp.tsx`, `mascot-client.ts`, `use-clawbox-login.ts`, `useOllamaModels`/`useLlamaCppModels`/`useDeviceAddress` hooks, `DoneStep`, `AIModelsStep`, `ClawKeepApp`, `FilesApp`, `SystemUpdateApp`, `app/page.tsx`).

Two options:
1. **Deepest fix (preferred):** make `src/middleware.ts` return a JSON 401 for `/setup-api/*` **unconditionally** (drop the `Accept`-header gate for that prefix) — fixes every caller at once with no client changes. Caveat: a few `/setup-api/*` routes serve non-JSON for direct navigation/embedding (`webapps?app=` iframes, `apps/icon/[appId]` `<img>`, `files` downloads); a 401 there is defensible (a login page inside an `<img>`/iframe is useless anyway) but needs a deliberate check.
2. Migrate the remaining callers to the `fetchSetupJson` helper from #303.

Option 1 makes the helper’s `Accept` trick reduce to hardening.

KrasimirKralev:
Fixed in #316 (merged to beta). Middleware now returns a JSON 401 for the whole `/setup-api/*` prefix regardless of the `Accept` header, so every fetch() caller gets a structured 401 instead of the HTML login redirect that made `.json()` throw on an expired session. 47/47 middleware tests green on-device. (Auto-close didn't fire because the PR targets beta, not the default branch.)

PR #649, opened 2026-09-04
Lock the desktop while an update owns the box

yalexx:
## The problem

An update runs `git reset --hard` and `git clean -fd` over the project **while the desktop is still on screen**, and every app on it can write through `/setup-api`. A window left open can save into a tree being rewritten underneath it.

Nothing stopped that: `src/app/page.tsx` had no awareness of updates at all.

## The lock

The flag lives on **disk** (`update_in_progress`, `src/lib/update-lock.ts`), not in the updater's memory, because the reader is the **middleware** — it answers before any route handler is entered. It reads the same mtime-cached `data/config.json` snapshot it already reads for `setup_complete`, so the lock costs no extra I/O.

While set, `/` and `/app/*` redirect to `/updating`. Deliberately narrow:

- **`/setup-api` is not redirected** — the updating page polls the status route, and an API answering a navigation redirect with HTML is the defect `#304` fixed.
- **The gateway is not redirected** — the assistant keeps answering on it throughout an update.
- **The check sits after the session check**, so an expired session still gets `/login` rather than a screen whose content needs a session.
- **It fails OPEN on an unreadable config**, unlike the auth fields beside it: a corrupt `config.json` is no evidence of an update, and locking the desktop on it would take away the surfaces the owner needs to fix the box.

## The two properties that decide whether this helps or harms

Neither shows on the happy path. Both are pinned by tests.

**It survives the reboot.** The release is in `launchUpdate`'s `finally`, which does *not* run when `do_rebuild` kills the server — so the flag is still set when the box returns for `post_update`, `gateway_verify` and `verify_build_identity`. A test asserts `updateClawBoxAndReboot` and `waitForRebuildToTakeOver` never clear it.

**It cannot outlive a dead update.** At boot, `resumeContinuation` finds nothing to resume and releases the lock, so a run that died between setting the flag and writing its continuation flag cannot lock the owner out.

## Three things found by actually serving it

**1. eslint caught a button that would not have worked.** The stuck panel had a "Back" link — but while locked, the middleware redirects `/` straight back to `/updating`. The escape that genuinely works is a restart (the boot release above), so the panel says that instead.

**2. The page rendered raw i18n keys.** `useT()` outside an `I18nProvider` returns a fallback that renders the *key*, and the root layout is a server component that mounts none. `/login` and `/app/[id]` each mount their own; this is the third.

**3. That alone was not enough.** `I18nProvider` loads translations in an effect through a dynamic import, so SSR always emits keys and the client fills them in. Every other page can rely on that. This one cannot — it is the **one screen guaranteed to be open while the box is offline**, and `i18n.tsx` says so itself:

> *"A chunk load can fail for reasons that are not bugs: the device is offline mid-update … Keep whatever copy is already in state (English, or the raw keys on a first load)"*

Raw keys for the length of an outage is not an acceptable answer, so every string has an English floor — translated when the catalogue is there, English when it is not, never the key.

## Verified on hardware

```
UNLOCKED    /            -> 200
LOCKED      /            -> 307 -> /updating
LOCKED      /app/notes   -> 307 -> /updating
LOCKED      /updating    -> 200        (no redirect loop)
LOCKED      /setup-api/… -> 200        (the page can still poll)
served HTML  "System Update" · "Updating your ClawBox with the latest software..."
UNLOCKED    /            -> 200        (desktop returns)
```

**No new translation keys** — `update.title`, `update.updatingDescription` and `update.preparingUpdate` already ship in all ten locales, which also avoids the collision hazard in the merged `update.*` namespace.

The screen fails **closed**: a failed poll is the normal course of an update, not evidence it ended, so it holds rather than dismissing itself. It shows a step count and elapsed time, never a percentage of a duration nothing has measured.

## Tests

259 pass across `update-lock` (new, 13), `middleware` (6 new), `updater`, `updater-continuation-boot` and the update routes. `tsc --noEmit` clean, eslint clean.

The middleware tests were verified to have teeth by removing the redirect and watching two go red.

## Not in scope

The lock stops the **owner**, not the **assistant** — middleware admits the MCP bearer, so `preferences_set`, an app install or a coding-agent run can still write mid-update. Same fence, small addition, but it is a separate decision about interrupting the agent.

https://claude.ai/code/session_01WS1zsR5phBpjQPJSepJDG9

## Summary by CodeRabbit

* **New Features**
  * Added an update status page showing progress, elapsed time, restart guidance, and automatic return to the desktop when complete.
  * Authenticated desktop navigation is redirected to the update status page while updates are in progress.
  * Update status remains available during temporary server outages, with offline-safe fallback text.

* **Bug Fixes**
  * Update state now persists through restarts and recovers safely if configuration storage is temporarily unavailable.
  * API and gateway access remain available during updates.


yalexx:
Addressed in b169ec3c — the part that matters was right, and I went a different way on the other half.

**Swallowed failures — fixed.** `setUpdateLock()` hid a failed config write, so an update could launch with the desktop unlocked and nothing anywhere would say so. Both functions report now and return whether they succeeded.

**But a failure still does not stop the update, deliberately.** Refusing to update a box because a *courtesy lock* could not be written is the worse outcome by some way: an unwritable `config.json` is exactly the kind of state an update exists to repair, and a reachable desktop during one update is a smaller harm than a box that can no longer be fixed at all. The same reasoning covers a failed release — and that one is already recoverable without anyone doing anything clever, because the next boot finds no update to resume and clears the flag.

**The fire-and-forget is gone, which is the real substance of the finding.** The acquisition moved into `runUpdate` and is awaited, replacing the two `void setUpdateLock()` calls. As you noted, `config-store.set` has no `await` inside today so the race was theoretical — but nothing here should depend on that staying true, and one awaited call before the first step is simpler than two that hope.

**It also fixed something neither of us was looking at.** Putting the acquisition in `startUpdate` meant the *OpenClaw-only* update flow locked the desktop too — and that flow reinstalls a package and bounces the gateway; it never runs `git reset --hard`. Locking the owner's desktop for that is over-reach. The new placement is scoped to the flow containing the RESTART step, the same test the drift baseline beside it already uses.

Three tests pin it: awaited and before the step loop, scoped to the destructive flow only, and a write failure that returns false and warns rather than throwing.

261 tests green across the five affected suites; `tsc --noEmit` clean, eslint clean.

PR #650, opened 2026-09-04
Update screen: draw it in the device's style, and stop the lock being lost mid-run

yalexx:
Two changes, both about the in-app update. Scope reduced at the owner's request — the memory-limits work briefly on this branch has been dropped, and `beta`'s cgroup limits, the Settings line and `free_memory_for_build` are all untouched.

---

## 1. The lock was being lost mid-update (`1378cbdd`)

Observed on hardware, and it is what the owner saw: **back on the desktop while the update was still running.**

```
17:44     update_in_progress = true
17:51     update_in_progress = absent      <- gone, mid-run
17:53:29  [Updater] Update process finished
```

Nothing in the web server clears the flag mid-run, so the write was lost **from outside**. `config-store.set` is an unlocked read-modify-write of the *whole* of `data/config.json`, and `post_update` runs `install.sh` and restarts the gateway — both of which have that file open by their own paths (`install.sh:2894`, `gateway-pre-start.sh`'s `CLAWBOX_DEVICE_STORE`). A writer that read the file before the flag was set and wrote after removes it silently, and this process would never know.

Re-taking the lock at every step boundary heals that within one step instead of leaving it lost for the rest of a ten-minute run. One small write per step, against a run that is already installing packages and rebuilding the tree.

**This is a repair, not a cure.** Two processes doing read-modify-write on one JSON file with no lock will keep losing writes; this only makes the lock survive it. The underlying race deserves its own change and I have not attempted it here.

## 2. The screen is drawn in the device's own style (`f9deeefe`)

The screen an owner watches for ten minutes was a bare ring and two lines of text. It now uses `ReconnectStage` — the same crab, pulse rings, orbiting dots and step checklist the setup wizard already uses for its restart overlay. That component *is* the device's wait screen; a second one styled by hand would be a second answer to a settled question, and it would drift.

It also says more: the full step list with the running step spinning and finished ones ticked, the count, the elapsed clock, and the line the running step last logged when the server offers one.

**A server-rendered floor sits underneath the overlay, and it is not decoration.** `ReconnectStage` draws through `createPortal` and returns `null` when there is no document — so on its own this page is **blank** until the client bundle runs. Every other page can live with that; this one cannot. The owner arrives here by a middleware redirect at the moment the update is about to stop the web server, and a chunk request that loses that race would leave them staring at nothing for the length of the outage — worse than the plain screen the styling replaces.

Verified by serving it: the body was **empty** with the portal alone, and carries the title and description with the floor.

## Tests

326 pass across the 14 affected suites, including the four that failed on the earlier version of this branch (`system-profile-scripts`, `desktop-power-wiring`, `root-steps`, `system` routes) — all collateral from the dropped memory-limits work. `tsc --noEmit` clean, eslint 0 errors.

https://claude.ai/code/session_012ihKizd8EqgJ3XHM8kPir5

### DIFF
--- a/src/app/updating/page.tsx
+++ b/src/app/updating/page.tsx
@@ -2,6 +2,7 @@
 
 import { useEffect, useRef, useState } from "react";
 import { I18nProvider, useT } from "@/lib/i18n";
+import ReconnectStage from "@/components/ReconnectStage";
 
 /**
  * The screen the owner sees while an update owns the box.
@@ -12,6 +13,12 @@ import { I18nProvider, useT } from "@/lib/i18n";
  * over the project, and an app left open on the desktop can write through
  * /setup-api into the tree being rewritten underneath it.
  *
+ * It is drawn with ReconnectStage — the same crab, pulse rings, orbiting dots
+ * and step checklist the setup wizard uses for its own restart overlay — rather
+ * than a bespoke spinner. That component already IS the device's wait screen;
+ * a second one styled by hand would be a second answer to a settled question,
+ * and it would drift.
+ *
  * Three things it must get right, in order of how badly each one bites:
  *
  *  1. FAIL CLOSED. Partway through, do_rebuild stops the web server and the box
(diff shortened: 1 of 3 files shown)


yalexx:
Both inline findings are against code that is no longer on this branch, so neither is actionable — and I would rather say so than quietly leave them unanswered.

They review `step_drop_memory_limits` and the `step_post_update` wiring from commit `410d84dd`, which removed the cgroup memory limits. That commit was **withdrawn** at the owner's request and the branch force-pushed; the scope is now the update screen and the update-process fix only.

Verified rather than asserted:

```
git diff beta...HEAD -- install.sh   ->  empty (identical to beta)

step_drop_memory_limits   0 occurrences
step_swapfile             0 occurrences
step_resource_limits      1 occurrence   (intact, as on beta)
```

And `desktop-power-wiring.test.ts:144-147` — the test the first comment asked me to change — correctly still requires `step_resource_limits` in `step_post_update`, because that is once again what `step_post_update` calls.

For the record, the second finding was a **good catch on the code it was reviewing**: the cleanup removed the GNOME session drop-in but the live `set-property` reset covered only Ollama and the browser, so `user@UID.service` would have kept its cap until a reboot. If the memory-limits removal is ever revived, that fix belongs in it.

The SC2329 "function is never invoked" notes are the usual false positives — `shellcheck` does not see `install.sh`'s dispatch call sites.

PR #652, opened 2026-09-04
feat: run memory search's embedder on llama.cpp, capped and on demand; retire the Ollama row

yalexx:
## Summary

Memory recall needs an embedding model resident on the box. It was Qwen3-Embedding-0.6B inside **ollama**, which launched its bundled llama.cpp with `-c 4096 -b 2048 -ub 2048` — **2.8 GB resident** whenever the agent was in use, of which 1.2 GB was a compute buffer reserved for a 2,048-token batch (`n_ubatch × n_vocab × 4 B`) that real traffic never approached (153 calls measured: median 317 tokens, max 484). Since every embed is agent traffic (OpenClaw syncs on search; there is no timer), that 2.8 GB was held exactly when Kokoro, Whisper and Gemma also wanted memory.

This PR moves the embedder onto **ClawBox's own llama.cpp** — the "llama.cpp pattern" Gemma already uses — as the system unit `clawbox-embed.service`, and retires the Ollama row from Settings → Local AI (nothing that ships needed it any more: Ollama was already gone from every provider picker, and Local-only mode routes to Gemma).

### What changes on the box

- **Engine**: `/usr/local/bin/llama-server --embedding --pooling last --embd-normalize 2 -c/-b/-ub 1024 -np 1 -ngl 99 -fit off --flash-attn on -ctk/-ctv q8_0` on port 8081, GGUF at `data/embed/models/` (`hf download`, same Q8_0 weights; cosine 0.9995 vs ollama's vectors). Predicted ~2.0 GB on the GPU; `EMBED_N_GPU_LAYERS=0` measured 1.4 GB CPU-only (queries 27 ms — faster than GPU's 50 — indexing 692 ms/chunk vs 53). All dials in `.env` (`EMBED_*`), defaults pinned across TS/shell/env by `embed-server-pin.test.ts`.
- **Lifecycle** = the existing local-AI proxy: OpenClaw's `memory.search` is pointed at `/setup-api/local-ai/embed/v1` (`provider: openai-compatible`, bearer = `data/.local-ai-token`), never at port 8081. The proxy wakes the unit (`sudo systemctl start clawbox-embed.service`) and stops it 10 idle minutes later — **0 MB when idle**. It has to be an always-listening endpoint: OpenClaw's embedding client retries a refused connection 3× inside ~1.5 s and then answers keyword-only.
- **Memory cap**: a system unit so it can carry a cgroup cap (`CLAWBOX_EMBED_MEMORY_HIGH/MAX` = 2560M/3072M) — the web server's unit has none, which is why Gemma's child is uncapped today, and a user unit would share the desktop's 2048M. Plus a wake guard: below 2,300 MB MemAvailable the wake is refused and OpenClaw falls back to keyword search rather than pushing the box into swap.
- **Retrieval quality preserved**: OpenClaw's ollama adapter prefixed every search query with `Instruct: Given a user query, retrieve relevant memory notes and documents\nQuery:`; its openai-compatible adapter does not. `queryInputType/documentInputType` make it label each request and the proxy restores the prefix on queries (`embed-query-instruction.ts`). Without this recall degrades silently while the index stays "healthy".
- **No dropped documents**: OpenClaw chunks by characters (400 × 4), so a 1,600-char CJK chunk can be 1,600 tokens and llama-server refuses anything over `-ub`. `embed-input-fit.ts` trims an overflowing input via `/tokenize`/`/detokenize` — what ollama's `truncate: true` did — instead of failing the document. `EMBED_BATCH=2048` remains a one-line dial for CJK-heavy boxes (+0.6 GB).
- **Provisioning**: `ensure-local-embeddings.sh` fetches the GGUF, writes the config (settings first, provider last), and `memory index --force` — the provider change alters the index identity and OpenClaw pauses vector search until a rebuild. `step_embed_model` caches the model as root ahead of time so a first request never downloads inside OpenClaw's 120 s batch budget. Runs from `post_update` (after the unit/sudoers/cap are installed) and detached at every gateway start; the gateway unit is now `After=clawbox-setup.service` so the first sync after boot does not race the proxy.
- **Update path**: `post_update` stops ollama after the switch (stop, never disable); `free_memory_for_build` stops the embed unit the way it stops ollama; `armIdleStopIfRunning("embed")` at web-server boot re-arms the idle stop for a unit that outlived a restart; a stale service token is repaired at every gateway start without a reindex.
- **UI**: the Memory search row IS the engine row — "Qwen 3 via llama.cpp", starts when needed / searching on this box / stopped after an error / not installed — with no switch (the proxy owns the lifecycle) and the Memory Shard pointer. The Memory Shard wizard fetches the model as a streamed root step (`embed/install`) and asks for a FULL first pass. Ollama's chat-provider backend (`local-ai/ollama`, `ollama/*`, reset wipe) is untouched.

### Verified on hardware (Jetson Orin Nano, 7.4 GB)

- ollama runner buffers from its own log: weights 604+157, KV 238, compute 1209 MiB → RSS 2827 MB. 
- `-ub 512 → 2048` on this llama.cpp build: RssAnon 697 → 1674 MB (+977, = the 888 MiB logits reserve + KV), so the batch is the lever.
- CPU-only run: 1373 MB RSS / 697 anon, cosine 0.9995 vs ollama, 27 / 692 / 721 ms for 9 / 241 / 366 tokens.
- OpenClaw client contract (request shape, `input_type`, retry budget, identity, timeouts) read out of the installed package; details in `src/lib/embed-query-instruction.ts` and `embed-input-fit.ts`.

### Tests

New: `embed-query-instruction`, `embed-input-fit`, `local-ai-runtime-embed` (verbatim sudoers argv), `embed-server-pin`, proxy embed cases, middleware pass-through. Rewritten: `ensure-local-embeddings` (stub `hf`/`curl`/`openclaw`, 40 cases), `install-ensure-local-embeddings` (`step_embed_model` driven against stubs), `local-models` (the embedder row's states), `memory-shard-wizard`, route/panel/e2e fixtures, 10 locales × 2 catalogues. `check-sudoers-coverage.sh`: 45 grants, 0 gaps.

### What the first update to this branch found (fixed here, commit 779a469)

The box came out of the update **half-switched with a clean log**: `memory.search.provider` still `ollama`, pointed at the OpenAI-shaped proxy, no input types, the gateway's embed probe answering 405, the index marked dirty. Two causes, both in the automatic path:

- **A killed write sequence.** The script wrote the six keys in six `openclaw config set` calls. The bearer write (`memory.search.remote`) makes the gateway schedule its own restart, and during an update the updater's health step turns that into `systemctl stop clawbox-gateway.service` — which ends every process in the unit's cgroup, the detached provisioning script included, between the bearer and the input types.
- **A name collision on `OPENCLAW_HOME`.** The run the updater then launched itself (`runCurrentGatewayPreStart`) inherited `OPENCLAW_HOME=/home/clawbox/.openclaw`. ClawBox uses that name for the `.openclaw` directory; the OpenClaw CLI reads it as the *account home* and puts its tree at `$OPENCLAW_HOME/.openclaw`. So the CLI wrote all six keys to `~/.openclaw/.openclaw/openclaw.json`, "reindexed" an empty workspace there, and the script — which reads the real file by path — logged success and cleared the reindex marker. (The env block dates from #567 and is on beta; it was latent there because beta's script never needed a write.)

The fix:

- `ensure-local-embeddings.sh`: **one** `config set memory.search '<json>' --strict-json --merge` — all-or-nothing, keeps `extraPaths`, bearer built by python — and the file is **read back** before success is logged; `OPENCLAW_CONFIG_PATH`/`OPENCLAW_STATE_DIR` exported and `OPENCLAW_HOME` unset, so the CLI (and `memory index`) can only touch the tree the script reads.
- `gateway-pre-start.sh`: the same pin for its own ~41 CLI calls and everything it launches; reads `CLAWBOX_OPENCLAW_HOME` first; `remove_stray_state_tree` deletes a literal `.openclaw/.openclaw` nesting that carries neither a workspace nor credentials.
- `updater.ts`: passes `CLAWBOX_OPENCLAW_HOME` plus the two canonical CLI overrides to the pre-start, never `OPENCLAW_HOME`.
- Tests: the stub CLI now follows the real CLI's path rules and `--merge`; new cases for the pin (a decoy `OPENCLAW_HOME` must stay empty), the read-back (a CLI that answers 0 after writing elsewhere is not "done"), the single write, and the cleanup's guards.

Verified on the box after deploying the fix (web-server restart, no reboot): a gateway restart's pre-start removed the stray tree; the real CLI run under a decoy `OPENCLAW_HOME` against a scratch copy wrote the copy in one call, left the decoy empty and reindexed in the pinned state dir; `memory status --deep` → `openai-compatible`, index complete, probe ok; `memory search` returns ranked hits; `ollama` inactive; 8081 on loopback only.

## Test plan

- [ ] `bun run test`, `bun run lint`, `tsc --noEmit`, `bash scripts/check-sudoers-coverage.sh`
- [x] On the box after the update: `systemctl cat clawbox-embed.service`; `openclaw config get memory.search` → `openai-compatible` at the proxy; `openclaw memory status --deep --json` → identity `valid`; `openclaw memory search --query …` returns hits; `ss -ltn` shows 8081 only on loopback
- [x] Settings → Local AI: no Ollama row; Memory search — Qwen 3 via llama.cpp — Starts when needed → On during a search; RSS of the unit ≈ 2 GB; unit inactive 10 min after the last search
- [x] Memory Shard → Start over → wizard → Index now: download skipped (model cached), provider switch, full pass, "On device" badge; `test memory` from the chat returns ranked hits
- [x] `ollama` inactive after post_update; a cold search wakes the embedder inside OpenClaw's 60 s budget (measured ~4 s)

https://claude.ai/code/session_01F5wDEic7kbWnVvMXjvnPbY

## Summary by CodeRabbit

* **New Features**
  * Added a local Qwen3 embedding engine for Memory Search, powered by llama.cpp.
  * Added model installation, status reporting, download progress, authentication, request sizing, memory safeguards, and idle shutdown.
  * Added local embedding proxy support for query and document requests.
* **Improvements**
  * Memory Search now configures local embeddings automatically and supports full reindexing.
  * Local AI inventory and setup messaging now reflect embedding-specific states and controls.
* **Documentation**
  * Documented service behavior, configuration options, resource limits, and local-only access.


yalexx:
Addressed every open thread in 9aa2fd50: HF pin validation in install.sh before anything runs as clawbox, same-origin gate on the memory-provider route, the wizard requiring the stream's closing success line, wall-clock bound on the helper's proxy wait, `CONFIG_PATH` for the remote-URL read, `||` for the wake guard, the bare-unit sudoers twins, the reworded headroom comment, localized Memory Shard names in all nine locales, and the test strengthenings (marker-file injection proof, folded stderr, both-markers-present ordering, stubbed /proc/meminfo in both directions, stopEmbed fallback matcher, remote URL in the "whatever the URL says" case, caps asserted beside their unit).

PR #673, opened 2026-09-05
fix: record plugin capability consent for every ClawBox-managed plugin

KrasimirKralev:
Closes TASK-603.

## Symptom

A gateway that will not come back, and printed advice that cannot fix it.

`openclaw update repair` ends with `Plugin "discord" requires capability consent; rerun with --accept-capabilities.` An operator who follows it literally loops: the flag belongs to a command he was not told to run. That is what turned the 2026-09-01 plugin problem into a dead end, and rebooting the box changed nothing.

## Harness first

The card is written as an upstream complaint. Half of it is refuted on the pinned core, and the other half is ours.

Verified read-only on an OpenClaw box (core **2026.8.1**):

* **`openclaw update repair --accept-capabilities` EXISTS** — it is in the command's own `--help` and in its examples. The card's claim 1 ("`update repair` has no such flag") does not hold on the core we ship. Not used here all the same, and the reason is in the code: `update repair --accept-capabilities` consents for **every** blocked plugin, including ones the owner installed himself from the Terminal. This PR consents only for the five ClawBox installs.
* The consent gate itself: `resolvePluginCapabilityConsent` / `resolvePluginArtifactCapabilityConsent` (`dist/capability-consent-2iHYP71I.js`) throw `Plugin "<id>" requires capability consent. Use openclaw plugins install or openclaw plugins enable with --accept-capabilities, then retry.` unless the install record already carries an accepted surface hash for the manifest on disk — or an **interactive** consent callback answers. A spawned CLI has neither.
* The native mechanisms used are the CLI's own: `plugins install … --accept-capabilities` and `plugins enable <id> --accept-capabilities`. Nothing is re-implemented.
* Claim 2 (legacy `exec-approvals.json`) is also weaker than the card says on 2026.8.1: `doctor` registers `ownerStep(detected.execApprovals, migrateLegacyExecApprovals)` in its own migration list, and the refusal text is built by `doctorFixInstruction`, which names `OPENCLAW_STATE_DIR` when it is set — the trap the card describes. Confirming the residue needs a destructive experiment on a box (planting a legacy file), so it was **not** attempted; recorded rather than guessed.

## Cause

ClawBox drove that CLI without the flag in four places, while passing it correctly for codex and the DeepSeek provider everywhere else:

| where | verb | effect |
|---|---|---|
| `src/lib/openclaw-channels.ts` `ensureChannelPlugin` | `plugins install @openclaw/discord\|whatsapp` | the Settings save answers `install_failed`; the npm error goes to `console.error` |
| same function | `plugins enable <id>` | same, one verb over |
| `install.sh`, in `step_openclaw_install` | `plugins install "$spec" --force` | refreshes every external plugin to the new core target — exactly when a declared surface changes — and swallows the refusal into `WARN: refresh failed` |
| `scripts/gateway-pre-start.sh` | `plugins enable codex --accept-capabilities` | the boot-time consent pass existed, for **codex only** |

And the recovery could not clear the state either. `src/lib/updater.ts`'s `repairCodexCapabilityConsent` matched the literal word `codex`, so a box blocked on `discord` went through the whole quiesce → pre-start → doctor → restart pass untouched and ended at `getGatewayFailureDetail`, which handed the owner the core's own sentence about a command he never ran.

Three of the repo's recurring classes are in that table: the **false success** of `|| echo WARN` over a refusal, the **false failure** of `install_failed` over a plugin that could have been installed, and a **dead end** the tool had the means to clear.

## Fix

Every ClawBox-driven `plugins install` / `plugins enable` records consent, and the repair covers every plugin ClawBox installs.

* `ensureChannelPlugin` passes `--accept-capabilities` on both verbs.
* `install.sh`'s refresh passes it for `codex|deepseek|discord|whatsapp|clawbox-email-directives` only — a plugin the owner installed himself is refreshed **without** it and the loop says so, because the flag consents to the *new* surface, not the one already on the box. Gated on `openclaw_is_v2`, like the codex install in the pre-start: a v1 pin (`OPENCLAW_PIN_VERSION`, a documented rollback override) rejects the option and would have failed every refresh behind that WARN.
* `gateway-pre-start.sh` gains the same consent pass for the other four managed plugins, **only for entries `openclaw.json` already says to load** — so it can never switch a channel on — time-boxed and non-fatal, as the codex arm is. This is the half that matters for a box that is already down: the owner reboots long before he updates.
* `updater.ts` reads **every** consent refusal in the journal, not the first. That was a live defect in the first draft of this PR: the tail spans the whole boot (`journalctl -b`) and the gateway restarts several times during an update, so a stale `codex` line ahead of a live `discord` one had the repair fix codex while `getGatewayFailureDetail` — which scans in reverse — reported discord. Ids are matched after stripping an `@openclaw/` or `openclaw-` prefix, since `ensureChannelPlugin` already handles a registry that calls the plugin `openclaw-discord`.
* Both new arms respect an owner-disabled entry, as `codexCapabilityRepairIsAllowed` always has. `plugins enable` is not a consent-only verb — it writes `plugins.entries.<id>.enabled = true` — so without that check an owner who ran `openclaw plugins disable discord` to get his box back would have had the channel switched on again, with consent granted in his name, by the next update.
* A refusal naming a plugin ClawBox does **not** manage is logged at `console.info` rather than skipped in silence, so the update log says why nothing happened.

## Both editions

Hermes has no OpenClaw plugins. `spawnOpenclaw` refuses on that edition (`OpenclawUnavailableError`), `step_openclaw_install` returns early for it, and `gateway-pre-start.sh` is the OpenClaw gateway's own ExecStartPre. Every path here is inert on the Hermes SKU; nothing is hidden or explained away in the UI because there is no control.

## RED

On unmodified `origin/beta` (`c2b1a44b`), with only the test files copied in:

```
 FAIL  src/tests/unit/gateway-pre-start-codex-runtime.test.ts [ … ]
Error: Codex plugin command flow not found
⎯⎯⎯⎯⎯⎯ Failed Tests 14 ⎯⎯⎯⎯⎯⎯⎯
 FAIL  install-plugin-capability-consent.test.ts > passes the consent argv it built, rather than a hard-coded flag
AssertionError: expected '      if ! as_clawbox -H "$OPENCLAW_B…' to contain '"${CAP_ARGS[@]}"'
 FAIL  install-plugin-capability-consent.test.ts > builds that argv only on OpenClaw 2
AssertionError: expected -1 to be greater than -1
 FAIL  install-plugin-capability-consent.test.ts > builds it only for the plugins ClawBox installs
 FAIL  install-plugin-capability-consent.test.ts > has no `plugins install` or `plugins enable` that skips consent entirely
AssertionError: expected [ Array(1) ] to deeply equal []
 FAIL  openclaw-channel-plugins.test.ts > installs the official npm plugin when the channel has none
AssertionError: expected [ 'plugins', 'install', …(1) ] to deeply equal [ 'plugins', 'install', …(2) ]
 FAIL  openclaw-channel-plugins.test.ts > enables a plugin the GATEWAY CONFIG does not carry …
AssertionError: expected [ 'plugins', 'enable', 'discord' ] to deeply equal [ Array(4) ]
 FAIL  openclaw-channel-plugins.test.ts > capability consent > accepts the declared capabilities when it installs a channel plugin
AssertionError: expected [ 'plugins', 'install', …(1) ] to include '--accept-capabilities'
 FAIL  openclaw-channel-plugins.test.ts > capability consent > accepts them on the ENABLE too …
AssertionError: expected [ 'plugins', 'enable', 'discord' ] to include '--accept-capabilities'
 FAIL  updater.test.ts > records consent for a NON-Codex managed plugin the gateway named
AssertionError: expected 'failed' to be 'completed'
 FAIL  updater.test.ts > repairs EVERY managed plugin the boot journal names, not just the first
 Test Files  4 failed (4)
      Tests  14 failed | … passed
```

(Five lines elided for length, all of the same two shapes. The pre-start file fails to LOAD on beta rather than failing test-by-test: the block its new harness extracts does not exist there, so the extraction throws at module scope. Honest, if blunt.)

## GREEN

Same four suites on this branch: **128 passed**. Full `--project unit` on the rebased head: **651 files, 10500 passed, 1 skipped**. `tsc --noEmit` clean; `eslint` on the changed files reports only the two `no-unused-vars` warnings `updater.ts` already had on beta; `bash -n` clean on both shell files.

## Sibling call sites swept

Every `plugins install` / `plugins enable` / `plugins disable` in the repo: `openclaw-deepseek-plugin.ts:60` and `gateway-pre-start.sh:3122,3124` already passed the flag; `gateway-pre-start.sh:2254,2271` (codex) already passed it; the four in the table above are fixed. `hermes-clawai.ts:1395` and `register-mcp.sh` are the Hermes CLI, which has no such flag.

## Named limits, not papered over

* `ensureChannelPlugin` still skips both verbs when the plugin is installed **and** already enabled in the config, so a save cannot re-consent a surface that widened underneath it. It does not have to: a surface widens when the package version changes, which is `install.sh`'s refresh, and the boot pass and the updater repair the state from the gateway's own refusal. Doing it unconditionally would buy a 10-12 s CLI cold start on every channel save.
* `trustChannelPlugin` (`openclaw-config.ts`) writes `plugins.entries.<id>.enabled` straight into openclaw.json, so it cannot pass the flag — but it runs on the same configure request, immediately after `ensureChannelPlugin` has consented that plugin. `gateway-pre-start.sh`'s `clawbox-email-directives` enable is the same shape, and the new boot pass now consents that id too.
* The updater's non-codex arm uses `plugins enable`, which cannot repair a plugin left as a project declaration with no `node_modules` (the state the codex arm's pinned force-install exists for). It warns instead of fetching an unpinned package from npm mid-update, and the gateway's own refusal still reaches the owner — what he does not get is a false claim that ClawBox repaired it.
* No device proof of the repair itself: it needs a box deliberately put into a consent-blocked state, which is a mutation of a box the owner is using. The gate and its message were read out of the installed core read-only, and the box's four managed plugins (`codex`, `deepseek`, `discord`, `whatsapp`) were enumerated the same way.

## Overlap

None with the concurrent `scripts/` / `production-server.js` build-race work: this touches `scripts/gateway-pre-start.sh` only, and only in the plugin-consent region.

## Review

`clawbox-review`, findings at `code-review-603.md` (2 HIGH, 7 MEDIUM, 5 LOW, 2 INFO). Applied: H1 (first-match-only), H2 (boot path unswept), M1 (unconditional flag on a v1 pin), M2 (blanket consent contradicting the whitelist), M3 (no owner opt-out on the new arm), M4/M5 (docstrings that overclaimed), M6 (missing `clawbox-email-directives`, id normalisation, the silent no-op), L1/L3/L5 (test docstring, path resolution and the `plugins enable` arm of the source sweep, commit body).
Refuted with reasons: L4 (a bespoke env for two `execFile` calls, where every other CLI call in this process uses the unit's env, would be inconsistent rather than safer; `--accept-capabilities` is itself the answer to the only prompt, and the 60 s timeout bounds it), and the M4 install-fallback (an unpinned npm fetch mid-update is the failure this repair exists to get out of). I1/I2 recorded as residuals.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved plugin installation, refresh, and startup handling for OpenClaw 2.
  - Automatically handles required capability consent for supported ClawBox-managed plugins, including Discord, WhatsApp, DeepSeek, Codex, and email directives.
  - Maintains compatibility with OpenClaw 1 and unknown releases.
  - Recovers from capability-consent failures across multiple plugins without unnecessary reinstallation.
  - Preserves explicitly disabled plugins and leaves owner-installed or unmanaged plugins unchanged.
  - Supports canonical and aliased plugin names.
  - Added clearer reporting and more resilient recovery when plugin configuration is missing or incomplete.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -2983,8 +2983,16 @@ for p in d.get("plugins", []):
       # rollback override) gives a CLI that rejects the unknown option, and
       # every iteration would then fail behind the WARN below with no plugin
       # refreshed at all.
+      #
+      # Matched on the NORMALISED id: the registry can key a plugin as
+      # `openclaw-discord` or `@openclaw/discord`, and the raw name would then
+      # fall through to the default arm and be refreshed without consent —
+      # silently, behind the WARN below. `$spec` keeps the raw name, which is
+      # what the CLI has to be given.
       local CAP_ARGS=()
-      case "$plugin" in
+      local PLUGIN_KEY="${plugin#@openclaw/}"
+      PLUGIN_KEY="${PLUGIN_KEY#openclaw-}"
+      case "$PLUGIN_KEY" in
         codex|deepseek|discord|whatsapp|clawbox-email-directives)
           openclaw_is_v2 && CAP_ARGS=(--accept-capabilities)
           ;;
--- a/install.sh
+++ b/install.sh
@@ -2963,7 +2963,36 @@ for p in d.get("plugins", []):
         @openclaw/*) spec="$pkg@$TARGET" ;;
       esac
       echo "    - $spec"
-      if ! as_clawbox -H "$OPENCLAW_BIN" plugins install "$spec" --force >/dev/null 2>&1; then
+      # --accept-capabilities, for the plugins CLAWBOX installs and only those.
+      #
+      # OpenClaw 2 refuses to install a managed plugin whose declared capability
+      # surface has not been consented to, and a refresh to a new core target is
+      # exactly when that surface changes. Without the flag this loop's only
+      # outcome on a widened plugin is the WARN below, and the gateway then
+      # refuses readiness with 'Plugin "<id>" requires capability consent' until
+      # somebody runs the CLI by hand (TASK-603).
+      #
+      # But the flag consents to the NEW surface, not to the one already on the
+      # box, so passing it for a plugin the owner installed himself would answer
+      # a widened-capabilities question in his name. Same whitelist and same
+      # reason as CLAWBOX_MANAGED_PLUGIN_IDS in src/lib/updater.ts; anything
+      # else is refreshed without it and says so, so he can rerun the CLI.
+      #
+      # Gated on the INSTALLED generation the way the codex install in
+      # gateway-pre-start.sh is: a v1 pin (OPENCLAW_PIN_VERSION, a documented
+      # rollback override) gives a CLI that rejects the unknown option, and
+      # every iteration would then fail behind the WARN below with no plugin
+      # refreshed at all.
+      local CAP_ARGS=()
+      case "$plugin" in
+        codex|deepseek|discord|whatsapp|clawbox-email-directives)
+          openclaw_is_v2 && CAP_ARGS=(--accept-capabilities)
+          ;;
+        *)
+          echo "      (not a ClawBox-managed plugin: refreshed without accepting new capabilities)"
+          ;;
+      esac
+      if ! as_clawbox -H "$OPENCLAW_BIN" plugins install "$spec" --force "${CAP_ARGS[@]}" >/dev/null 2>&1; then
         echo "      WARN: refresh failed (non-fatal; gateway-pre-start will retry on next boot)"
       fi
     done <<< "$INSTALLED_PLUGINS"
--- a/scripts/gateway-pre-start.sh
+++ b/scripts/gateway-pre-start.sh
@@ -2275,6 +2275,55 @@ elif [ "$CLAWBOX_OPENCLAW_V2" = "1" ] && [ "$CODEX_SHOULD_LOAD" = "1" ]; then
   fi
 fi
 
+# ── Capability consent for the OTHER ClawBox-managed plugins ────────────────
+#
+# The block above is the codex half of this, and has been here since OpenClaw 2
+# added declared-capability consent. The gateway refuses readiness for ANY
+# enabled plugin whose declared surface has not been consented to, and ClawBox
+# installs four more: deepseek (the provider ClawBox AI rides on), discord and
+# whatsapp (installed by the Settings panel when the owner asks for that
+# channel) and clawbox-email-directives (ours, copied out of the checkout
+# below).
+#
+# WHY IT MATTERS THAT THIS IS THE BOOT PATH. src/lib/updater.ts repairs the same
+# state from the gateway's journal, but only during an update. A box that is
+# already down — the 2026-09-01 outage was exactly this, on discord — gets a
+# reboot from its owner long before it gets an update, and every boot runs this
+# script. Without this block the reboot changed nothing and the gateway came
+# back to the same refusal (TASK-603).
+#
+# `enable` is the idempotent consent verb, as for codex: it records the current
+# reviewed surface when needed and leaves an already-consented plugin alone.
+# Only for entries openclaw.json ALREADY says to load, so this can never switch
+# a channel on — a plugin the owner disabled stays disabled, and one he never
+# asked for is never enabled by a boot script. Time-boxed and non-fatal,
+# because this is a blocking ExecStartPre.
+if [ "$CLAWBOX_OPENCLAW_V2" = "1" ]; then
+  MANAGED_ENABLED_PLUGINS="$(python3 - "$OPENCLAW_CONFIG" <<'MANAGEDPY' || true
+import json, sys
+MANAGED = ("deepseek", "discord", "whatsapp", "clawbox-email-directives")
+try:
+    with open(sys.argv[1], encoding="utf-8") as fh:
+        entries = (json.load(fh).get("plugins") or {}).get("entries") or {}
+except (OSError, json.JSONDecodeError):
+    raise SystemExit(0)
+if not isinstance(entries, dict):
+    raise SystemExit(0)
+for name in MANAGED:
+    entry = entries.get(name)
+    if isinstance(entry, dict) and entry.get("enabled") is True:
+        print(name)
+MANAGEDPY
+)"
+  for MANAGED_PLUGIN in $MANAGED_ENABLED_PLUGINS; do
+    if timeout 60 "$OPENCLAW_BIN" plugins enable "$MANAGED_PLUGIN" --accept-capabilities </dev/null >/dev/null 2>&1; then
+      echo "  $MANAGED_PLUGIN plugin capabilities accepted/current"
+    else
+      echo "  WARN: could not confirm $MANAGED_PLUGIN plugin capabilities; gateway readiness may remain blocked"
+    fi
+  done
+fi
+
 # Codex reads its ChatGPT session from a Codex CLI-style auth.json. Without
 # one the app-server falls back to api.openai.com with no bearer -> 401
 # "Missing bearer or basic authentication in header", which is what users hit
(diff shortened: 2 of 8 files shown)


KrasimirKralev:
Addressed all four review comments in `ce01e1c6` (rebased head `38b2165d`). All four are the same root — the registry can key a plugin as `openclaw-discord` or `@openclaw/discord`, and `ensureChannelPlugin` enables it under whichever name `plugins list` reported — plus the v1 gating.

**`install.sh:2988` — normalize before the whitelist.** Fixed. `$PLUGIN_KEY` strips `@openclaw/` then `openclaw-` and the `case` matches on that; `$spec` keeps the raw name, which is what the CLI has to be given. Pinned by `install-plugin-capability-consent.test.ts` → "matches the whitelist on the NORMALISED plugin id".

**`scripts/gateway-pre-start.sh:2315` — normalize the configured keys.** Fixed. The loop now walks `plugins.entries` and matches each key through a `canonical()` helper, while printing the CONFIGURED key so `plugins enable` is given the name the registry answers to. Two tests added against the shipped shell fragment: an enabled `openclaw-discord` and `@openclaw/whatsapp` are both consented under their own keys, and a disabled `openclaw-discord` is left alone exactly as a disabled canonical entry is.

**`src/lib/openclaw-channels.ts:261` — gate on the installed generation.** Fixed, and thank you: this was a real false failure on a rolled-back box. `capabilityArgs()` asks the installed binary (`installedOpenclawRelease()`) and appends the flag only for 2026.8+, the same rule `install.sh`'s `openclaw_is_v2` and `gateway-pre-start.sh`'s `CODEX_CAPABILITY_ARGS` already follow. Asked **once per save** and passed to both verbs — not memoised, which would be a probe-once over a fact that changes when the package is reinstalled — and an unreadable version answers v2, the shipped pin, so a probe that times out cannot turn the ordinary path into the consent refusal. Three tests: v1 omits the flag, unknown passes it, and the generation is asked exactly once for a save that runs both verbs.

**`src/lib/updater.ts:1100` — match entries by normalized id.** Fixed. `pluginConsentRepairIsAllowed` scans the entries and refuses the repair when ANY key normalising to the journal's plugin id is explicitly `enabled: false`. New test: a journal naming `discord` against `plugins.entries["openclaw-discord"].enabled === false` runs no `plugins enable`.

GREEN on the rebased head: `--project unit` 651 files / 10507 tests, `tsc --noEmit` clean, `bash -n` clean on both shell files. The one `Tests` failure on the previous head was `src/tests/components/hermes-oauth-inline.test.tsx`, a file this PR neither touches nor imports; it passes on its own here and is on this repo's known CI-flake list.

KrasimirKralev:
Addressed the follow-up comment in `c7d1b0c`-equivalent (rebased head `d1499732`).

**`src/lib/updater.ts:1072` — deduplicate by normalized plugin id.** Fixed, and you are right about the cost: `pluginsNeedingConsent` classified with `normalizeManagedPluginId` but stored the RAW id in the set, so one boot's journal carrying both `codex` and `@openclaw/codex` — a restart either side of a registry key change, or an alias from `plugins list` — put two entries in the same bucket and the loop repaired the plugin twice. For codex that is two pinned force-installs at 360 s each, back to back, inside a quiesced-gateway window on a Jetson.

Both buckets are now a `Map` keyed by the normalised id and valued by the FIRST raw spelling seen, which is the name the registry answered to when it refused — so the repair still runs under a name the CLI accepts, once. New test: a journal naming `discord` and then `openclaw-discord` produces exactly one `plugins enable`, and it is `discord`.

GREEN on the rebased head: `--project unit` 657 of 658 files, 10565 tests. The one failure is `src/tests/unit/webapp-icon.test.ts`, which this PR neither touches nor imports; it passes on its own here (27/27) and is on this repo's known parallel-run flake list. `tsc --noEmit` clean, `bash -n` clean on both shell files.

KrasimirKralev:
Merged into beta (dd058938) after final review MERGE-OK on the merged head and green CI on that head. Deploy of beta to both boxes follows; the box proof will be recorded on TASK-603.

PR #674, opened 2026-09-05
fix: show the agent harness version on the System Update screen

KrasimirKralev:
**TASK-548** — Hermes: System Update app never shows the Hermes agent version.

## Customer-visible symptom
The System Update screen is where an owner goes to ask "what version am I on". On a Hermes box it
named only ClawBox: the agent's version was readable in Settings → About and nowhere on the screen
built for the question. `VersionInfo` there carried `clawbox` and `openclaw` only, while
`/setup-api/update/versions` has carried a `hermes` block on every SKU that ships Hermes since
`getVersionInfo()` gained it.

## Cause
The payload grew `hermes` and `edition`; About was taught to read them and the update screen was
not. A stale comment in that file also claimed "The OpenClaw version is still shown in the ClawBox
card's release notes / version-info section" — nothing rendered it; the screen showed no harness
version at all.

## Harness-first
Nothing native is re-implemented. The version itself comes from the harness through ClawBox's
existing native call — `runHermesCli(["--version"])`, parsed by the purpose-built
`parseHermesVersion` — and this PR only reads the route that already does that. The reuse that was
missing was intra-repo: the payload shape and the per-edition rule were written out twice, in About
and here, which is what let the two panels drift.

## What changed
- `SystemUpdateApp` gains `hermes?` / `edition?` on `VersionInfo` and a read-only **Agent** card
  under the ClawBox card: OpenClaw unless the SKU is `hermes`, Hermes whenever the payload carries
  it, both on `dual`. Rendered outside the components gate, so it is still there on "Update
  complete" — which is exactly when an owner wants to read the version their agent came back on.
- The hero's `updates.push("OpenClaw")` is now gated on the edition as well as on `updateAvailable`.
  A guard, not a live defect: that flag is false on Hermes today only because `openclaw.current` is
  null, and a headline offering an OpenClaw update on a box that ships no OpenClaw should not rest
  on that.
- **Both harnesses are pinned by ClawBox**, so both rows say so. `HERMES_PIN_COMMIT`
  (`install.sh:625`) is re-checked and repaired by `step_hermes_install` on every update
  (`install.sh:2277`, called from `step_post_update` at `:4925`) — the direct analogue of
  `config/openclaw-target.txt`. The first draft of this card said "Installed from its own upstream",
  copied from a stale comment in `updater.ts`; that comment is corrected here too. Only the
  *installer script* comes from upstream, and it is fetched **from the pinned tree**.
- The Hermes version is rendered as the server produced it. `cleanVersion`'s grammar is shaped for
  OpenClaw/git-describe output — `parseHermesVersion` exists precisely because it does not fit
  Hermes — and running it a second time turned `Hermes Agent (dev build)` into `Hermes Agent` here
  while About, reading the field raw, kept it. The desktop can show both windows side by side.
- A null version renders `—`, not "not installed". The payload only carries a `hermes` block on a
  box that HAS Hermes, so null means the probe failed — which is what `hermes --version` does while
  `step_hermes_install` is moving the checkout aside mid-update. `ComponentCard` ten lines below
  already says the same nothing the same way. (This also removes the hardcoded-English outlier: the
  rest of the file is hardcoded English by convention, but "not installed" is the one string About
  has translated into all ten locales.)
- `VersionInfo` and the `shipsOpenclaw` predicate are exported and `SettingsApp` now imports them
  instead of restating both. That duplication is what the three divergences above were made of.

## Both editions
- `hermes`: Hermes row, no OpenClaw row, no OpenClaw in the update headline.
- `openclaw`: OpenClaw row, unchanged headline.
- `dual`: both rows — that SKU really runs both.

## RED → GREEN
New suite `src/tests/components/system-update-agent-version.test.tsx` (9 tests). Against the
unmodified beta components (`git checkout origin/beta -- src/components/SystemUpdateApp.tsx
src/components/SettingsApp.tsx src/lib/updater.ts`):

```
 ❯ |components| src/tests/components/system-update-agent-version.test.tsx (9 tests | 9 failed)
     × names Hermes and its version on the Hermes SKU 5007ms
     × does not offer an OpenClaw row on a device that ships no OpenClaw 5002ms
     × names OpenClaw and its version on the OpenClaw SKU 5002ms
     × does not re-parse a Hermes version the server already parsed 5002ms
     × says nothing rather than 'not installed' when the version could not be read 5002ms
     × tells the owner both harnesses ride the ClawBox update 5002ms
     × names both on the dual SKU, which really runs both 5002ms
     × renders an OpenClaw row for a body with no edition field at all 5002ms
     × keeps OpenClaw out of the update headline there 16ms

 FAIL … > keeps OpenClaw out of the update headline there
 AssertionError: expected 'New version available for ClawBox and…' to be 'New version available for ClawBox.'
```

(The eight timeouts are the agent card not existing to be found; the ninth is the headline.)

With the fix:

```
 Test Files  1 passed (1)
      Tests  9 passed (9)
```

Neighbouring suites — `system-update-app`, `system-update-app-drift`, `settings-update-section`,
`settings-about-version`, `settings-app`, plus the new one: 52 passed. `updater`,
`updater-branch-resolution`, `updater-build-identity`, `routes/update/versions`, `version-utils`:
110 passed. `tsc --noEmit`: clean. `eslint` on the changed files: 0 errors (7 pre-existing warnings
in `updater.ts`, none on a changed line).

Full `--project components`: 1364 passed, 2 failed in `chat-spoken-reply.test.tsx` — a known flake
in that file under a full parallel run, unrelated to this diff (it names none of these modules) and
18/18 green in isolation.

## Sibling call sites — handled or cleared
- `src/components/SettingsApp.tsx` About panel — **changed**: now imports the shape and the
  predicate rather than restating them.
- `SettingsApp.tsx` sidebar subtitle (`componentNeedsUpdate`) — reads `versionInfo.clawbox` only,
  so the edition gate does not apply. Cleared, unchanged.
- `mcp/tools/system.ts` `update_check` — returns the payload, so it inherits `hermes` for free.
- `mcp/tools/orientation.ts` `device_status` emits no harness version on `hermes` or `dual` — the
  same gap, agent-facing. Being fixed on the branch for TASK-543 (PR #671), which already owns that
  block and whose `VersionsPayload` already carries `hermes`; kept out of here so two PRs do not
  edit the same eight lines.
- `src/app/page.tsx` desktop update notice — unchanged and cleared: its openclaw term is guarded by
  the producer's explicit `updateAvailable: false`. Recorded as a residual (its `??` fallback would
  be true on a Hermes payload if that flag ever went missing).

## Recorded, deliberately not taken
`hermes.target` is hardcoded `null`, so the screen can never tell an owner their agent is off the
pin even though `HERMES_PIN_COMMIT` is a real target. Surfacing it needs the payload to compare a
40-char commit SHA against a release string, which is a different change; the comment in
`updater.ts` now says that is what `target: null` means, instead of the false "ships from its own
upstream".

## Summary by CodeRabbit

- **New Features**
  - System updates now show read-only agent version details for OpenClaw and Hermes devices.
  - Version information is displayed with edition-aware formatting after updates complete.
  - Update controls now identify ClawBox as the available update target.
  - The About page displays OpenClaw details only when supported by the device configuration.

- **Bug Fixes**
  - Hermes devices no longer incorrectly advertise OpenClaw updates.
  - Improved handling and messaging for unavailable, legacy, development, and failed version information.

### DIFF
--- a/src/components/SettingsApp.tsx
+++ b/src/components/SettingsApp.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import { useState, useEffect, useRef, useCallback } from "react";
-import SystemUpdateApp, { componentNeedsUpdate } from "@/components/SystemUpdateApp";
+import SystemUpdateApp, { componentNeedsUpdate, shipsOpenclaw, type VersionInfo } from "@/components/SystemUpdateApp";
 import Image from "next/image";
 import { createPortal } from "react-dom";
 import StatusMessage from "./StatusMessage";
@@ -582,14 +582,10 @@ export default function SettingsApp({ ui }: SettingsAppProps) {
      itself — its run, the beta channel, the branch pin, the force — is the
      System Update page (SystemUpdateApp, embedded), not state of this
      component any more. ── */
-  const [versionInfo, setVersionInfo] = useState<{
-    clawbox: { current: string; target: string | null; updateAvailable?: boolean };
-    openclaw: { current: string | null; target: string | null; updateAvailable?: boolean };
-    // Both optional: a device that has not been updated yet still answers
-    // /update/versions with the old two-key shape.
-    hermes?: { current: string | null; target: string | null; updateAvailable?: boolean };
-    edition?: "openclaw" | "hermes" | "dual";
-  } | null>(null);
+  // The shape and the per-edition rule come from SystemUpdateApp, the other
+  // reader of this payload. They were restated here, and the two panels had
+  // already drifted three ways by the time TASK-548 lined them up.
+  const [versionInfo, setVersionInfo] = useState<VersionInfo | null>(null);
 
   // Load version info on mount
   useEffect(() => {
@@ -5751,7 +5747,7 @@ export default function SettingsApp({ ui }: SettingsAppProps) {
                     actually runs instead; `dual` has both, so it shows both.
                     A server that predates the `edition` field falls through to
                     the OpenClaw row exactly as before. */}
-                {versionInfo?.edition !== "hermes" && (
+                {shipsOpenclaw(versionInfo) && (
                   <div className="flex justify-between text-sm">
                     <span className="text-[var(--text-muted)]">OpenClaw</span>
                     <span className="text-[var(--text-primary)]">{cleanVersion(versionInfo?.openclaw.current) ?? t("settings.notInstalled")}</span>
--- a/src/components/SystemUpdateApp.tsx
+++ b/src/components/SystemUpdateApp.tsx
@@ -8,13 +8,13 @@ import type { StepStatus, UpdateState } from "@/lib/updater";
 import { RESTART_STEP_ID } from "@/lib/update-constants";
 import { cleanVersion } from "@/lib/version-utils";
 
-interface ComponentVersion {
+export interface ComponentVersion {
   current: string | null;
   target: string | null;
   updateAvailable?: boolean;
 }
 
-interface VersionInfo {
+export interface VersionInfo {
   clawbox: ComponentVersion & { current: string };
   openclaw: ComponentVersion;
   // Both optional: a device that has not been updated yet still answers
@@ -30,7 +30,7 @@ interface VersionInfo {
  * payload from a server that predates the `edition` field reads as OpenClaw,
  * which is what those builds were.
  */
-function shipsOpenclaw(versions: VersionInfo | null): boolean {
+export function shipsOpenclaw(versions: VersionInfo | null): boolean {
   return versions?.edition !== "hermes";
 }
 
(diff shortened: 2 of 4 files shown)


KrasimirKralev:
**Device proof, read-only, on the Hermes box** (edition lock `CLAWBOX_EDITION=hermes`, no mutation, no lock taken).

The payload the new Agent card reads, as the device answers it today:

```
$ curl -b "$CJ" http://127.0.0.1/setup-api/update/versions
{"clawbox":{"current":"v4.0.0","target":"beta@26ad5df","updateAvailable":true},
 "openclaw":{"current":null,"target":"2026.8.1","updateAvailable":false},
 "hermes":{"current":"v0.20.5","target":null,"updateAvailable":false},
 "edition":"hermes"}
```

So on that SKU the screen now renders **Hermes — v0.20.5** and no OpenClaw row, and the hero cannot offer an OpenClaw update. Two details worth recording because they bear on the review round:

- `hermes.current` really is `"v0.20.5"` — already reduced server-side by `parseHermesVersion` from the multi-line banner the box prints (`Hermes Agent v0.20.5 (2026.8.19) · upstream 9dd6634c · local fcbd1076 …`). For **this** value `cleanVersion` happens to be a no-op, so the panels would have agreed here; the divergence needs the `Hermes Agent (dev build)` shape, where `cleanVersion` strips the parenthesis and About does not. Rendering it raw is right either way — the server already owns the parsing — but I am not claiming the box reproduced that half.
- `openclaw.current` is `null` while `updateAvailable` is `false`, which is what keeps the hero honest today. The new edition gate is what stops that resting on it.

Not a device proof of the fix: the box is on `v4.0.0`, two merges behind beta, and this branch is not deployed there.

PR #675, opened 2026-09-05
fix: unmatched ClawBox paths answer 404 instead of the gateway shell

KrasimirKralev:
Closes TASK-631 (F-29).

## Symptom

An authenticated GET of a path that does not exist in ClawBox's own namespace answers **200 with the OpenClaw Control UI**, and for an owner session that page carries the injected gateway token.

Measured read-only on an OpenClaw box at beta head `c2b1a44b`, owner-authenticated, with a browser's own navigation metadata (`Sec-Fetch-Mode: navigate`):

```
GET /setup-api        -> 200 text/html   Control UI, token script present
GET /setup-api/       -> 308 -> /setup-api, then the same
GET /portal           -> 200 text/html   Control UI, token script present
GET /portal/nope      -> 200 text/html   Control UI, token script present
GET /login-api/nope   -> 200 text/html   Control UI, token script present
GET /app              -> 200 text/html   Control UI, token script present
GET /app/x/y          -> 200 text/html   Control UI, token script present
GET /Setup-Api/nope   -> 200 text/html   Control UI, token script present
GET /Portal/nope      -> 200 text/html   Control UI, token script present
GET /setup-api/nope   -> 404 text/plain
```

(The token itself is never printed here or in the tests — the probe counts the presence of the injection script, `h.set("token", …)`, not its value.)

The card called this "routing hygiene rather than a leak", which still holds: `serveGatewayHTML` injects only for a caller who proved they are the owner, so a stranger gets the shell without a credential. It is a wrong answer either way — the owner asked for a ClawBox page or endpoint that does not exist and was handed somebody else's application.

## Cause

The `/setup-api` half was already fixed on beta (`4d46ac6a`), and the fix is one character short: the guard is `pathname.startsWith("/setup-api/")` — **with** the trailing slash — so the namespace root itself falls through. The other four namespaces were never covered, and the match was case-sensitive while Next's router is too.

`/app` is the surprising one: `src/app/app/[id]/page.tsx` matches exactly ONE segment and there is no page at the root, so both `/app` and `/app/x/y` land in the catch-all.

## Fix

`src/lib/clawbox-namespaces.ts` asks the ownership question once:

* **API namespaces** `/setup-api`, `/login-api` — an unmatched path is a missing endpoint and its caller is code, so it gets `{ error: "Not found" }`, the shape the twelve real 404s under those roots already use.
* **Page namespaces** `/setup`, `/login`, `/portal`, `/updating`, `/app` — `text/plain` "Not found".
* Matching is on a **segment boundary** (`/a` matches `/a` and `/a/b`, never `/ab`) — load-bearing, because `/setup-api` also starts with `/setup` and folding one into the other is the original auth bypass the middleware still carries a comment about.
* Lower-cased at the **route's** call site only. `src/middleware.ts` feeds the predicate an already-lower-cased pathname and three of its allow-gates depend on that exact comparison, so `isUnderRoot` itself stays case-sensitive.
* The block sits **above** the Hermes gate, which answers `text/plain` for every path: below it, `/setup-api/nope` would have been a JSON 404 on OpenClaw and a text one on Hermes — one shared surface, two shapes by edition.
* `src/middleware.ts`'s four `startsWith("/setup-api/")` gates now call `isSetupApiPath`, so the namespace root is the namespace everywhere. The only behavioural delta is at the 401-vs-redirect fork: an unauthenticated `GET /setup-api` gets the JSON 401 the rest of the namespace gets, instead of a login redirect. The other three are inert on the root (the bootstrap allow-list does not contain it; test-mode and the MCP bearer just reach a route that now 404s). Nothing gains access.

## Harness first

There is no OpenClaw mechanism to reuse here — this is ClawBox's own router deciding which paths it owns, and the catch-all exists precisely to hand the rest to the gateway. What was checked against the harness is the **boundary**: `/api`, `/assets` and the static trees in `gateway-static.ts` stay the gateway's, and the claim that `/portal`, `/updating` and `/app` are safe to take was verified rather than assumed — the pinned Control UI's own bundle (`dist/control-ui/assets/*.js`) carries route literals for `/chat`, `/sessions` and `/logs`, and **zero** for `/portal`, `/updating`, `/app`, `/setup` or `/login`.

## Both editions

The same code runs on both, and the reorder above is what makes the answer identical on both. The gateway namespaces are untouched, so nothing changes for the Control UI on OpenClaw and nothing new is reachable on Hermes.

## RED

On unmodified `origin/beta` (`c2b1a44b`), with only the two test files copied in:

```
⎯⎯⎯⎯⎯⎯ Failed Tests 13 ⎯⎯⎯⎯⎯⎯⎯
 FAIL  middleware.test.ts > answers the /setup-api ROOT as the API namespace, not as a page
AssertionError: expected 307 to be 401
 FAIL  gateway.test.ts > answers 404 for http://localhost/setup-api rather than the gateway shell
 FAIL  gateway.test.ts > answers 404 for http://localhost/Setup-Api/nope …
 FAIL  gateway.test.ts > answers 404 for http://localhost/login-api/nope …
 FAIL  gateway.test.ts > answers 404 for http://localhost/portal …
 FAIL  gateway.test.ts > answers 404 for http://localhost/portal/nope …
 FAIL  gateway.test.ts > answers 404 for http://localhost/Portal/nope …
 FAIL  gateway.test.ts > answers 404 for http://localhost/setup/nope …
 FAIL  gateway.test.ts > answers 404 for http://localhost/login/nope …
 FAIL  gateway.test.ts > answers 404 for http://localhost/updating/nope …
 FAIL  gateway.test.ts > answers 404 for http://localhost/app …
 FAIL  gateway.test.ts > answers 404 for http://localhost/app/x/y …
AssertionError: expected 200 to be 404
 FAIL  gateway.test.ts > answers 404 for http://localhost/setup-api/nope …
AssertionError: expected 'text/plain' to contain 'application/json'
 Test Files  2 failed (2)
      Tests  13 failed | 158 passed (171)
```

Two of the new tests are **guards, not RED evidence**, and pass on beta as well: `/setup-apiary` (must still redirect to /login) and `/setupsomething` (must still get the shell). They are the other half of the segment boundary.

## GREEN

Both suites on this branch: 178 passed. Full `--project unit`: **648 files, 10490 passed**. `--project components`: 143/144, the one failure being `chat-spoken-reply.test.tsx`, a file this PR does not touch or import and which is on the repo's known-flake list; it passes on its own and passed on the same head in a second run. `tsc --noEmit` and `eslint` clean.

## Deliberately not in this PR

* **The `isUnderRoot` consolidation.** Four other copies of `p === prefix || p.startsWith(prefix + "/")` exist — `isWizardPagePath`, `isGatewayOnlyPath`, the `PUBLIC_PREFIXES` loop (all `src/middleware.ts`) and `isBootstrapAllowedPath` (`src/lib/setup-api-gate.ts`, which also owns a `normalizeApiPath`). Each carries its own comment about why it is not a bare `startsWith`, because getting it wrong once WAS the auth bypass. Converting four security gates in a hygiene PR is the wrong trade; the new module is the right home for it and the follow-up is recorded.
* **Moving `/login-api` out of `PUBLIC_PREFIXES`.** It is there because the login POST needs it. The anonymous `GET /login-api/nope` shell fall-through — the one namespace where a stranger got a 200 HTML page, though never a token — is closed by the route's 404, which is what the new middleware test pins.

## Overlap

`production-server.js` is untouched: it intercepts only WebSocket upgrades, so all HTTP goes through Next and the catch-all is the single chokepoint. No overlap with the concurrent build-race work in `scripts/` / `production-server.js`.

## Review

`clawbox-review`, findings at `code-review-631.md` (1 HIGH, 2 MEDIUM, 5 LOW). Applied: H-1 (`/app` missing, and the comment excusing it was wrong), M-1 (case-sensitivity), M-2 (the Hermes gate made the JSON body OpenClaw-only), L-1 (`"not found"` → `"Not found"`, the only lowercase one in the repo), L-2 (assert the content-type and body, and pin `readEdition` so the suite tests the branch it means to), L-3 (the anonymous `/login-api/*` case), L-5 (the bundle check above). L-4 is the consolidation named under "deliberately not in this PR".

## Summary by CodeRabbit

* **Bug Fixes**
  * Unknown ClawBox API paths now return structured JSON 404 responses.
  * Unknown ClawBox page paths now return plain-text 404 responses instead of loading the gateway.
  * Route matching now respects namespace boundaries, preventing similarly prefixed paths from being misclassified.
  * Setup API authentication and access checks now consistently apply to the intended routes.

* **Tests**
  * Added coverage for namespace matching, authentication behavior, 404 responses, and gateway navigation handling.

### DIFF
--- a/src/app/[...gateway]/route.ts
+++ b/src/app/[...gateway]/route.ts
@@ -3,6 +3,7 @@ import { getAll } from "@/lib/config-store";
 import { proxyGatewayRequest, redirectToSetup, serveGatewayHTML } from "@/lib/gateway-proxy";
 import { isGatewayStaticPath } from "@/lib/gateway-static";
 import { readEdition } from "@/lib/edition-source";
+import { clawboxNamespaceKind } from "@/lib/clawbox-namespaces";
 
 export const dynamic = "force-dynamic";
 
@@ -15,6 +16,33 @@ export async function GET(request: NextRequest) {
     if (!config.setup_complete) {
       return redirectToSetup(request);
     }
+    // A path in ClawBox's OWN namespaces — /setup-api and /login-api for its
+    // endpoints, /setup, /login, /portal, /updating and /app for its pages —
+    // that reached this catch-all is a route that does not exist, and the
+    // gateway serves nothing there. The honest answer is 404: proxied, it came
+    // back 502 from a gateway that had never heard of it, and answered as a
+    // NAVIGATION it came back as the Control UI shell with the gateway token
+    // injected into it — somebody else's application, with a credential in it,
+    // over a ClawBox address the owner mistyped (TASK-631).
+    //
+    // ABOVE the Hermes gate on purpose. That gate answers text/plain for every
+    // path, so leaving this below it made `/setup-api/nope` a JSON 404 on
+    // OpenClaw and a text one on Hermes — a shared surface answering two
+    // shapes by edition.
+    //
+    // The list and the segment-boundary matching live in
+    // src/lib/clawbox-namespaces.ts, beside the evidence.
+    const owned = clawboxNamespaceKind(request.nextUrl.pathname);
+    if (owned) {
+      // Code asked, so code is answered: the endpoint namespaces get the same
+      // `{ error: "Not found" }` shape the real routes under them use.
+      return owned === "api"
+        ? NextResponse.json({ error: "Not found" }, { status: 404 })
+        : new NextResponse("Not found", {
+          status: 404,
+          headers: { "Content-Type": "text/plain" },
+        });
+    }
     // On the Hermes SKU there is no OpenClaw gateway — it is disabled and
     // masked by install.sh — so proxying to 127.0.0.1:18789 is a guaranteed
     // ECONNREFUSED. This route matches EVERY otherwise-unhandled path (it wins
@@ -22,19 +50,6 @@ export async function GET(request: NextRequest) {
     // alone was not enough), so without this check /chat, /sessions and every
     // typo answered with a 500 from the catch below instead of a plain 404.
     if (readEdition() === "hermes") {
-      return new NextResponse("Not found", {
-        status: 404,
-        headers: { "Content-Type": "text/plain" },
-      });
-    }
-    // /setup-api/ is ClawBox's OWN namespace — the route handlers live there
-    // precisely so they cannot collide with the gateway's /api/* — and the
-    // gateway serves nothing under it. A /setup-api path that reached this
-    // catch-all is therefore a route that does not exist, and the honest
-    // answer is 404: proxied, it came back 502 from a gateway that had never
-    // heard of it (or the SPA shell, before the resource rule below), and a
-    // client probing for an endpoint could not tell "not here" from "down".
-    if (request.nextUrl.pathname.startsWith("/setup-api/")) {
       return new NextResponse("Not found", {
         status: 404,
         headers: { "Content-Type": "text/plain" },
--- a/src/lib/clawbox-namespaces.ts
+++ b/src/lib/clawbox-namespaces.ts
@@ -1,0 +1,92 @@
+/**
+ * Which top-level paths belong to ClawBox rather than to the OpenClaw gateway.
+ *
+ * WHY THIS EXISTS. `src/app/[...gateway]/route.ts` answers every path Next did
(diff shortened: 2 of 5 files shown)


KrasimirKralev:
Merged into beta (dd058938) after final review MERGE-OK on the merged head and green CI on that head. Deploy of beta to both boxes follows; the box proof will be recorded on TASK-631.

PR #682, opened 2026-09-05
fix: say what was actually removed, and confirm a Hermes skill removal at all

KrasimirKralev:
**TASK-544** — Hermes: uninstalling a desktop app claims a SKILL was removed.

## Customer-visible symptom
Removing a tile from the desktop posted, **as the owner's own chat bubble**,
`I just removed the "<id>" skill. Can you confirm it is gone?` — and the agent confirmed it.

On Hermes that was wrong 100% of the time. `isInstalledAppVisible` hides every non-webapp there,
so a **webapp** is the only installed app an owner can remove, a webapp is not a skill, and "skill"
is a live separate concept on that edition with its own store and its own `skill_uninstall`. An id
collision had the agent confirm the removal of a skill that was still installed. The same message is
wrong on OpenClaw for a `webapp_create` tile, for the same reason.

Inverted the other way: `HermesSkillsStore` — the one surface on that edition where a **real** skill
is installed or removed — dispatched nothing, so removing a real skill produced no confirmation at
all.

## Cause
`src/lib/skill-change-message.ts` had one noun for every branch, and `src/app/page.tsx`'s uninstall
sent `{action, id}` with nothing to say what had been removed. The confirmation flow was wired to
the OpenClaw App Store and never extended to the Hermes one.

## Harness-first
Nothing native is re-implemented. The confirmation is deliberately the agent's own answer rather
than a claim by the UI (the module's docblock: nothing restarts, so the cheapest honest confirmation
is the agent saying so itself), and it is checked against the harness's own listings — `ui_list_apps`
for a desktop app, `skill_list` / `skill_uninstall` for a Hermes skill. What decides the wording is
ClawBox's own record of what the thing is (`installed_meta[<id>].webappUrl`) and the uninstall
route's own `skillRemoved` field, both of which already existed and were being thrown away.

## What changed
- `SkillChangeEvent` gains `kind: "skill" | "app"`. `installedAppKind(meta)` derives it from the
  webapp URL — edition-independent, so it fixes the OpenClaw `webapp_create` case too — and
  `installedAppRemovedDetail(appId, meta, skillRemoved)` builds the whole detail.
- **Both strings travel.** The message carries the display name the owner read *and* the id the
  agent's lists actually report: `ui_list_apps` emits an installed app as
  `{ id: "installed-<id>", name: <id> }` and `skill_list` leads with the lock id, so a sentence with
  only the display name asks the agent to verify against a string neither list contains — and
  "confirmed, it's gone" would be a guess dressed as verification.
- **One id can be both.** `webapp_create` replaces `installed_meta[<id>]` for any id with no
  collision check, and the uninstall route removes the webapp *and* a store skill of that name,
  reporting which in `skillRemoved`. The desktop already parsed that body for `skillHalfChecked` and
  dropped this field; it is now carried, and the line says the skill went too rather than hiding a
  capability the owner just lost.
- `HermesSkillsStore` announces install and uninstall — on the success paths only. Both calls sit
  after the `!res.ok` throw and after the `rollback_incomplete` / `removal_incomplete` /
  `incomplete_install` / `dangerous_skill` branches; the uninstall route answers 200 only when both
  the lock entry and the directory are verified gone, and `too_large` / `uninstall_unproven` are
  502s. The removal announcement carries the card's **display name** with the lock key beside it —
  the install path already sent the display name, and the same skill under two names in one
  transcript is the defect this card is about, one surface up.
- The event name is now one exported constant (`SKILL_CHANGE_EVENT`) with one dispatcher
  (`announceSkillChange`), used by all four senders and both listeners. A magic string four senders
  agreed on is how "skill-installed" came to be dispatched for an app uninstall.
- `ChatPopup`'s handler casts the detail to the shared `SkillChangeEvent` instead of restating three
  of its fields — the local cast omitted `kind`, which survived only because the object was passed
  on by reference, one destructure away from reverting the whole card with every suite green.

## Both editions
- Hermes: a webapp removal says "app"; a real skill removal now confirms at all, as a skill.
- OpenClaw: a `webapp_create` tile removal says "app" (it never should have said skill); a store
  skill still says skill; `AppStore` passes `kind: 'skill'` explicitly (`store` is OpenClaw-only and
  `handleInstallApp` never records a `webappUrl`).
- `InstalledAppSettings`'s enable/disable leaves `kind` at its default: it writes openclaw.json's
  skill enable flag, and a webapp renders as `type: "webapp"` and never reaches that panel.

## RED → GREEN
Two suites. Against the unmodified beta sources
(`git checkout origin/beta -- src/lib/skill-change-message.ts src/components/HermesSkillsStore.tsx
src/app/page.tsx src/components/AppStore.tsx src/components/InstalledAppSettings.tsx
src/components/ChatPopup.tsx`):

```
 ❯ |unit| src/tests/unit/skill-change-message.test.ts (23 tests | 11 failed)
     × calls a removed webapp an app, and points at the desktop
     × carries the id the agent's own lists report, beside the name the owner read
     × says the skill went too when one id was both
     × names an app an app on install too
     × names an app an app on enable too
     × names an app an app on disable too
     × calls a webapp an app
     × calls everything else a skill — including an app whose meta the desktop lost
     × sends the name off the tile the owner clicked, and the slug beside it
     × carries the route's skillRemoved only when it really said yes
     × falls back to the id when there is no meta to read a name from

 FAIL … > calls a removed webapp an app, and points at the desktop
 AssertionError: expected 'I just removed the "Pomodoro" skill. …' to match /app/
 + Received: "I just removed the \"Pomodoro\" skill. Can you confirm it is gone?"
```

```
 ❯ |components| src/tests/components/hermes-skills-chat-confirmation.test.tsx (5 tests | 2 failed)
     × announces the removal, as a skill, with the name the owner saw
     × announces the install, as a skill

 ❯ |components| src/tests/components/chat-skill-install-no-freeze.test.tsx (8 tests | 1 failed)
     × calls a removed webapp an app, all the way to the wire
```

With the fix: `skill-change-message` 23/23, `hermes-skills-chat-confirmation` 5/5,
`chat-skill-install-no-freeze` 8/8 (two new cases drive the whole path
dispatch → ChatPopup → the sent frame, which is the one hop `kind` has to survive).

Neighbouring component suites — the two above plus `app-store-errors`,
`app-store-install-modal`, `installed-app-settings`, `hermes-skills-store-errors`,
`hermes-skills-leftover-refresh`, `hermes-skills-warming`,
`hermes-skills-preexisting-incomplete`, `hermes-skill-docs-loading`: 83 passed.

Whole unit project: 650 files, 10500 passed. (Two unhandled rejections from
`use-clawbox-login.test.ts`, a teardown timer in a file this diff does not touch.)

Whole `--project components`: 1361 passed, 3 failed in `chat-spoken-reply.test.tsx` and
`chat-email-refs-surfaces.test.tsx` — both known flakes under a full parallel run, neither naming
any module in this diff, and 26/26 green when the two are run together on their own.

`tsc --noEmit` clean; `eslint` on the seven changed files, 0 errors. The
`state-updater-purity` guard passes: every `announceSkillChange` call is a plain statement, none
inside a state updater.

## Sibling call sites — handled or cleared
- `src/components/AppStore.tsx` and `src/components/InstalledAppSettings.tsx` — converted to the
  shared dispatcher; their `kind` decided above.
- `src/app/page.tsx`'s `register_webapp` path announces nothing, correctly: the agent created the
  webapp itself and does not need to be told.
- `src/tests/components/chat-skill-install-no-freeze.test.tsx` still dispatches the literal event
  name; `SKILL_CHANGE_EVENT` is that exact string, so it is equivalent, and the file now also
  covers `kind` end to end.

## Recorded, not claimed
- **`/app/hermes-skills` (the "Open in new tab" window) mounts no chat surface**, so the new
  announcement goes nowhere there — only `case "clawbox"` renders a chat, and that component does
  not listen for this event. Pre-existing for the App Store and the settings toggle too; the
  confirmation is closed for the desktop window, not for the standalone one.
- **The chat opens on every announcement** (`page.tsx`'s other listener is
  `() => setChatOpen(true)`), so installing or removing a Hermes skill now pops the chat over the
  Skills window and queues one agent turn. That is exactly what the OpenClaw App Store already does
  and there is no batch path to storm it (one confirm dialog per action), but it is new behaviour on
  that edition and worth a look on the box.
- **The four sentences are English-only.** Pre-existing in this module, but `HermesSkillsStore` is
  localized end to end, so this PR makes a localized surface emit an untranslated line as the
  owner's own bubble — which also steers the agent's reply into English for a non-English owner.
  Routing them through `t()` is a separate change.
- **`confirmUninstallApp` still has no re-entry guard** — a double-click sends two POSTs. Refuted
  in-thread on #637 and already on the residual list; untouched here.

## Summary by CodeRabbit

- **New Features**
  - Added clearer notifications for app and skill installations, enabling, disabling, and removals.
  - Notifications now distinguish between apps and skills and include relevant names or identifiers.
  - App removal messages can indicate when an associated skill was removed as well.

- **Bug Fixes**
  - Improved consistency of skill-change updates across the app, store, and chat interfaces.
  - Prevented incorrect item types from appearing in removal messages.

### DIFF
--- a/src/app/page.tsx
+++ b/src/app/page.tsx
@@ -1060,8 +1060,9 @@ function ChromeDesktopInner() {
 
   // Same reason, for the same callback: it needs the removed app's meta to say
   // WHAT it removed, and taking `installedMeta` as a dependency would rebuild
-  // the callback on every install. The meta is read after the route confirms
-  // the removal, before the state that holds it is pruned below.
+  // the callback on every install. Captured before the request is issued, so
+  // neither the prune below nor a preferences reload in flight can take it
+  // away.
   const installedMetaRef = useRef<Record<string, InstalledMeta>>({});
   useEffect(() => {
     installedMetaRef.current = installedMeta;
@@ -1085,6 +1086,10 @@ function ChromeDesktopInner() {
     // is safe: a second uninstall of an app already gone answers `ok:true`.
     const controller = new AbortController();
     const timer = setTimeout(() => controller.abort(), 10_000);
+    // Read inside the try, said outside it: the route reports whether a store
+    // skill of the same id went with the tile, and the sentence that mentions
+    // it is dispatched after the state cleanup below.
+    let skillRemoved: boolean | null | undefined;
     try {
       const res = await fetch("/setup-api/apps/uninstall", {
         method: "POST",
@@ -1107,6 +1112,7 @@ function ChromeDesktopInner() {
       // desktop, so the cleanup below is right — but saying nothing would put
       // this route's own defect back in the one surface the owner watches.
       const removed = await res.json().catch(() => null);
+      skillRemoved = removed?.skillRemoved;
       if (removed?.skillHalfChecked === false) {
         window.dispatchEvent(new CustomEvent("clawbox:toast", {
           detail: {
@@ -1164,7 +1170,10 @@ function ChromeDesktopInner() {
     // live separate concept with its own store, so the old wording sent the
     // agent to check a list the app was never in. `installedAppRemovedDetail`
     // reads the meta captured above, which is still the pre-prune copy.
-    announceSkillChange(installedAppRemovedDetail(appId, removedMeta));
+    // `removed?.skillRemoved` is the route's own answer to "did a store skill
+    // of this id go too" — one id can be both, and a line that mentions only
+    // the tile hides a capability the owner has just lost.
+    announceSkillChange(installedAppRemovedDetail(appId, removedMeta, skillRemoved));
   }, []);
 
   // Get all apps including installed ones
--- a/src/app/page.tsx
+++ b/src/app/page.tsx
@@ -41,6 +41,7 @@ import { cleanVersion } from "@/lib/version-utils";
 import { fetchHarness } from "@/lib/client-harness";
 import { samePairingToken } from "@/lib/telegram-pairing-token";
 import type { InstalledMeta } from "@/lib/store-categories";
+import { SKILL_CHANGE_EVENT, announceSkillChange, installedAppRemovedDetail } from "@/lib/skill-change-message";
 import { apps, type AppDef } from "@/lib/desktop-apps";
 import { hiddenAppIdsForHarness, isInstalledAppVisible } from "@/lib/desktop-app-editions";
 import {
@@ -615,14 +616,14 @@ function ChromeDesktopInner() {
   // the user can watch the agent's response.
   useEffect(() => {
     const handler = () => setChatOpen(true);
-    window.addEventListener('clawbox-skill-installed', handler);
+    window.addEventListener(SKILL_CHANGE_EVENT, handler);
     window.addEventListener(FIX_ERROR_EVENT, handler);
     window.addEventListener(CHAT_MESSAGE_EVENT, handler);
     // The Coding Agent's "Create app" button: the chat has to be open before
     // the card inside it can be seen.
     window.addEventListener(NEW_APP_EVENT, handler);
     return () => {
-      window.removeEventListener('clawbox-skill-installed', handler);
+      window.removeEventListener(SKILL_CHANGE_EVENT, handler);
       window.removeEventListener(FIX_ERROR_EVENT, handler);
       window.removeEventListener(CHAT_MESSAGE_EVENT, handler);
       window.removeEventListener(NEW_APP_EVENT, handler);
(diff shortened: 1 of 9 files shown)


KrasimirKralev:
**Device leg: declared gap, with what was checked.**

Read-only on both boxes (no mutation, no lock taken), the uninstall path could not be exercised: neither has an installed app to remove.

```
Hermes box  (CLAWBOX_EDITION=hermes)   pref:installed_apps -> None,  installed_meta -> {}
OpenClaw box (CLAWBOX_EDITION=openclaw) pref:installed_apps -> None,  installed_meta -> {}
```

Producing one means `webapp_create` or a store install — a mutation, and this batch is read-only — so the wording is proved by the suites rather than by the hardware: 23 unit cases over `buildSkillChangeMessage` / `installedAppKind` / `installedAppRemovedDetail`, 5 component cases driving `HermesSkillsStore` through the real confirm dialog (including three that assert **silence** over `removal_incomplete`, `uninstall_unproven` and `cli_timeout`), and 2 more driving dispatch → `ChatPopup` → the sent frame, which is the hop `kind` has to survive.

What the boxes DID confirm is the premise the wording rests on: the Hermes box is a locked `hermes` SKU with the gateway masked and no `openclaw` binary, so `isInstalledAppVisible` really does leave a webapp as the only installed app an owner can remove there — and "skill" really is a separate live concept on it (`hermes --version` -> `Hermes Agent v0.20.5`, its own skills store behind `skill_search`/`skill_install`).

Worth a click when this lands: create a webapp on the Hermes box, remove its tile, and read the chat bubble — it should say **app**, name the tile, carry the id, and the agent should answer from `ui_list_apps`.

PR #685, opened 2026-09-05
fix: let a waiting client force one catalogue retry per backoff window

KrasimirKralev:
Closes TASK-669 (the #587 follow-up M2).

## Symptom

The box's boot warmup fails a model enumeration — the network is not up yet, the CLI is mid-upgrade — while nobody is looking. The failure records a backoff that starts at two minutes and doubles up to the full six-hour refresh interval. The first person to open the model picker afterwards is shown the curated fallback list, and **nothing on the box is willing to try again**. Before the change generations landed, any picker open re-forked.

## Two things the card names, neither of which is the mechanism

Both were reproduced before anything was written, and both are recorded in the code rather than left for the next reader:

* **The generation counters defaulting to 0 does make `serveCurrent` a structural no-op on a cold process — and it is not what bites.** With a failed warmup the GET's own freshness test is true (`refreshFailedLast`, and `ageMs = Infinity` with no cache), so it takes the `isStale || !isLivePayload` branch and never reaches `serveCurrent` at all. `refreshIsBlocked` is the blocker.
* **`?refresh=1` is not the owner's force button.** `src/hooks/useProviderCatalog.ts` sends it only from `forceNextLoad`, i.e. as the client's echo of a `PROVIDERS_CHANGED` event — by which point `notifyProviderSetChanged` has already cleared the wait server-side. The plain picker open, the request that actually finds the box in this state, sends no parameter at all.

So the attempt belongs to the request, not to a query parameter.

## Fix

The request that finds itself **without the box's own current answer** spends one attempt against a recorded failure.

* **"Without the box's own current answer" is the question the GET has already answered**, three lines above the call, and is passed in rather than re-derived: `!isLivePayload(cached) || servedEmpty || ageMs > REFRESH_INTERVAL_MS`. That covers two states a `!isLivePayload` test alone gets wrong — a payload stamped `live` whose every row the current sanitiser rules filter away (the state that arrives the first time a filter is tightened, which `catalog-sanitised-empty-cache.test.ts` exists to pin), and a real enumeration from three weeks ago on a box that was switched off. In both the owner is looking at the curated rows. Deliberately **not** `isStale`, which includes `refreshFailedLast` and is therefore true by construction whenever the wait is running.
* **Rationed by a minimum gap, not a once-ever count.** This process runs for weeks. A single allowance would be burnt by the first picker open after the blip — very likely the one where the uplink is still coming up — and the provider would then be unreachable for the life of the server, which is the probe-once class this repo keeps producing. `src/lib/hermes-model-options.ts` solved the same problem on the other edition with `EXPLICIT_REFRESH_MIN_GAP_MS`; this is that primitive. The gap is longer than an enumeration (~3 minutes on a Jetson) so a mounted picker's two-second warming poll cannot chain one fork into the next.
* **Nothing is spent while another enumeration is running.** `refreshing` is per-provider, so apart from `bootWarmup`'s 5 s stagger it is the only fork-concurrency discipline in the file — and the wizard's provider list re-runs `useProviderCatalog` on every click, so a walk down anthropic → openai → google could otherwise start three ~2-core, ~3-minute forks at once on an Orin. Leaving it unspent is also the honest answer: something IS on its way, and the response says `warming`.
* **The wait is LIFTED, not cleared.** `clearFailedRefresh` deletes the record, so `recordFailedRefresh` restarts the doubling from its two-minute floor: a provider that had climbed to the six-hour cap would re-climb it through eight more picker-driven forks, and `refreshFailedLast` would go false for the length of the retry, dropping the `stale` flag off a payload no device produced. Expiring the deadline while keeping the window it had reached does neither.
* **Both the gap and the wait are dropped** by a successful refresh and by a provider-set change — the two events after which an old failure describes a box that no longer exists, which is the doctrine `clearFailedRefresh` already followed.
* **Not for a provider that can never enumerate here.** `cliMissing` (the Hermes SKU) and the no-enumeration-on-this-core case both `recordFailedRefresh` precisely so their one log line appears once per backoff window instead of once per picker open; `changeCouldMakeItEnumerable` is the question they already answer, and the new branch is behind it.

## Harness first

Nothing here re-implements a harness mechanism: the enumeration IS `openclaw models list --provider <id> --all --json`, and this change only decides when ClawBox is allowed to ask again. The scheduling primitive is borrowed from this repo's own Hermes-side answer to the same question rather than invented.

## Both editions

On Hermes `openclawIsAbsent()` makes `cliMissing` true for every CLI-enumerated provider, so `changeCouldMakeItEnumerable` is false and the retry never fires — which is required, because that branch's `recordFailedRefresh` is what keeps its "no CLI on this edition" log line to once per window.

## RED

On unmodified `origin/beta`, with only the test file copied in:

```
⎯⎯⎯⎯⎯⎯⎯ Failed Tests 3 ⎯⎯⎯⎯⎯⎯⎯
 FAIL  catalog-forced-retry.test.ts > re-enumerates ONCE when a picker opens over a failed warmup
 FAIL  catalog-forced-retry.test.ts > rations the attempt, so a polling picker cannot chain forks
 FAIL  catalog-forced-retry.test.ts > counts a LIVE list older than the refresh interval as not current
AssertionError: expected 1 to be 2 // Object.is equality
 Test Files  1 failed (1)
      Tests  3 failed | 1 passed (4)
```

`1` is the boot warmup's own attempt and nothing after it. The fourth case — a provider with a live, current catalogue is not owed a fork because a picker mounted — passes on beta too; it is the boundary guard, not evidence.

## GREEN

`catalog-forced-retry.test.ts` 4 passed; all 18 `routes/ai-models` suites 330 passed on the rebased head; full `--project unit` 658 files / 10546 passed. `tsc --noEmit` and `eslint` clean.

## Overlap

`src/app/setup-api/ai-models/catalog/route.ts` is also touched by the TASK-668 branch, which adds an exported `notifyModelsModeChanged` in a different region of the same file (the notify block, not the refresh gate). Both are based on the same `origin/beta`; whichever merges second may want a trivial rebase.

## Review

`clawbox-review`, findings at `code-review-669.md` (1 HIGH, 6 MEDIUM, 4 LOW). The departure from the card's stated fix direction was independently reproduced and upheld. Applied: HIGH-1 (the one-shot `Set` was a probe-once decision held for a process that runs for weeks — now a minimum gap), MEDIUM-1 (the retry was asking a narrower "is there a live payload" question than the GET had already answered), MEDIUM-2 (the allowance is renewed by a provider-set change and by a successful refresh), MEDIUM-3 (lift the deadline instead of clearing the record, and fix the two comments that claimed it), MEDIUM-4 (nothing spent while another fork runs), MEDIUM-5 (a discriminating test — the live-but-three-weeks-old cache — so the condition cannot be deleted with the suite still green).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved AI model catalog recovery when provider responses are missing, empty, or outdated.
  * Limited client-triggered retries to one attempt every five minutes, while preserving exponential backoff.
  * Retry state now resets after successful updates or provider changes, preventing unnecessary repeated refreshes.

* **Tests**
  * Added coverage for transient provider failures, stale catalogs, repeated polling, provider changes, and avoiding unnecessary retries for current catalogs.

### DIFF
--- a/src/app/setup-api/ai-models/catalog/route.ts
+++ b/src/app/setup-api/ai-models/catalog/route.ts
@@ -220,6 +220,47 @@ const FAILED_REFRESH_BACKOFF_MS = 2 * 60_000;
 const changeSeq = new Map<string, number>();
 const publishedSeq = new Map<string, number>();
 
+/**
+ * When each provider last spent a client-facing retry.
+ *
+ * TASK-669. The failed-refresh backoff starts at two minutes and doubles to
+ * the full six-hour interval, which is the right schedule for the BOX's own
+ * retries and was also the only schedule there was. A blip during boot warmup
+ * — the network not up yet, the CLI mid-upgrade — records a wait while nobody
+ * is looking, and the first person to open the picker afterwards is shown the
+ * curated fallback with nothing on the box willing to try again. Before the
+ * change generations landed, any picker open re-forked; now nothing does.
+ *
+ * `?refresh=1` is not the answer, and that is worth writing down because the
+ * card proposed it: the client sends it ONLY as its echo of a provider-set
+ * change (`useProviderCatalog`'s `forceNextLoad`), and by then
+ * `notifyProviderSetChanged` has already cleared the wait server-side. The
+ * plain picker open — the one that finds the box in this state — sends no
+ * parameter at all.
+ *
+ * So the attempt is spent by the REQUEST that finds itself without the box's
+ * own current answer, and it is a MINIMUM GAP rather than a once-ever count.
+ * This process runs for weeks: a single allowance would be burnt by the first
+ * picker open after the boot blip — very likely the one where the uplink is
+ * still coming up — and the provider would then be unreachable for the life of
+ * the server. `src/lib/hermes-model-options.ts` had the same problem on the
+ * other edition and solved it with `EXPLICIT_REFRESH_MIN_GAP_MS`; this is that
+ * primitive.
+ *
+ * The gap is longer than an enumeration (~3 minutes on a Jetson) so a mounted
+ * picker's warming poll cannot chain one fork into the next, and it is cleared
+ * by the two events that make an old failure meaningless — a successful
+ * refresh, and a provider-set change.
+ */
+const CLIENT_RETRY_MIN_GAP_MS = 5 * 60_000;
+const clientRetryAt = new Map<string, number>();
+
+/** May a client-facing retry be spent for this provider now? */
+function clientRetryIsDue(provider: string): boolean {
+  const last = clientRetryAt.get(provider);
+  return last === undefined || Date.now() - last >= CLIENT_RETRY_MIN_GAP_MS;
+}
+
 function currentSeq(provider: string): number {
   return changeSeq.get(provider) ?? 0;
 }
@@ -258,6 +299,10 @@ function recordFailedRefresh(provider: string): void {
 
 function recordSuccessfulRefresh(provider: string): void {
   failedRefreshes.delete(provider);
+  // The gap only exists to ration retries against a FAILING provider. One that
+  // has just answered owes nothing, and the next time it fails the client's
+  // first look should get an attempt.
+  clientRetryAt.delete(provider);
 }
 
 /**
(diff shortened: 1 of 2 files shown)


KrasimirKralev:
Addressed in `a721559f` (rebased head `3cb1988d`).

**`catalog/route.ts:1466` — include the generation mismatch.** Fixed, and it is the sharpest case in the PR: after a provider-set change whose refresh failed, the payload on file is seconds old and stamped `live`, so age and `isLivePayload` both say "current" while the generation says otherwise. `notifyProviderSetChanged` cleared the wait and forked; when that fork fails the wait is back and every later request is blocked — so the picker sat on the PRE-change list for the whole window, which is the shape this card is about with a live list instead of a curated one.

`|| !publishedIsCurrent(provider)` added to the expression, with the reason in a comment beside it.

Regression test added — "counts a LIVE list from before a provider-set change as not current": warm up successfully, make the CLI fail, `notifyProviderSetChanged("anthropic")` (which forks once and fails), then a picker open, and assert exactly one further enumeration. It fails without the new term.

GREEN on the rebased head: all 18 `routes/ai-models` suites 331 tests, `--project unit` 657 of 658 files — the one failure is `src/tests/unit/run-tunnel.test.ts`, which this PR neither touches nor imports, passes on its own, and is on this repo's known-flake list.

KrasimirKralev:
Merged into beta (dd058938) after final review MERGE-OK on the merged head and green CI on that head. Deploy of beta to both boxes follows; the box proof will be recorded on TASK-669.

PR #694, opened 2026-09-05
fix: keep the checkout .env and .git out of the standalone build

KrasimirKralev:
**TASK-692** — the standalone build ships the checkout's `.env` and `.git`.

## Symptom

Every build writes a byte copy of `$PROJECT_DIR/.env` — the 0600 file systemd hands
`clawbox-setup` as an `EnvironmentFile`, holding `GOOGLE_OAUTH_CLIENT_SECRET` and, where
`install.sh` was given one, `CLAWBOX_AI_API_KEY` — into `.next/standalone/`, together with
`.env.example` and a full copy of `.git`. A secrets-shaped file gets a second home inside a
build artifact that is re-made by every update, copied by `set_previous_build_aside`, and
included in anything that archives the build. `.git` also costs 88-103 MB of eMMC per build.

Measured read-only on both editions, 2026-09-05:

| | OpenClaw box | Hermes box |
|---|---|---|
| `.next/standalone/.env` | present, 0600, 2105 bytes | present, 0600, 2105 bytes |
| byte-identical to `$PROJECT_DIR/.env` | `cmp` says yes | `cmp` says yes |
| `.next/standalone/.git` | 88 MB | 103 MB |
| `EnvironmentFile=` on clawbox-setup | `/home/clawbox/clawbox/.env` | `/home/clawbox/clawbox/.env` |

(The Hermes box has since gone into a rebuild that has been running 2.5 h and currently has no
`.next/standalone` at all — noted for the driver, unrelated to this PR.)

## Harness first — the native mechanism, corrected

**An earlier version of this PR said `outputFileTracingExcludes` reaches neither file. That is
right for `.env` and wrong for `.git`, and the review caught it.** Read out of the installed
`next@16.3.3` in this checkout:

* `.env` and `.env.production` are **not traced at all**, so no config key can reach them.
  `writeStandaloneDirectory()` walks `loadedEnvFiles`, filters to exactly those two names and
  `copyFile()`s them *after* `copyTracedFiles` (`next/dist/build/index.js`). The sweep is the
  only thing that removes those — the native mechanism genuinely does not exist here.
* `.git` arrives on the **instrumentation** trace, and that trace **is** reachable.
  `next-trace-entrypoints-plugin.js` keys `entryNameFilesMap` by `entrypoint.name` for *every*
  server-compiler entrypoint, not only route entries, so `instrumentation` has an entry;
  `collect-build-traces.js` iterates that map and applies these globs to each `.nft.json`,
  skipping only static pages and `edgeRuntimeRoutes` — which is exactly why MIDDLEWARE keeps its
  `data/` entry (the observation already recorded in `next.config.ts`) while the server-compiler
  traces are cleaned. Checked with the vendored matcher:
  `picomatch("*", { dot: true, contains: true })("instrumentation")` is `true`, and `.git/**`
  matches `.git/objects/…`. `build/utils.js` copies the standalone tree *from* those lists.

So `".git/**"` is added to `outputFileTracingExcludes["*"]` — the harness's own switch, used.
**Not measured:** a real `next build` on a box with that line in place; that is device work.
The postbuild sweep therefore stays and still FAILS the build if a copy survives, which is the
post-condition the suites pin — the exclude is meant to make that sweep find nothing.

So the mechanism this repo already owns is the right one: the `postbuild` guard that F-27
(#592) built for `data/` and TASK-725 (#672) moved into `scripts/postbuild.sh`. This extends
it rather than adding a second thing.

## The fix

`scripts/postbuild.sh`, right after the `data/` block: sweep `.env`, `.env.*` and `.git` out
of the standalone tree and `fail` when one survives.

* **Depth-unbounded**, because that trace copies whole project subdirectories.
  `e2e-install/.env.test` is gitignored and its tracked `.example` documents seven provider
  keys; it lands at `.next/standalone/e2e-install/.env.test`, two levels down, where a
  top-level sweep would have shipped it and reported the build clean.
* **`node_modules` pruned** — packages ship their own `.env` fixtures, and failing a build
  over somebody else's test data would be a false failure.
* **`|| true` on the removal** so the named check reports it, rather than `errexit` killing
  the script on `rm` with nothing said — the shape the `data/` removal above already uses.
* **`-print -quit`, not `| head -n 1`** — under `pipefail` a `find` still writing when `head`
  exits dies of SIGPIPE and would end the build silently.

## Why removing them is safe

The standalone copy is byte-identical to the checkout file (`cmp`, both boxes); systemd loads
the checkout file into the service environment; and `@next/env`'s `processEnv` only fills keys
**absent** from its initial `process.env` snapshot, so the copy could never have overridden
systemd. Nothing in the repo reads `.next/standalone/.env` or `.next/standalone/.git`
(`build-identity.ts` pins `PROJECT_ROOT` to `CLAWBOX_ROOT` or the absolute default, never the
cwd).

One behaviour does go, deliberately: `playwright.config.ts`'s `webServer` runs
`node .next/standalone/server.js` with no systemd, and Next's runtime `loadEnvConfig(this.dir)`
used to pick a developer's checkout `.env` up through the copy. CI never had one (gitignored,
never created by `e2e-tests.yml`), so local runs now match CI. If a spec ever needs a variable,
pass it on the `webServer` command rather than restoring the copy.

## The blocker from the first merge gate, fixed

The `.env`/`.git` sweep added two unconditional `find` calls, and
`src/tests/unit/postbuild-standalone-entry.test.ts` — the OTHER test file for this same script,
already on beta — puts a stub `find` on `PATH` that ignores its arguments and answers every call
with a parked build's entry. Both new calls got that answer, `$LEFT` was non-empty, and the guard
failed the build: two deterministic failures, and the `test` check red.

RED on the previous head:

```
 FAIL  src/tests/unit/postbuild-standalone-entry.test.ts > stamps the entry next build wrote…
 FAIL  src/tests/unit/postbuild-standalone-entry.test.ts > takes the entry from the path Next recorded for it
AssertionError: postbuild: .next/standalone/.next-old/standalone/server.js survived removal …
 Test Files  1 failed (1)   Tests  2 failed | 5 passed (7)
```

The stub is now argument-aware: it answers only the call naming `server.js` — the ENTRY lookup it
exists to pin — and `exec`s the real `find` for everything else, because entry *selection* does
not need `find` and the sweep does. The negative proof it carries still bites: forcing the entry
lookup down the `find` branch (`if false` on both preceding conditions) makes both cases fail
again with `…/.next-old/standalone/server.js is not a standalone entry`.

```
 Test Files  5 passed (5)
      Tests  36 passed (36)     (both postbuild suites + the three shared guard suites)
tsc --noEmit  exit 0
```

## RED → GREEN

RED, on unmodified `origin/beta` (d428b044):

```
 ❯ src/tests/unit/postbuild-standalone-data.test.ts (9 tests | 4 failed)
   × removes every copy of the checkout's .env, wherever the trace put it
   × removes the copied .git
   × fails the build, naming the path, when the .env copy survives the removal (non-root uid only)
   × removes data/ and still links server.js in the nested standalone layout

AssertionError: expected true to be false
 ❯ src/tests/unit/postbuild-standalone-data.test.ts:174
    expect(fs.existsSync(path.join(standalone, ".env"))).toBe(false);
```

GREEN, with the fix, plus the neighbouring suites and the three beta guards
(`openclaw-config-mock-completeness`, `test-timeout-hygiene`, `state-updater-purity`):

```
 Test Files  6 passed (6)
      Tests  79 passed (79)
```

The cases run the real `scripts/postbuild.sh` against a temp tree, so they cannot drift from
what ships. One of them pins that a `node_modules` `.env` fixture SURVIVES, and the
failure-mode case asserts on the guard's own line — `rm` writes its own "Permission denied"
naming the same path onto the same stream, so an assertion on the path alone would pass with
the guard silent.

## Sibling call sites

Every build entry point goes through `bun run build`, which runs `postbuild`: `install.sh`
`step_build` and `do_rebuild`, `install-x64.sh`, `scripts/force-update.sh`,
`.github/workflows/build-identity.yml`, `playwright.config.ts`. There is no direct
`next build` caller in the repo. No reader of the standalone `.env`/`.git` exists.
`.dockerignore` already keeps `.env*` out of the e2e-install build context.

## Both editions

The guard is edition-blind — it runs in the one build step both SKUs use — and the defect was
measured on both boxes (table above).

## Review

**Second pass, on the head that fixed the first pass's blocker.** It found three things, all
applied here:

* The `.git` half of the harness-first claim was wrong (above) — the exclude is now used and the
  four comments that stated the opposite are corrected.
* The `find` stub added in the previous commit `exec`s whatever `command -v find` prints, and
  `command -v` prints a BARE NAME when `find` is a shell function — which makes the stub `exec`
  itself forever inside `spawnSync`, where the vitest timeout cannot interrupt it. Reproduced at
  108,894 invocations. `CAN_RUN` now requires an absolute path, so such a host skips the suite.
* The corrected `scripts/` comment had dropped `hermes-image-plugin.ts`, which is still a cwd
  reader on beta until #697 lands. Both readers are named again, with #697 noted.

Also from that pass: every `-name` pattern in the sweep is quoted (this script runs with the
project root as cwd, where `.env` and `.git` both exist), and the entry stub's docblock records
what its narrowing does not cover.

**First pass**, findings applied: the depth-unbounded sweep with the `node_modules`
prune, `|| true` on the removal, `-quit` instead of `| head -n 1`, the generic failure
message, the mis-aimed assertion in the failure-mode case, and folding the `next.config.ts`
note into the existing TASK-725 block instead of adding a second one. Its blocker — that
`origin/beta` had replaced the inline `postbuild` one-liner with `scripts/postbuild.sh` under
`set -euo pipefail` — is why this was re-authored from `origin/beta` rather than rebased
through a conflict.

Not done here: nothing narrows the instrumentation trace itself, which is what puts the whole
project root in there. `next.config.ts` records why (parts of the sweep are load-bearing) and
that is TASK-670's neighbourhood.

## Summary by CodeRabbit

* **Security & Reliability**
  * Standalone build artifacts now remove copied checkout environment files and `.git` directories, including deeply nested instances.
  * The postbuild process reports an error when sensitive files or directories cannot be removed.
  * Package dependencies remain intact during cleanup.

* **Documentation**
  * Documented environment-file handling and secret removal for standalone deployments.

### DIFF
--- a/next.config.ts
+++ b/next.config.ts
@@ -73,6 +73,18 @@ const nextConfig: NextConfig = {
   // for this build's entry. Narrowing the sweep is a separate change: parts of
   // it are load-bearing today (`.next/standalone/scripts` comes from it, and
   // system-profile.ts resolves scripts/ from the process cwd).
+  //
+  // The `.git` in that list is the second thing scripts/postbuild.sh sweeps
+  // (TASK-692): 88 MB of it on the OpenClaw box, measured 2026-09-05. Its
+  // companion, the checkout's own `.env`, does not come from the trace at all
+  // and so cannot be excluded here either — Next copies `.env` and
+  // `.env.production` ITSELF, after the trace, in writeStandaloneDirectory()
+  // (next/dist/build/index.js), with no config switch anywhere in that path.
+  // On a box that file is 0600 and holds GOOGLE_OAUTH_CLIENT_SECRET and, where
+  // install.sh was given one, CLAWBOX_AI_API_KEY, so the copy is removed from
+  // the build artifact and its survival fails the build. Nothing loses by it:
+  // systemd hands the real file to clawbox-setup as an EnvironmentFile and
+  // @next/env never overwrites a variable that is already in the environment.
   outputFileTracingExcludes: {
     // No "./" prefix: Next matches these globs relative to the tracing root,
     // and "./data/**" silently matched nothing — the build kept dying on
--- a/next.config.ts
+++ b/next.config.ts
@@ -74,23 +74,45 @@ const nextConfig: NextConfig = {
   // it are load-bearing today (`.next/standalone/scripts` comes from it, and
   // system-profile.ts resolves scripts/ from the process cwd).
   //
-  // The `.git` in that list is the second thing scripts/postbuild.sh sweeps
-  // (TASK-692): 88 MB of it on the OpenClaw box, measured 2026-09-05. Its
-  // companion, the checkout's own `.env`, does not come from the trace at all
-  // and so cannot be excluded here either — Next copies `.env` and
-  // `.env.production` ITSELF, after the trace, in writeStandaloneDirectory()
-  // (next/dist/build/index.js), with no config switch anywhere in that path.
-  // On a box that file is 0600 and holds GOOGLE_OAUTH_CLIENT_SECRET and, where
-  // install.sh was given one, CLAWBOX_AI_API_KEY, so the copy is removed from
-  // the build artifact and its survival fails the build. Nothing loses by it:
-  // systemd hands the real file to clawbox-setup as an EnvironmentFile and
-  // @next/env never overwrites a variable that is already in the environment.
+  // `.git` is in that 6186-file list too — 88 MB of it on the OpenClaw box,
+  // measured 2026-09-05 — and unlike the middleware half it CAN be excluded
+  // here. Read out of the installed next@16.3.3 in this checkout (TASK-692),
+  // because the paragraph above led a review to the opposite conclusion:
+  //
+  //   * next-trace-entrypoints-plugin.js keys `entryNameFilesMap` by
+  //     `entrypoint.name` for EVERY server-compiler entrypoint, not only route
+  //     entries, so `instrumentation` has an entry in it.
+  //   * collect-build-traces.js iterates that map and applies these globs to
+  //     each `.nft.json`, skipping only static pages and `edgeRuntimeRoutes` —
+  //     which is exactly why MIDDLEWARE keeps its `data/` entry and the
+  //     server-compiler traces do not.
+  //   * `picomatch("*", { dot: true, contains: true })("instrumentation")` is
+  //     `true`, and `.git/**` matches `.git/objects/...` under the same
+  //     options. Both checked with the vendored matcher.
+  //   * build/utils.js copies the standalone tree FROM those lists, so an
+  //     entry dropped here is a file never copied.
+  //
+  // So `.git` is excluded at the source rather than deleted afterwards. What
+  // is NOT measured: a real `next build` on a box with this line in place —
+  // that is device work. scripts/postbuild.sh therefore still sweeps `.git`
+  // and still FAILS the build if a copy survives, which is the post-condition
+  // the suites pin; this line is meant to make that sweep find nothing.
+  //
+  // Its companion, the checkout's own `.env`, genuinely has no switch: Next
+  // copies `.env` and `.env.production` ITSELF, AFTER the trace, in
+  // writeStandaloneDirectory() (next/dist/build/index.js) — nothing on that
+  // path reads this config. On a box that file is 0600 and holds
+  // GOOGLE_OAUTH_CLIENT_SECRET and, where install.sh was given one,
+  // CLAWBOX_AI_API_KEY, so the copy is removed from the build artifact and its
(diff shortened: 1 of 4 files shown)


KrasimirKralev:
Merged into beta (7df0583e) after final review MERGE-OK on this head and green CI on this head (e2e-install passed on its rerun; the first attempt was the cross-branch updater-idle signature). Deploy of beta to both boxes follows; box proof lands on TASK-692.

PR #721, opened 2026-09-06
feat: a coding run verifies its work in the browser on the box's own screen

yalexx:
The owner watched a delegated run drive a headless Chromium nobody could see and asked for the one on the screen instead — the same browser the Browser app shows — with a switch to turn it off and a step in the wizard that offers it.

## The switch

`coding_agent_real_browser`: **absent means ON**, because the screen is the answer that was asked for, and only an explicit `false` turns it off. Written by `POST /setup-api/coding-agent/enable { realBrowser }` (owner-only, like every field there, and refused to the MCP bearer), read back on the status, and shown in the Coding Agent's Settings beside the review-pass and media switches, rendered `?? true` so a server that predates the field reads as on. A factory reset clears the key **ahead of** the consent switch, so a reset box asks again.

## The wizard step

A `browser` step between `project` and `harness` — four numbered steps now. **Enable** writes `{ realBrowser: true }` first and on its own, because a Chromium that will not install today must not turn the owner's "yes" into "no", then opens the window and installs Chromium only when the route refuses with its stable `chromium_not_installed` code (discovered as a refusal rather than probed for, so no box pays a round trip to learn what all but a fresh one already answer). **Skip** writes `false`. Neither can strand the wizard: any failure is worded from the locale through `browserErrorText` with a Continue button under it that reaches the last step.

## The route

The browser route obeys the switch for **every** session. On: the desktop Chromium, started if nothing is up — the start page written to `browser.env` first, because systemd launches Chromium and `launch-browser.sh` reads it from there. Off: straight to the headless one, with the CDP port neither probed nor started, so the desktop is left exactly as the owner left it. The fallback stays for the three cases that need it: a foreign Chromium on the port, a desktop launch that never comes up (a 500 before this, a working session now), and the switch off.

`getSharedBrowser` is split into `getDesktopBrowser` / `getOwnBrowser` with **one cache slot per kind**, because the switch can move between two sessions and closing the handle a live session is driving would take that session's page down with it. The kind lives on the session — a required third argument to `openSession`, no default, so a caller that forgot cannot silently claim the owner's screen — and `sessionCount(kind)` is what lets the headless browser be closed on its own zero while a desktop tab is still open.

Every answer names the browser that served it: `browser: "desktop" | "headless"` on launch, on every action and on close, so a run that asked for the screen and got the headless one can tell — including a caller that kept nothing from the launch reply. The MCP browser tools relay the same field and their guidance says so.

## Tests

`coding-agent-browser-switch` (default on, off only on an explicit false, non-boolean refused, in the reset list), `browser.test.ts` (desktop when on, headless when off with no launch-env write and no `connectOverCDP`, the service started when on and nothing is up, a flip between two launches giving two browsers with both sessions still working, the headless one closed on its own count; the idle sweep and per-run page ownership re-verified unchanged), `enable.test.ts` (both wizard writes, the bearer refused, the field named in the 400), the settings panel, the wizard step, the MCP relay, and the ten locales. Type-check clean; the six suites above are 102 passing.

Ten locales, `mcp/README.md` and `CLAUDE.md` updated.

https://claude.ai/code/session_013Fz3uZVzYQzcNAkw1cQ8ff

## Summary by CodeRabbit

- **New Features**
  - Coding Agent runs can use either the visible desktop browser or an invisible headless browser.
  - Added a browser-selection step and setting, including installation, skipping, fallback, and recovery options.
  - Browser tools now identify the active browser and provide clearer viewing guidance.
  - Browser sessions can use desktop and headless modes independently.

- **Bug Fixes**
  - Desktop browser startup issues now fall back to headless browsing when possible.

- **Documentation**
  - Updated browser setup, session behavior, Coding Agent guidance, and translations.


yalexx:
@coderabbitai review

PR #723, opened 2026-09-06
fix: end the chat and terminal reconnect ladders when the socket constructor throws

KrasimirKralev:
**TASK-712** — the chat's reconnect ladder could never exhaust, and neither could the Terminal's.

## Symptom

Open the chat on a box where the browser refuses the gateway socket URL — mixed content behind a
proxy that hid the scheme, or a CSP `connect-src` block — and the window sits behind
"Switching AI provider… / This may take up to 30 seconds" for as long as it stays open, quietly
opening one doomed connection every three seconds. The error panel and its Try again button are
hidden while that overlay is up, so there is nothing the owner can do but reload the page.

The Terminal had the same shape from the same construct: a throwing `new WebSocket(...)` left the
connect lock raised, and the Reconnect button in the status bar became a no-op.

## Cause

`src/components/ChatPopup.tsx` — the `new WebSocket(wsUrl)` catch set `status: 'error'` without
calling `tearDownReloadOverlay()`, unlike the other terminal-failure paths. `reloadingSkill` stayed
true, so the safety-net retry effect stayed armed — and that effect RESETS `retryCountRef` to 0
before reconnecting, which is the counter both exhaustion branches are gated on. The ladder could
never reach its own end.

`src/components/TerminalApp.tsx` — `connect()` raises `connectLockRef` on entry and only
`onopen`/`onclose` lower it. A constructor that throws reaches neither, so every later attempt
returned at the guard.

## Harness-first

Nothing here belongs to OpenClaw or Hermes. The gateway socket, its reconnect ladder and the
`/terminal-ws` PTY socket are ClawBox's own browser code; no harness CLI, RPC or config owns a
browser reconnect policy, and no model list, session store or catalogue is touched. The change is a
`try/catch` around a DOM constructor plus the teardown the sibling paths already do.

## Both editions

The chat surface is shared. The gateway socket exists only on OpenClaw (a Hermes box runs no
gateway and `hasLiveConnection` is false, so `connect()` returns at its first line), which is why
the RED tests stub the harness to `openclaw` — the code path under test does not exist on Hermes.
The Terminal window is edition-independent and its fix applies to both.

## Sibling call sites

Every `new WebSocket(` in the tree was checked:

- `src/components/ChatApp.tsx:437` — already catches, and has no overlay, connect lock or retry
  ladder to strand. **Cleared, not changed.**
- `src/components/TerminalApp.tsx` — had no catch at all. **Fixed here**, plus every throw between
  the lock going up and the socket handlers being installed (the three dynamic `import()`s and the
  xterm construction) now releases the lock and reports, instead of stranding it as an unhandled
  rejection. That is the ChunkLoadError an in-app update produces under an already-open tab.
- `src/lib/hermes-dashboard-rpc.ts:100`, `src/lib/hermes-dashboard-turn.ts:795`,
  `src/app/setup-api/hermes/chat/clarify/route.ts:157` — server-side `ws`, already wrapped.
  **Cleared.**

The review pass then found a fourth terminal-failure path in the same component that had the same
omission and is a far more likely trigger: the connect-request `reject` (`ChatPopup.tsx`), which a
gateway takes when it REFUSES the connect frame — a protocol skew after an update, a rejected
device identity, a denied scope — while leaving the socket open. It now tears the overlay down and
lets the refused socket go, so `connect()`'s `readyState === OPEN` guard can no longer swallow both
the automatic retry and the manual Try again. Also from that pass: `tearDownReloadOverlay` now
clears `pendingModelSwitchResetRef`, which could otherwise survive an abandoned reload and reset the
owner's transcript during an unrelated provider change minutes later.

## RED → GREEN

RED, against unmodified `origin/beta` (`8653b118`):

```
 ❯ src/tests/components/chat-ws-constructor-throw.test.tsx (2 tests | 2 failed)
   × stops trying instead of reconnecting every 3 s forever
     AssertionError: expected 12 to be 2   // ten retry windows, ten more attempts
   × takes the overlay down so the owner gets the error and a Try again
     AssertionError: expected <span>Switching AI provider...</span> to be null

 ❯ src/tests/components/terminal-ws-constructor-throw.test.tsx (1 test | 1 failed)
   × says why, and lets Reconnect try again
     Error: Test timed out in 5000ms
   Unhandled Rejection: SyntaxError: The URL's scheme must be either 'ws' or 'wss'.
     ❯ src/components/TerminalApp.tsx:349

 ❯ src/tests/components/terminal-connect-lock-release.test.tsx (1 test | 1 failed)
   × reports the failure and can be retried
     Error: Test timed out in 5000ms          // never even reaches the error state

 ❯ src/tests/components/chat-connect-refused.test.tsx (1 test | 1 failed)
   × takes the overlay down, shows the reason, and leaves Try again working
     AssertionError: expected <span>Switching AI provider...</span> to be null
```

GREEN on this head, whole components project:

```
 Test Files  160 passed (160)
      Tests  1443 passed (1443)
```

`npx tsc --noEmit` exits 0.

## Notes

Found by the review pass and NOT fixed here, because it is not in this diff and wants its own card:
on the Hermes SKU the same "Switching AI provider…" overlay is raised by a provider change in
Settings and nothing can ever take it down — the two places that clear `reloadingSkill` both live
inside the gateway `onMessage` that edition never runs, so the transcript stays hidden until a page
reload. Recorded for the queue.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved failure handling for chat and terminal connections.
  - Refused or unavailable connections now close cleanly, dismiss reconnect overlays, release connection locks, and display the underlying error.
  - Reconnect actions remain available after failed connection attempts.
  - Prevented repeated automatic retries after terminal connection setup failures.

- **Tests**
  - Added coverage for refused connections, socket creation errors, overlay cleanup, retry behavior, and terminal lock recovery.

### DIFF
--- a/src/components/ChatPopup.tsx
+++ b/src/components/ChatPopup.tsx
@@ -2539,6 +2539,15 @@ function ChatPopup({ isOpen, onClose, onOpenFull, onOpenSettingsSection, onThink
     try {
       ws = new WebSocket(wsUrl)
     } catch {
+      // The same terminal-failure teardown the other three error paths do, and
+      // for the same reason. A wsUrl the browser's parser rejects throws here
+      // on EVERY attempt, so leaving the overlay up left the safety-net retry
+      // effect armed — and that effect resets `retryCountRef` to 0 before it
+      // reconnects, which is the counter both exhaustion branches are gated
+      // on. The ladder could never run out: one doomed attempt every
+      // RETRY_DELAY for as long as the chat window stayed open, behind a
+      // spinner that never came down (TASK-712).
+      tearDownReloadOverlay()
       setStatus('error')
       setErrorMsg('WebSocket creation failed')
       return
--- a/src/components/ChatPopup.tsx
+++ b/src/components/ChatPopup.tsx
@@ -2112,7 +2112,17 @@ function ChatPopup({ isOpen, onClose, onOpenFull, onOpenSettingsSection, onThink
           }
         },
         reject: (err: Error) => {
-
+          // The fifth terminal-failure path, and it used to be the one that
+          // forgot both halves. A gateway that REFUSES the connect frame
+          // (protocol skew, a rejected device identity, a denied scope) keeps
+          // the socket open, so without the close below `connect()` returns at
+          // its `readyState === OPEN` guard and even the manual Try again is a
+          // no-op; and without the teardown the reconnect overlay hides the
+          // error panel, which is the TASK-712 spinner-that-never-comes-down
+          // from a far likelier trigger than an unparsable URL.
+          tearDownReloadOverlay()
+          if (wsRef.current === ws) wsRef.current = null
+          try { ws.close() } catch { /* already closing */ }
           setStatus('error')
           setErrorMsg(err.message || 'Auth failed')
         },
@@ -2538,7 +2548,7 @@ function ChatPopup({ isOpen, onClose, onOpenFull, onOpenSettingsSection, onThink
     // Create WebSocket AFTER all handlers are defined to avoid race conditions
     try {
       ws = new WebSocket(wsUrl)
-    } catch {
+    } catch (err) {
       // The same terminal-failure teardown the other three error paths do, and
       // for the same reason. A wsUrl the browser's parser rejects throws here
       // on EVERY attempt, so leaving the overlay up left the safety-net retry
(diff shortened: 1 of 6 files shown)


KrasimirKralev:
Merged into beta (b4074ae6) after final review MERGE-OK on this head and green CI on this head, all review threads answered and closed. Deploy of beta to both boxes follows; box proof lands on TASK-712.

PR #729, opened 2026-09-06
fix: keep the interrupted reply on the adapter chat path

KrasimirKralev:
**TASK-721** — pressing Stop on a Hermes box threw the half-written reply away.

## Symptom

Mid-answer on a Hermes box (or a dual box running Hermes), the owner presses Stop. Everything the
box had already written vanishes from the transcript. On an OpenClaw box the same Stop keeps it.

## Cause

`src/components/ChatPopup.tsx`. The gateway path's `aborted`/`error` branch reads
`dropUnfinishedDirective(streamingRef.current)` and appends it before anything else. The **adapter**
path — the one a Hermes box runs — did `applyStreaming('')` in `dispatchTurn`'s catch and appended
nothing. TASK-703's fix, which made the interrupted turn appear exactly once instead of twice, only
put it there at all on one of the two editions, and the behavioural test that pinned it stubs the
harness to `openclaw`.

A second cause found by the review pass, in `src/lib/harness/hermes-adapter.ts`: the abort check
sits **after** the body parse, and an abandoned *stream* never reaches it — `readStreamedTurn`
throws on an `error` frame and on a stream that ends without `done`, so control leaves for the
catch. The check was live only for the JSON path, and its comment said the opposite. A Stop whose
stream closed cleanly first (the route's own `controller.close()` racing the client's abort, a
proxy dropping the SSE body) therefore came back as `upstream` + "The reply was cut off before it
finished." — and with this PR keeping the fragment, that red line would now sit *under* the answer
and read as though the answer had failed. The check moved into the catch, which covers both paths.

## Harness-first

There is no native mechanism to reclaim here: what the surface keeps on screen after an interrupt is
ClawBox's own transcript state on both editions, and the two harnesses already answer this the same
way at the protocol level — the OpenClaw gateway emits an `aborted` chat event, the Hermes route
answers 499 and the adapter turns it into `HarnessError('aborted')`. The defect was that only one of
the two client paths acted on it.

## Both editions

The mascot chat (`ChatPopup`) is the shared surface and it is the one fixed. `ChatApp`, the
full-screen chat, is gateway-only — it opens `/setup-api/gateway/ws-config` and drives
`wsRequest('chat.send')` with no harness branch — so it has no adapter path to fix; its own abort
branch already keeps the fragment. **Cleared, not changed**, except for the sentinel guard below.

## Sibling call sites

Every place either surface clears the streaming buffer on an abort was read:

- `ChatPopup.tsx` — the gateway `aborted`/`error` branch already kept it. The adapter catch is what
  this PR fixes.
- `ChatApp.tsx` — its `aborted`/`error` branch already keeps it.
- Every other `applyStreaming('')` in both files is a run START or a transcript clear, not an
  abort. **Cleared.**

The review pass then found that all three keeps asked `isSentinel(kept)` on the raw buffer.
`SENTINEL_RE` anchors the whole string, so a Stop landing between `NO_REPLY` and an `EMAIL:` line
the reply had begun leaves a value that no longer looks like a sentinel — it would be appended
verbatim and an email card minted off text the bubble never showed. All three now ask the question
the bubble asks, through `streamingEmailRefsText`.

## RED → GREEN

RED, against unmodified `origin/beta` (`381f34ed`):

```
 ❯ src/tests/components/chat-hermes-stop-keeps-partial.test.tsx (4 tests | 3 failed)
   × keeps the half-written reply, exactly once
     AssertionError: expected +0 to be 1        // the fragment is gone
   × keeps it above the failure line when the turn dies instead
     Error: Test timed out in 5000ms            // no answer bubble to order against
   × still says so when the stream ends without finishing and no Stop was pressed
```

GREEN on this head — the new file plus the neighbouring chat suites and the purity guard:

```
 Test Files  7 passed (7)
      Tests  71 passed (71)
```

`npx tsc --noEmit` exits 0.

The first cut of that suite was wrong in a way worth naming, because the review caught it and not
the assertions: its fake stream reader ignored the fetch signal, so Stop changed nothing and the
test ended the body by hand — which the adapter reports as an upstream failure, not an abort. All
three cases ran the *error* branch while claiming to test the abort branch, and the "shows no error
line" case passed over a DOM containing a full error line. The reader now rejects the pending
`read()` with an `AbortError` exactly as a browser does, the stream-ends-without-`done` case is its
own test, and the "exactly once" case renders under `StrictMode` so it actually pins the
updater-purity regression it names.

## Not fixed here, and stated rather than left implied

The kept fragment is **live-session only on Hermes**. The Hermes route records no assistant row for
an aborted or failed turn — only the question, plus a system `Error:` row on failure — and
`loadHistory` replaces the transcript with the box's copy, preserving only un-echoed *user* turns.
So switching chat tabs, reloading, or any reconcile drops the fragment again, and one screen later
the two editions still disagree. Closing that means recording the partial server-side on the abort
and error arms of the Hermes route's `streamTurn` (through `appendTranscript`, which that route
already calls three times a turn) — a route-level change to what the box deliberately does *not*
record on a 499, with its own regression surface, and outside this card's stated scope. Recorded for
the queue.

## Summary by CodeRabbit

- **Bug Fixes**
  - Interrupted assistant responses are now preserved instead of discarded.
  - Stopping a streamed response no longer displays an incorrect upstream error.
  - Partial responses remain visible when a stream fails, with the relevant error shown beneath them.
  - Prevented incomplete control messages and email-related content from appearing in chat transcripts.
  - Improved handling of interrupted or incomplete streamed replies to avoid duplicate or misleading messages.

### DIFF
--- a/src/components/ChatApp.tsx
+++ b/src/components/ChatApp.tsx
@@ -400,8 +400,13 @@ function ChatApp({ onThinkingChange, hideHeader = false }: ChatAppProps) {
             // bubble was showing, since the streaming render strips the line
             // whatever its payload.
             const keptMedia = splitAssistantMedia(kept)
+            // `streamingEmailRefsText` first: `SENTINEL_RE` anchors the whole
+            // string, so a Stop landing between `NO_REPLY` and an `EMAIL:` line
+            // the reply had begun leaves a value that no longer looks like a
+            // sentinel and would be appended verbatim. The same question the
+            // bubble asks, and the same one the popup's two branches now ask.
             if ((keptMedia.text.trim() || keptMedia.images.length > 0 || keptMedia.audio.length > 0)
-                && !isSentinel(keptMedia.text)) {
+                && !isSentinel(streamingEmailRefsText(keptMedia.text))) {
               setMessages(msgs => [...msgs, {
                 role: 'assistant',
                 text: prettifyAssistantText(keptMedia.text),
--- a/src/components/ChatPopup.tsx
+++ b/src/components/ChatPopup.tsx
@@ -4071,7 +4071,19 @@ function ChatPopup({ isOpen, onClose, onOpenFull, onOpenSettingsSection, onThink
       // Nothing is coming on either path, so the run ends here.
       sendingRef.current = false
       setSending(false)
+      // What the box managed to write is the OWNER'S, and it survives here for
+      // the same reason it survives on the gateway path's `aborted`/`error`
+      // branch: pressing Stop mid-answer keeps the fragment rather than wiping
+      // it. This path — the adapter's, i.e. the one a Hermes box and a dual
+      // box on Hermes run — used to clear the buffer and append nothing, so
+      // the same Stop lost the reply on one edition and kept it on the other
+      // (TASK-721). Read, clear, THEN append, all outside any updater, and
+      // before the failure line below so the answer stays above it.
+      const kept = dropUnfinishedDirective(streamingRef.current)
       applyStreaming('')
+      if (kept.trim() && !isSentinel(kept)) {
+        setMessages(prev => [...prev, { role: 'assistant', text: kept, timestamp: Date.now() }])
+      }
       clearToolCalls()
       // The agent is no longer parked on anything, so neither is the customer.
       // A card outliving its turn is a control with nowhere to post to.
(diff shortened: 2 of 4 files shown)


KrasimirKralev:
Merged into beta (ebaef9b6) after final review MERGE-OK on this head and green CI on this head, zero open threads. Deploy of beta to both boxes follows; box proof lands on TASK-703 TASK-721 .

PR #730, opened 2026-09-06
fix: let the core resolve which agent store the ClawBox AI credential goes to

KrasimirKralev:
**TASK-730** — a box can hold two ClawBox AI keys and silently prefer the revoked one.

## Customer-visible symptom

A Max customer was dead for two days on `HTTP 403 "Invalid token"`. His box held two ClawBox AI credentials and served the wrong one:

```
deepseek effective = profiles: ~/.openclaw/agents/main/agent/openclaw-agent.sqlite
         deepseek:default = claw_bd1…   <- sent, revoked, 403
         models.json      = claw_c98…   <- present, ignored
         source = models.json: ~/.openclaw/agents/pro-agent/agent/models.json
```

Every ClawBox surface reported the good key. The gateway sent the revoked one. Nothing on the box could tell him which — the truth came out of a hand-written probe an engineer composed in an email.

## Cause

`src/app/setup-api/ai-models/configure/route.ts` pinned the agent it writes credentials for:

```ts
const CLAWBOX_AGENT_ID = "main";
```

and passed `--agent main` to `models auth paste-api-key`, to the `models auth list` sign-in guard, to `models auth order set/clear`, and into the remedy sentence the refusal prints. Its docblock argued the pin was necessary: the CLI resolves a WRITE through `resolveSoleAgentId` and a READ through `resolveAmbientOwnerAgentId`, so an unpinned pair could address two stores.

The reasoning was sound and the constant was not. On a box whose roster names something other than `main`, the pin writes this device's credential into an agent store the gateway does not read — and whatever the real store holds, however stale, keeps serving turns. That is the customer's box.

## Measured on the installed core (2026.8.1)

All of the following was run with an isolated `OPENCLAW_HOME` / `OPENCLAW_STATE_DIR` / `OPENCLAW_CONFIG_PATH` / `HOME` under a scratch directory — never against a real box or a real config.

With a sole `pro-agent` roster and **no `--agent` flag at all**, the core resolves the same agent for the read and for the write:

```
$ openclaw models auth list --json
{ "agentId": "pro-agent",
  "agentDir": "…/agents/pro-agent/agent",
  "authStatePath": "…/agents/main/agent/openclaw-agent.sqlite",   <- inherited, see below
  "profiles": [] }

$ printf '…' | openclaw models auth paste-api-key --provider openai --profile-id openai:default
Auth profile: openai:default (openai/api_key)
  -> …/agents/pro-agent/agent/openclaw-agent.sqlite

$ openclaw models auth list --json
{ "agentId": "pro-agent",
  "authStatePath": "…/agents/pro-agent/agent/openclaw-agent.sqlite",
  "profiles": [ { "id": "openai:default", "provider": "openai", "type": "api_key" } ] }
```

And the same box, with the pin this PR removes:

```
$ openclaw models auth list --agent main --json
{ "ok": false, "error": { "message": "Unknown agent id \"main\". Use \"openclaw agents list\" to see configured agents." } }

$ openclaw models auth paste-api-key --agent main --provider openai --profile-id openai:probe
Unknown agent id "main". Use "openclaw agents list" to see configured agents.
```

So on that box beta cannot save a ClawBox AI credential at all, and on a box that declares `main` alongside the agent it actually runs, it saves it into the wrong store — silently, with a `200`.

## The fix (harness-first: **delete the pin**)

`--agent` is dropped from all three `models auth …` calls and from the remedy sentence. The agent is the harness's to resolve, and ClawBox now asks it to.

The read/write asymmetry the old docblock invoked is real but only bites on a roster with SEVERAL agents — and there `main` was never the right answer either: the write refuses with `Multiple agents are configured … Pass --agent <id>.`, which is a clearer failure than `Unknown agent id "main"`. On every box where a save can succeed, the guard, the paste and the order address one store, which is exactly the invariant the pin existed to protect.

The alternative — parsing `agents.entries` / `agents.list` in ClawBox and re-implementing `tryResolveSoleAgentId` — was written first and thrown away. It duplicated a rule this repo does not own (and got several roster shapes wrong: the core keeps a `list` element that is any object, keeps an `entries` key only when its value is a record, and runs every id through `normalizeAgentId`, which lowercases and rewrites punctuation), it read `openclaw.json` four separate times per request through a reader that returns `{}` on an unreadable file, and it answered a question the CLI answers correctly for free. `openclaw agents list --json` is named here as the native mechanism for the one thing a parser would still be needed for — an agent's resolved **directory** — and is not used, because after this change nothing needs one.

## Sibling call sites — swept

- The three `models auth …` calls and the remedy string in `configure/route.ts`: all unpinned, in one change, so they cannot drift apart.
- `AUTH_PROFILES_PATH` — the pre-v2 legacy `auth-profiles.json` — still resolves under `agents/main/`, renamed `LEGACY_AGENT_ID` to say why: the builds that wrote that file hardcoded `main`, and `doctor --fix` migrates it from wherever it finds it (`listAuthProfileRepairCandidates` enumerates every existing `agents/*/agent`). Unchanged from beta, deliberately.
- `src/lib/local-ai-token.ts:38` — **cleared, not a defect.** It reads `agents/main/agent/auth-profiles.json` only to hunt the `llamacpp-local` / `ollama-local` sentinels a pre-`b0c6e452` build wrote, and those builds hardcoded `main`, so that is exactly the right place to look. Read-only, fail-closed, no `claw_` key.
- `scripts/codex-auth-mirror.js` — **cleared.** It already enumerates every directory under `<stateDir>/agents` and models the shared-store ownership row; its `agents/main` mention is prose about the core's default.
- `src/lib/clawkeep-memory.ts:356` (`memory status --agent main --deep --json`) and `:827` (`memory index --agent main`) — **the same latent defect, different subsystem** (ClawKeep memory, not credentials), so not widened into this PR. Recorded as a residual for the ClawKeep lane rather than fixed here.

## What this PR deliberately does NOT take

The card names three problems. This is the second, and it is the only one that lives in ClawBox's own code:

1. **"The profile store silently wins over models.json."** That is the core's own documented precedence — a usable auth profile outranks `models.providers.<p>.apiKey`, recorded in this repo already — not something to override from here.
3. **"`doctor --fix` reports healthy over a revoked key."** That is `openclaw doctor`'s health rule; it belongs upstream. The ClawBox-side answer to "the box says it is fine while every turn 403s" is TASK-419, which is a separate PR in this batch.

## Residuals found while measuring (named, not fixed)

- **The revoked credential survives, shadowed.** A sole non-main agent INHERITS the shared/main store until it has one of its own — visible above as `authStatePath` pointing at `agents/main/` while `agentId` is `pro-agent`. Writing the agent's own profile shadows the inherited one, which is what stops the box serving the revoked key; it does not remove it. `openclaw models auth logout <profileId>` exists, is quoted as advice in this file, and is never invoked from anywhere in ClawBox — so a stale profile is never retired. That is the honest answer to "why did the box hold two credentials", and it wants its own card.
- **ClawBox's own reader uses the opposite precedence to the core's.** `resolveClawaiToken` (`src/lib/harness/credentials.ts`) reads `models.providers.<clawai>.apiKey` and never opens the profile store — so every ClawBox surface reported the good key for two days while the gateway sent the revoked one. Making that reader agree with the core (or at minimum report a mismatch) turns a silent outage into a visible one. In-repo, harness-native, and the next card after this one.
- **Known limit, measured and documented in the code:** because a fresh agent inherits the shared store, the sign-in guard can see a profile the paste would only have shadowed and refuse the save. That is the safe direction, and it is strictly better than what it replaces — the same box previously failed the whole save with `Unknown agent id "main"`.

## RED → GREEN

RED, on unmodified `origin/beta` (`8653b118`) — the branch's tests against beta's `configure/route.ts`:

```
     × refuses an anthropic API key that would replace the sign-in at anthropic:default
     × refuses an openai API key that would replace the sign-in at openai:default
     × reads and writes the SAME agent's credential store
     × clears an order ClawBox itself wrote once the box is down to one profile
     × addresses the same agent store the credential was written to
     × lets the core pick the agent when it writes the ClawBox AI credential
     × lets it pick for the guard and the order too, so all three address one store
     × names both OpenAI profiles, the sign-in first, when the box holds an API key too
     × revises the preference to the API key when THAT is the save
     × names a failed auth-order preference in the answer instead of hiding it

 Test Files  2 failed | 16 passed (18)
```

The headline one:

```
 FAIL  … > lets the core pick the agent when it writes the ClawBox AI credential
AssertionError: expected [ 'models', 'auth', …(7) ] to not include '--agent'
 ❯ src/tests/routes/ai-models/configure.test.ts:867:55
```

GREEN, on this branch: `src/tests/routes/ai-models/` `Test Files 18 passed (18)`, `Tests 340 passed (340)`. Whole unit project `Test Files 703 passed (703)`, `Tests 11480 passed | 1 skipped (11481)`. `tsc --noEmit` clean.

Eight pre-existing cases asserted `--agent main` and now assert that no agent is pinned; the one that documented "the order write and the credential write must address the SAME agent store" keeps its subject and gets a stronger assertion — both calls, and the paste beside them.

## Both editions

Hermes is unaffected and cannot reach the changed code: `assertNoSignInAt` returns early on `openclawIsAbsent()`, and a Hermes box has no `~/.openclaw/openclaw.json` and never runs `paste-api-key` — its ClawBox AI credential goes to the config store and `~/.hermes/config.yaml` through `applyClawaiToHermes`. Nothing new runs there and nothing throws.

**Device leg:** not run, and not applicable. The change is entirely in which arguments ClawBox passes to a CLI, the CLI's own resolution was measured against the pinned core in an isolated home (above) rather than reasoned about, and no box state was mutated. CI exercises the route.

## Device leg — read-only, both editions

**OpenClaw box** (`CLAWBOX_EDITION=openclaw`, core 2026.8.1):

```
$ openclaw agents list --json
[ { "id": "main",
    "agentDir": "/home/clawbox/.openclaw/agents/main/agent",
    "model": "deepseek/deepseek-v4-pro",
    "isDefault": true } ]
```

One agent, and it is `main` — so **this change is a no-op on our own boxes**: with `--agent` dropped the core resolves the same `main` the pin named, the credential lands in the same store, and nothing about those boxes moves. That is the "never worse" half proved on hardware rather than argued. What it fixes is the box shape the customer had, where the roster names something else and the pin either writes into a store the gateway does not read or fails the save outright.

**Hermes box** (`CLAWBOX_EDITION=hermes`): no OpenClaw binary and no `dist` on the box at all, so `paste-api-key` is unreachable there and `openclawIsAbsent()` short-circuits the guard — the edition leg confirmed by absence rather than by reasoning. **Declared gap:** `clawbox-setup` and `clawbox-hermes-gateway` were both `inactive` on that box at the time, so no route was exercised there.

### DIFF
--- a/src/app/setup-api/ai-models/configure/route.ts
+++ b/src/app/setup-api/ai-models/configure/route.ts
@@ -115,86 +115,21 @@ const OPENCLAW_HOME_DIR =
   || path.join(process.env.HOME ?? "/home/clawbox", ".openclaw");
 const CLAWBOX_HOME_DIR = process.env.HOME ?? "/home/clawbox";
 /**
- * The core's implicit agent, for a config that declares no roster at all.
+ * The agent whose LEGACY credential file ClawBox writes, pre-v2.
  *
- * `LEGACY_IMPLICIT_AGENT_ID` in the installed core
- * (`dist/session-key-*.js`), and what `listAgentIds` answers when
- * `agents.entries` / `agents.list` are both absent — which is every stock box.
+ * `main` is the core's implicit agent (`LEGACY_IMPLICIT_AGENT_ID`) and the
+ * directory it resolves for a config with no roster, which is every box that
+ * still has a legacy `auth-profiles.json` to write. Deliberately NOT used for
+ * the `models auth …` calls any more — see `pasteAuthApiKey`.
  */
-const IMPLICIT_AGENT_ID = "main";
-
-/**
- * The agent whose store ClawBox writes to, resolved the way the CLI would.
- *
- * ONE id is computed and passed explicitly to BOTH the credential path and
- * every `models auth …` call, because those commands do not resolve alike when
- * `--agent` is omitted: `resolveModelsTargetAgent` sends a WRITE through
- * `resolveSoleAgentId` and a READ (`auth list`) through
- * `resolveAmbientOwnerAgentId`, so a guard that read one store while the paste
- * wrote another would be no guard at all.
- *
- * What it must NOT be is a constant. `main` was pinned here, and on a box whose
- * roster declares another agent that wrote the device's ClawBox AI credential
- * into a store the gateway never reads — leaving whatever was in the real store
- * to keep serving turns, however stale. That is TASK-730: two credentials on
- * one box, the revoked one preferred, and every ClawBox surface reporting on
- * the good one.
- *
- * The rule is the core's own `tryResolveSoleAgentId`, not an invention:
- *
- *   - no roster property at all  → `main` (the implicit agent)
- *   - exactly one declared agent → that agent, whatever it is called
- *   - more than one              → the core cannot pick either, and neither do
- *                                  we: `main` is passed on, and the CLI
- *                                  answers `Unknown agent id "main"` when it is
- *                                  not one of them. That is today's behaviour
- *                                  and it is loud; guessing would not be.
- *
- * Both roster shapes are read because the core reads both
- * (`readAgentRosterProperty`: `agents.entries` as a record, `agents.list` as an
- * array).
- */
-function resolveAgentIdFrom(config: unknown): string {
-  const agents = (config as { agents?: Record<string, unknown> } | null)?.agents;
-  if (!agents || typeof agents !== "object" || Array.isArray(agents)) return IMPLICIT_AGENT_ID;
-  const ids = rosterAgentIds(agents);
-  // `undefined` is "no roster declared"; an empty roster IS declared, and there
-  // the implicit agent is not implied — pass `main` on and let the CLI object.
-  if (ids === undefined) return IMPLICIT_AGENT_ID;
-  return ids.length === 1 ? ids[0] : IMPLICIT_AGENT_ID;
-}
-
-/** The declared agent ids, or undefined when no roster property is present. */
-function rosterAgentIds(agents: Record<string, unknown>): string[] | undefined {
-  const entries = Object.hasOwn(agents, "entries") ? agents.entries : undefined;
-  if (entries !== undefined) {
-    if (!entries || typeof entries !== "object" || Array.isArray(entries)) return [];
-    return Object.keys(entries).map((id) => id.trim()).filter(Boolean);
-  }
-  const list = Object.hasOwn(agents, "list") ? agents.list : undefined;
-  if (list === undefined) return undefined;
-  if (!Array.isArray(list)) return [];
-  return list
-    .map((entry) => (entry && typeof entry === "object" ? (entry as { id?: unknown }).id : undefined))
-    .map((id) => (typeof id === "string" ? id.trim() : ""))
-    .filter(Boolean);
-}
-
-/**
- * The resolved agent for THIS request.
- *
- * Read per call rather than cached in a module constant: the roster is owned by
- * the owner and by `openclaw agents add`, and a value captured at import time
- * would keep addressing yesterday's store for the lifetime of the server.
- */
-async function clawboxAgentId(): Promise<string> {
-  return resolveAgentIdFrom(await readOpenClawConfig().catch(() => null));
-}
-
-/** Where that agent keeps the pre-v2 credential file. */
-function authProfilesPathFor(agentId: string): string {
-  return path.join(OPENCLAW_HOME_DIR, "agents", agentId, "agent", "auth-profiles.json");
-}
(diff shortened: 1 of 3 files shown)


KrasimirKralev:
Merged into beta (ebaef9b6) after final review MERGE-OK on this head and green CI on this head, zero open threads. Deploy of beta to both boxes follows; box proof lands on TASK-419 TASK-730 .

PR #739, opened 2026-09-06
fix: a worker's bytecode cache is not a stray file, and never a merge conflict

yalexx:
Found on real hardware: `.61` (rig board) on `beta`, reproduced twice — team runs **team-6rgz8cyx** and **team-5oxkp7a9**, 2026-09-06.

## Symptom

A trivial goal (add `subtract`/`multiply` to `calc.py`, write `tests.py`, write `USAGE.md`) killed the team both times. Every alert of the second run was the same file:

```
ALERT t1 | touched files outside its task: __pycache__/calc.cpython-310.pyc   -> rejected
REVIEW t1 | accepted                       <-- the work was correct
ALERT t2 | touched files outside its task: __pycache__/calc.cpython-310.pyc   -> rejected
ALERT t3 | Merge conflict: warning: Cannot merge binary files:
           __pycache__/calc.cpython-310.pyc (HEAD vs clawbox/team-5oxkp7a9-t3-1)
TEAM_STATUS | Team -> failed: Stopped after 3 alerts.
```

## Cause

CPython writes `__pycache__/*.pyc` beside the source it imports — the very file the task named. That one artifact cost the run three ways:

1. `outsideHint()` counted it as the worker touching files outside its task, so `clean` was false and the reviewer's rule rejected a task it had already **accepted**.
2. Each run's settle commits `git add -A`, which swept it into the worker's branch.
3. Two worker branches each carrying their own copy met at the merge, and git refused: *Cannot merge binary files*.

Three false alerts reached the ceiling. The work itself was correct throughout.

## Fix

- The artifacts are written into the repository's `info/exclude` — extending the mechanism already used for `.clawbox/`, and shared by every linked worktree — so they never enter a commit and an artifact-only conflict becomes **unreachable** rather than classified after the fact. The project's own `.gitignore` is not touched; the failing project had none, so the guard must not depend on one.
- `outsideHint()` ignores them, so producing one is not straying.

**The list is deliberately short.** `dist/`, `build/`, `.next/` and `coverage/` are generated too, but a task can legitimately be asked to produce them, and excluding those would silently drop the work the task was given. A false alert costs a retry; dropped work is not recoverable. Only names that are never a task's output are listed.

The alert ceiling is unchanged, and stray detection for real files is unchanged — a worker that writes `secrets.env` is still caught (covered by a test).

## Tests

Both halves were verified to fail without the fix, not merely to pass with it.

Stray detector, with the `outsideHint` filter reverted:
```
 × does not call a generated artifact a stray file
 Tests  1 failed | 20 passed (21)
```

Exclude writer, with the artifact patterns reverted:
```
 × keeps a worker's bytecode cache out of git entirely
 Tests  1 failed | 6 passed (7)
```

With the fix, the whole coding-team suite:
```
 Test Files  5 passed (5)
      Tests  54 passed (54)
```
(`coding-team`, `coding-team-worktree`, `coding-team-board`, `coding-team-reviewer`, `mcp-coding-team-tools`)

`eslint` on the four changed files: clean, exit 0.

### Note on the full unit run

`vitest --project unit` shows 6 unrelated files failing in my environment (`telegram/status`, `telegram/status-hermes`, `build-typecheck`, `code-highlight`, `install-tts-verdict-siblings`, `novnc-pin`). These are **not** from this change: they fail identically on a clean `origin/beta` checkout here, and the typecheck failures are all `Cannot find module` for `prismjs` / `@noble/*` — my worktree shares a `node_modules` from another checkout. **No coding-team type error appears in the typecheck output.** Worth a second pair of eyes on CI to confirm they're green there.

## Summary by CodeRabbit

- **Bug Fixes**
  - Generated files such as Python caches, virtual environments, dependency folders, macOS metadata, and TypeScript build information are now excluded from worktree changes and stray-file detection.
  - Worker commits no longer include irrelevant generated artifacts, while source files and requested build outputs remain included.
  - Worktree merging continues successfully when generated artifacts are present.


yalexx:
Verified on the box (Orin, beta merged in at 89f6e1ed, built and deployed), with the scenario from the description: a fresh git repo holding only `calc.py`, no `.gitignore`, and the goal "add subtract/multiply to calc.py, write tests.py and run it, write USAGE.md".

Team `team-23fow64w`:

- planner → 3 tasks; t1 ran first, t2 and t3 in parallel worktrees; all three **accepted** by the reviewer; team **done** in 75 s.
- Both t1 and t2 ran the interpreter in their worktrees (`python3 -c "import calc; …"` and `python3 tests.py`), so each worktree carried its own `__pycache__/calc.cpython-310.pyc` — the two-copies case that used to end in "Cannot merge binary files".
- Alerts naming `__pycache__`: **0**. `.pyc` files in any commit on any branch: **0**. Both worker branches merged home cleanly.
- `.git/info/exclude` carries the artifact patterns; running `python3 tests.py` on the merged checkout afterwards leaves `git status` clean (`git check-ignore -v` → `info/exclude:8: __pycache__/`).

Unit suites (`coding-team`, `coding-team-worktree`, `coding-team-board`, `coding-team-reviewer`, `mcp-coding-team-tools`) pass on the merged branch. The earlier `e2e-install` failure was the updater's upgrade spec never starting the update (`phase: idle`, `checkout-dirty` drift after 45 min), the known flake, not this change; CI is re-running on the merge commit.

PR #743, opened 2026-09-06
fix: stop offering a provider the box can run no model from

KrasimirKralev:
## TASK-668 — a Providers row for a provider the box can run no model from

**Customer-visible symptom.** A box that has saved a local model at primary scope carries
`models.mode: "replace"`, and under that flag the pinned core skips the authenticated catalogue for
**every** provider. Measured on a box, `openclaw models list --provider <p> --all --json | jq .count`
before and after the flip: anthropic 15 → 9, openai 30 → **2**, google 10 → **0**. Settings still shows
the Google row as *Connected*, the chat picker still offers Gemini, and every pick reaches a gateway
that will not route it. The owner's ruling (2026-09-06) is that such a row is not offered at all, while
a box that can run at least one of that provider's models keeps the row exactly as today.

**Cause.** Nothing on the box wrote down what an enumeration answered. `openclaw models list` returning
zero rows was logged, recorded as a failed refresh and thrown away, so the surfaces had no way to tell
"this provider has no models here" from "nobody has asked yet".

**The harness mechanism this uses.** `openclaw models list --provider <p> --all --json` is the
catalogue, and `/setup-api/ai-models/catalog` already runs it (boot warmup, a picker open) and caches
the result. This adds no probe and no fork: it writes the COUNT that enumeration already produced to
`data/catalog-cache/_enumerations.json` and reads it back. The core's own `replace` semantics
(`appendAuthenticatedCatalogRows`, `dist/list.row-sources-*.js:409`) are what make the zero real.
The earlier attempt at this card — marking every catalogue behind-generation on a `models.mode` flip —
was refused for buying a permanent `stale` flag plus a ~3-minute two-core Jetson fork per backoff
window; **this PR changes no freshness or backoff rule at all**. The empty-enumeration branch behaves
exactly as it does on beta; the recorded count is an extra fact beside it.

### What is hidden, and what is never hidden

Hidden: a provider whose row is `connected` or `needs-reauth` and whose own enumeration answered zero.

Never hidden, each for its own reason:
* **A provider the owner has not connected.** That row *is* the fix — connecting a cloud provider goes
  through `writeOpenAICompatProvider`, which writes its configured model rows **and**
  `models.mode: "merge"`, which is exactly what turns a zero-row provider back into a routable one.
  Hiding it would leave the box with no way out of the state that hid it.
* **The provider the box is pointed at.** It is why chat is broken; a row nobody can see is a row
  nobody can change.
* **A row carrying `needsRepair`** — it holds the Retry, the one affordance that fixes it.
* **Anything unknown**: no record, an unreadable record, a count older than the catalogue's own
  refresh interval, or a row whose other catalogue (`codex` for the OpenAI row) has not answered.

### The row comes back without a restart
* `notifyProviderSetChanged(<p>)` — a key pasted, a plugin switched on, a plan changed — forgets that
  provider's count, so the row returns on the next render and the enumeration that follows records the
  new truth. No fork is started by the forgetting.
* A `models.mode` flip forgets every count: they were all taken under a rule that no longer holds.
* A count ages out after the catalogue's own six-hour refresh interval.

### Surfaces swept, so they agree
`/setup-api/providers/status` (the strip and `AiProviderList`), the "Connect AI Provider" list
(`AIModelsStep`, which now reads the server's `unrunnable` list rather than deciding for itself), and
the chat picker (`/setup-api/chat/model` GET). `POST /setup-api/providers/default` resolves against the
UNFILTERED state, so naming a hidden provider explicitly is not answered `provider_unconfigured`, and
`setProviderEnabled` still flips the switch for a hidden provider rather than calling it unknown.
The MCP `ai_list_models` tool is Hermes-only and unaffected.

**Hermes.** No equivalent, deliberately. Hermes' rows carry a `total`, but a zero there is documented in
this repo (`providerHasModels`, `hermes-model-options.ts`) as "credentialed and its `/v1/models` could
not be enumerated" as often as "serves nothing", and `include_unconfigured=true` makes it
indistinguishable from "not set up". The Hermes strip therefore keeps every panel row until that
answer exists.

### RED → GREEN

RED on unmodified beta (`/setup-api/providers/status`, a box whose google enumeration answered zero):

```
FAIL src/tests/routes/providers/status-unrunnable.test.ts > drops the row when the box's own
     enumeration came back with nothing
AssertionError: expected [ 'clawai', 'openai', …(3) ] to not include 'google'
 ❯ src/tests/routes/providers/status-unrunnable.test.ts:86:27
```

and, for the fact the row is decided from:

```
FAIL src/tests/routes/ai-models/catalog-empty-enumeration.test.ts > records zero when the CLI answers
     with no rows at all
AssertionError: expected false to be true   // the enumeration record was never written
 ❯ src/tests/routes/ai-models/catalog-empty-enumeration.test.ts:78:35
```

GREEN — the new suites plus every neighbouring one:

```
npx vitest run src/tests/routes src/tests/unit src/tests/components/ai-models-step.test.tsx \
  src/tests/components/ai-provider-list.test.tsx src/tests/components/ai-providers-section.test.tsx \
  src/tests/components/openclaw-provider-panel.test.tsx src/tests/components/provider-rows-narrow.test.tsx
 Test Files  730 passed (730)
      Tests  11763 passed | 1 skipped (11764)
```

`npx tsc --noEmit` clean; `npm run lint` reports nothing new for the changed files.

### Review

One the reviewer review pass, applied: the hide is now gated on a connected row (it hid the connect
affordance); the empty branch keeps beta's backoff and publishes nothing (it had removed the only brake
and would have forked per catalog GET); a provider-set change and a TTL bring a hidden row back; the
record is generation-guarded like `publish`, so a fork from before a credential landed cannot hide the
provider that was just connected; `providerRowRunnable` now demands unanimity, so a ChatGPT-subscription
box cannot lose its OpenAI row to an empty API-key catalogue; the second publish path records too;
`AIModelsStep` never drops the selected row; `providers/default` and `setProviderEnabled` no longer
answer a false failure over a hidden provider.

Two findings deliberately not taken, with reasons in the code: a payload whose every row reports
`available: false` is **not** recorded — it is also exactly what a provider with no credential looks
like, and the two cannot be told apart from the payload; and the clean-zero test cannot wait for the
child's exit code, because the CLI's wrapper holds the pipe for about three more minutes after the JSON
arrives, so it requires an explicit `count: 0` with `ok !== false` and no stderr instead.

### Device

Read-only on the OpenClaw box: `models.mode` is `merge` today and the configured rows are deepseek 3,
llamacpp 1, openai 1, openrouter 4 — no google rows, which is why the flip takes google to zero. The
count itself comes from the measurement recorded on the card. Nothing was mutated and no enumeration was
forked. The change touches no updater, gateway, systemd unit, device path or edition lock, so CI
exercises it fully.

## Summary by CodeRabbit

* **New Features**
  * Provider lists now hide connected providers with no runnable models while preserving active and default providers.
  * Explicitly selected providers remain usable even when hidden from selection lists.
  * Provider status reports unavailable providers.
  * Model catalog refreshes track valid model counts and preserve prior results for uncertain responses.
  * Availability data refreshes after relevant configuration changes.

* **Bug Fixes**
  * Improved handling of empty, refused, malformed, and filtered catalog responses.

### DIFF
--- a/src/app/setup-api/ai-models/catalog/route.ts
+++ b/src/app/setup-api/ai-models/catalog/route.ts
@@ -14,7 +14,7 @@ import {
   isNonChatModelId,
   subscriptionSurfaceProvider,
 } from "@/lib/provider-models";
-import { recordProviderEnumeration } from "@/lib/provider-runnable";
+import { forgetProviderEnumeration, recordProviderEnumeration } from "@/lib/provider-runnable";
 
 export const dynamic = "force-dynamic";
 
@@ -569,6 +569,13 @@ async function fetchSubscriptionSurfaceIds(provider: string): Promise<Set<string
     publishedSeq.set(surfaceProvider, currentSeq(surfaceProvider));
     recordSuccessfulRefresh(surfaceProvider);
     await writeDiskCache(surfaceProvider, payload);
+    // The same count `publish` records. This is the file's second full publish
+    // path, and a catalogue published here while the record still said zero
+    // would keep a provider hidden that had just answered with rows. Dead
+    // today — no `SUBSCRIPTION_SURFACE` entry names a separate provider — and
+    // closed rather than left for the day one does, like every other rule this
+    // function had to be taught twice.
+    await recordProviderEnumeration(surfaceProvider, payload.models.length);
     return new Set(payload.models.map((m) => m.id));
   } catch (err) {
     console.warn(
@@ -1033,10 +1040,24 @@ function toFetchResult(
       ? `all ${rows.length} listed rows report available: false (no route this box can take yet)`
       : `${rows.length} rows listed, none of them chat models this picker can offer`;
   }
-  // A clean zero: the command ran, said nothing was wrong, and listed nothing.
-  // `diagnostic` is empty in exactly that case — a refusal or stderr fills it
-  // above, and rows that were all filtered out fill it just now.
-  return { models, diagnostic, emptyIsAnswer: models.length === 0 && !diagnostic };
+  // A clean zero: the command ran, said nothing was wrong, and STATED that it
+  // has nothing — `count: 0` beside an empty `models`. Read positively rather
+  // than inferred from absence, because a truncated or shape-shifted payload
+  // parses into the same emptiness as a real answer and this verdict hides a
+  // row. `diagnostic` covers the rest: a refusal or stderr fills it above, and
+  // rows that were all filtered out by OUR chat rule fill it just now.
+  //
+  // Deliberately NOT extended to "rows listed, every one `available: false`",
+  // which the core does report and which is the other way a box says it can
+  // route nothing here. It is also exactly what a provider with no credential
+  // looks like, and the two cannot be told apart from the payload — so that
+  // case keeps beta's behaviour and records nothing.
+  const emptyIsAnswer = models.length === 0
+    && rows.length === 0
+    && parsed.ok !== false
+    && parsed.count === 0
+    && !diagnostic;
+  return { models, diagnostic, emptyIsAnswer };
 }
 
 function fetchOpenclawCatalog(provider: string): Promise<CatalogFetchResult> {
@@ -1355,9 +1376,9 @@ export function refreshInBackground(
     await writeDiskCache(provider, payload);
     // What the box can run, for the surfaces that only need the COUNT (the
     // Providers rows, the picker). Written from the published list rather than
-    // the raw one so it cannot disagree with what the picker is offered, and
-    // written on every publish so a provider that comes back from zero stops
-    // being hidden without anything else having to notice.
+    // the raw one, so the number and the catalogue the picker is offered cannot
+    // disagree. Reached only past the generation guard above, so a fork from
+    // before the box changed records nothing.
     await recordProviderEnumeration(provider, payload.models.length);
     console.log(
       `[catalog] refreshed ${provider}: ${models.length} models`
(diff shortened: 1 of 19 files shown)


KrasimirKralev:
Both actionable comments addressed in `d77c39bd`, and thank you — the first one is the exemption the change most needed a test for.

- **`ai-models-step.test.tsx`** — added *"keeps the row the panel is CURRENTLY on, whatever the box says about it"*. The filter deliberately never drops the selected provider, because everything else in that component that decides what to render (the keep-the-selection-valid effect, the credential panel, the deep-link handler) resolves against the unfiltered list; dropping the selected row would leave its form on screen with no radio above it and no "show more" toggle to get back. The new case sits beside the two that assert the row is omitted when it is not selected, so the pair pins the exemption rather than the filter.
- **`providers/default.test.ts`** — added *"still promotes a provider the picker hides because it can run no model"*: the filtered state offers only ClawBox AI, the unfiltered one carries the Google row, and the POST promotes `google/gemini-2.5-pro` with a 200. That is the fallback's whole purpose — hiding a row is about what is OFFERED, and answering `provider_unconfigured` over a provider that holds a working credential would be a false failure.

Both fail without their respective code path. Full run after the change: `src/tests/routes` + `src/tests/unit` + the provider component suites green, `tsc --noEmit` clean.

KrasimirKralev:
Merged into beta (cd3a2458) after final review MERGE-OK on this head and green CI on this head, all review threads closed. Deploy of beta to both boxes follows; box proof lands on TASK-668 .

PR #745, opened 2026-09-06
fix: scope model catalogue fallback by provider

Tiancheng-Xu:
Follow-up implementation for #738 from a fork, because the contributor account cannot push to the upstream head branch.

Changes:
- keep model catalogue fallback scoped to the requested provider;
- add coverage for missing provider entries and top-level fallback behavior.

Validation:
- target tests — 33 passed
- `npx tsc --noEmit` — passed
- target ESLint — passed
- full-repository lint still reports pre-existing issues outside the changed files.


yalexx:
@Tiancheng-Xu PR requests need to merge with beta before main

PR #758, opened 2026-09-06
fix: Memory Shard — the full-UI sweep's findings

yalexx:
## What this is

The Memory Shard app was tested end to end **in the box's own Chromium** (CDP 18800, a new tab of the owner's browser) plus headless passes for the standalone page, a phone viewport and the German locale — every face: home, settings, switch off/on, Start over, the whole wizard with a real folder, the first index, and the home card's controls. What works is listed at the end; this PR fixes what did not.

## Defects fixed

| | Found | Fix |
|---|---|---|
| **Folder writes raced; last writer wins** | `POST`/`DELETE …/memory/sources` each read → filter → write `memory.search.extraPaths` through one `openclaw config set` (~5 s). Two overlapping writes: the second add's *answer* listed the folder while the delete landed last — openclaw.json ended with `[]` and the wizard provisioned with no folder. | `mutateExtraPaths(fn)` in `src/lib/memory-shard.ts`: one in-process queue for every read-modify-write (tail always settled; a failed turn is the caller's). Inside the queue the config is read STRICTLY (`readConfigStrict`) so an unreadable openclaw.json is 500 `read_failed`, never `[]` written over the owner's list. The route answers `write_failed` for a CLI failure and invalidates the status cache after a real change. |
| **No feedback and no lock during the write** | "Use this folder"/"Remove" were only `disabled` for the ~5 s; the picker stayed open and the list unchanged — which invites the second click above. | One shared `MemoryShardFolders` component (wizard step and the new settings card): one busy state locks every folder control, the pressed button reads *Adding…*/*Removing…*, the picker closes only on a successful add, the route's own sentence is shown on failure, and the wizard's Next waits for a write in flight. |
| **A timezone change never reached the running web server** | `computeNextMemoryRunMs` and ClawKeep's backup scheduler use `Date#setHours`, i.e. the process zone Node fixes at start. The owner set Europe/Sofia at 19:51; the server had started 19:44 in UTC; "weekly Tue 04:30" was armed for **07:30** while the card said "device-local". | `applyProcessTimeZone(tz)` (canonical IANA name → `process.env.TZ`, which Node honours at runtime). The timezone route applies it after the OS leg lands and re-arms both schedulers; `instrumentation.ts` seeds it at boot from the stored zone when it is recorded as applied, ahead of both schedulers — covers the update-ordering case too. Verified on the box: the same schedule now arms at 04:30 local. |
| **The amber "run a full reindex" banner right after a full reindex** | One status read after a run carried the mid-rebuild reading (identity `mismatched`, pending 1): a probe that straddled the run's cache invalidation, or one the 120 s TTL stored fresh during the run. The card flips `running` off on that read and then polls every 30 s. | `clawkeep-memory.ts` tracks an *unsettled* cached reading (set at a run's finish, or by a probe that saw a run in flight); `getMemoryStatus` waits for the post-run probe in that one case (never while a run is going, never for `peekMemoryStatus`), and a straddled probe is retried once. The card keeps its fast poll for three ticks after the flip it observed, with no clock comparison. |
| **No way to see or change folders after setup** | The folders lived only in the wizard's step 1; the only path back was Start over, which also clears the schedule. The card's "Sources" stat is OpenClaw's source kinds and stayed 2 with a folder configured. | A **Folders it reads** card on the settings page, built from the shared component. Five new keys in all ten locales. |
| Wizard: a half-typed time was saved as `""` and sanitised to 03:00 | | The home card's rule (`TIME_OF_DAY`, now exported once from `memory-shard-state.ts`) applied to the wizard. |
| Wizard: a failed Remove was silent | | The route's sentence is shown. |

## Behaviour changes a person could notice

- The card says "Indexing…" for about one status probe (~8 s warm) longer than the pass itself, and then shows the settled figures rather than a wrong banner.
- Adding or removing a folder locks the other folder controls for the length of the write (~5 s; the gateway restarts on that config change, as before).
- Setting the timezone re-arms the backup and memory schedules in the new zone at once.

## Verification

- Unit: the Memory Shard, sources, timezone and status-cache suites (new: `memory-shard-folders`, `timezone-process-zone` ×2, `instrumentation-timezone`), plus the full suite on the box — 935 of 940 files pass; the 4 failing files (`clawbox-firewall-policy`, `ai-models/configure`, `failover-waits-for-route`, one timing-flaky chat component test) fail identically on beta before this branch, reading the box's own files. The one failure this branch caused — the time-of-day regex parity test, after the regex moved into `memory-shard-state.ts` — is fixed in the second commit.
- Lint on every changed file and `tsc --noEmit` over the tree: clean.
- Live, on the box after build + restart, in the box's own Chromium:
  - Settings → **Folders it reads** lists the configured folder; Remove shows *Removing…* and greys Add for the ~5 s write; the list and `memory.search.extraPaths` agree at every step; Add through the picker shows *Adding…*, a second click during the write is refused, the picker closes only once the config holds the folder.
  - `weekly, Tue 04:30` arms `nextRunAtMs` at **04:30 local** (was 07:30); the boot log reads `Process timezone seeded from the applied zone: Europe/Sofia`; ClawKeep's 02:00 backup arms at 02:00 local.
  - A full reindex watched at 500 ms intervals: `running/valid` → `succeeded/valid/healthy`, no stale reading; the card shows "finished just now · by hand · full · 8 s" with no amber banner.
  - No raw keys, no unfilled placeholders, the owner's own tab untouched (tabs before/after 1/1).

## What the sweep confirmed works (unchanged)

Home card with stats and chips; Index now; the Full reindex confirmation (Cancel leaves no run); schedule toggle, daily/weekly, weekday row, time field; settings switch off → chip OFF, paused hint, `index` refused 409 `disabled`; Start over (two taps) → wizard; Settings reachable over the wizard; the wizard's folder picker (root = home, up, entries), schedule step, provision phases; the first pass; the standalone `/app/memory-shard` page on desktop and phone with its own Settings; no console errors, raw keys, unfilled placeholders or horizontal overflow; every locale carries every key.

https://claude.ai/code/session_01FqyBxqRNSGTzDuQJzAYUqD

## Summary by CodeRabbit

- **New Features**
  - Added Memory Shard folder management in setup and settings, including browsing, adding, and removing source folders.
  - Added localized labels and error messages for folder operations.
  - Added timezone application during setup and startup, with scheduler refresh after successful changes.

- **Bug Fixes**
  - Improved Memory Shard status updates after indexing completes.
  - Prevented stale status results during active indexing.
  - Improved handling of invalid schedule times and failed folder configuration updates.
  - Prevented conflicting folder changes from overwriting each other.


yalexx:
Addressed CodeRabbit's three findings in 00ca731c, each with a test that fails without the fix:

- **Read-back after a landed write** — `mutateExtraPaths` now answers the written list when only the post-write read fails (logged), so the route says success and marks the status cache stale; the pre-write `read_failed` refusal is unchanged. CLAUDE.md's contract sentence updated.
- **Late mount read in `MemoryShardFolders`** — a revision ref bumped on every landed write; a GET that began before one is not applied.
- **Overlapping timezone POSTs** — the whole transition (state read → env write → store keys → root step → process zone → harness → applied marker) runs under an in-process queue, and a second tab's identical zone is answered `changed: false` off the state the first left.

PR #763, opened 2026-09-06
fix: stand the Hermes box down when the cloud voice is withdrawn (TASK-745)

KrasimirKralev:
**TASK-745** — found by the reviewer of #757 (MEDIUM-2, LOW-3, LOW-5), and since **reproduced on
the real Hermes box under the device mutex** (device lane 2026-09-07 run "e", finding 1).

Built on **#762 (TASK-744)**, now merged — this branch is rebased on it, so the withdrawal fix sits
on the plan-tier gate as it shipped. The one rebase conflict was two `describe` blocks inserted at
the same line in the shared test file; both kept.

## What the customer got

On a Hermes box whose plan drops, the boot seed withdrew the cloud voice — and left
`tts.provider: openai` naming a slot it had just deleted. Measured on hardware:

```
store.set("clawai_tier","flash"); sudo systemctl restart clawbox-setup
journal: [register-mcp] removed the ClawBox AI cloud voice: this box's plan no longer includes it
yaml.safe_load(~/.hermes/config.yaml)["tts"]
  → {'provider': 'openai', 'providers': {'clawbox-local': …}}      # no `openai` key at all
```

On this edition `openai` is the **harness's own built-in** provider
(`BUILTIN_TTS_PROVIDERS`, `tools/tts_tool.py:780`), not a named entry ClawBox created. So the
selection still resolves — to a provider with no `base_url`, which is **api.openai.com**. Where the
box used to get a 403 at our own proxy it now gets a 401 at OpenAI, and the owner's text goes to a
third party on every spoken reply. TASK-718's refused round trip was relocated, not removed.

The OpenClaw analogue is genuinely different and is left alone: there the delete removes a NAMED
provider entry, so the surviving selection names nothing and the panel showing "no cloud voice" is
the truth.

## Why this is not "silently rewriting the owner's choice"

#757's ruling — the panel's job is to show the choice is gone, and rewriting it would hide the
downgrade — protects a provider **the owner chose**. `tts.provider: openai` on a box whose
`tts.openai` ClawBox has just emptied is not that: it is ClawBox's own residue, written by
ClawBox's arm, pointing at a slot ClawBox removed. Restoring the selection the box had before the
arm moved it is the opposite of overriding the owner.

That is exactly what the arm's own ownership test already says, read backwards. The arm moves
`tts.provider` only from `""`, Hermes' factory `edge`, or `clawbox-local` — the three values that
mean *nobody has chosen*. The withdrawal now uses the same three, so anything else (`elevenlabs`,
`piper`, a hand-picked `openai`) is the owner's and is never touched.

**This is queued as an owner decision**; the PR implements the reading above and nothing here
depends on it beyond the one corner named under "The open question".

## What it does

The plan step's verdict carries where the box speaks from afterwards — two answers only:

| verdict | when | writes |
|---|---|---|
| `withdraw clawbox-local` | the selection is `openai` AND the box has a resolvable on-device definition | `config set tts.provider clawbox-local`, then the slot removal |
| `withdraw keep` | the selection is anything else | the slot removal only |
| `hold …` | the selection is `openai` and there is nowhere safe to go | nothing at all |

The selection moves **only from `openai`** — what our own arm wrote. An owner's `elevenlabs`, the
factory `edge`, an unset key and a box already on its own voice are all left where they are, because
removing our slot says nothing about where any of them sends a request.

**The selection moves FIRST**, which is the mirror of the arm's definition-before-selection rule
and the opposite of this PR's own first draft. Every intermediate state has to be safe, and the
dangerous one is a slot our endpoint has been taken out of while `tts.provider` still says
`openai` — the built-in then resolves to api.openai.com. De-selecting first means the only
intermediate state is a box speaking from itself with a stale definition nothing points at, and a
failed move leaves the route complete so a reply is refused at our own proxy instead.

**One atomic call removes the slot.** `hermes config unset tts.openai` — `_unset_nested`
(`hermes_cli/config.py`) walks a dotted path and deletes whatever it lands on, then prunes the
empty container. That takes `tts.openai.voice` (the Voice tab's `KEYS.cloudVoice`, LOW-3 of the
#757 review) and any `format`/`speed` an owner added beside our three, which a fixed per-key list
cannot know about.

## Two things measured on the box that changed the design

Read-only, from the installed Hermes 0.20.5:

1. **`hermes config unset` exits 1 when the key was absent.** `unset_config_value` prints
   `Config key not set` and `sys.exit(1)` on a false return from `_unset_nested`. No box in the
   field carries `tts.openai.voice` — the real box's `tts.openai` keys are
   `['api_key','base_url','model']` — so this PR's first draft, which added `voice` to the per-key
   loop, would have reported a **failed withdrawal on every real box**. This contradicts a comment
   in this repo at `src/lib/hermes-local-ai.ts:501-504` recording unset-on-an-absent-key as a
   proved no-op; that comment is wrong for the pinned build and is noted for its owner.
2. **`unset` accepts a non-leaf key**, which is what makes the atomic slot removal possible.

## Where there is nowhere safe to stand down to

A box whose `tts.providers.clawbox-local` the harness cannot resolve gets **no withdrawal at all**,
and says so on the boot log.

This branch first claimed the opposite — that selecting `clawbox-local` undefined fails locally, so
nothing leaves the box. **That claim was false**, and it was disproved by executing the shipped
resolver on the box rather than by reading it: `_resolve_command_provider_config`
(`tts_tool.py:857`) answers `None` for a name that is unknown or not a command type,
`_dispatch_to_plugin_provider` answers `None`, and the dispatch then walks its `elif` chain into
`else: # Default: Edge TTS`. The module states it at `:3276-3282` — *"unknown names hit the Edge
TTS default at the bottom"*. The raise the branch cited (`:1339`) is unreachable from that path.
So the shortcut handed the owner's speech to **Microsoft** while the journal said the box spoke for
itself.

Every way out of that state sends the owner's words to somebody:

| option | where the words go |
|---|---|
| leave `tts.provider: openai` over a removed slot | api.openai.com — this card's own defect |
| `hermes config unset tts.provider` | Edge, i.e. Microsoft (`DEFAULT_PROVIDER`, `:211`/`:661`) |
| select `clawbox-local` with no definition | Edge, by the fall-through above |
| **keep the cloud voice** | **our own proxy**, which answers 403 |

So the withdrawal holds. It costs the refused round trip TASK-718 set out to remove, on the one box
shape that cannot speak for itself, and it is the only outcome that adds no new recipient. It also
self-heals: `install.sh` `step_openclaw_tts` re-registers the definition on every install and every
update, after which the next boot withdraws normally. The real box carries it —
`{'command': '…/clawbox-tts.sh --text-file={input_path} -- {output_path}', 'output_format': 'wav',
'type': 'command'}` — so this is a corner, not the common path.

"Resolvable" is `_is_command_provider_config`'s own rule (`tts_tool.py:846`), transcribed: a
non-empty `command`, and `type` absent or `command`.

## RED → GREEN

**RED on unmodified `origin/beta`** (scratch worktree at beta, only this suite copied in):

```
 Test Files  1 failed (1)
      Tests  4 failed | 33 passed (37)
```

— the selection left over an emptied slot, the residual `tts.openai.voice`, the stand-down, and
the swallowed plan-step failure.

**GREEN on this head:** whole suite **948 files / 14,297 tests**; `tsc --noEmit` clean; `bash -n`
clean; all 30 embedded python heredocs `py_compile` clean. Two guards are mutation-proven: deleting
`sys.excepthook = _plan_failed` fails the excepthook case, and the `hermes` stub now models the
CLI's real `unset` contract so the atomicity change fails on the per-key shape.

## Proved on hardware, read-only

The plan step extracted from the shipped script and run against the real Hermes box's own
`~/.hermes/config.yaml` and device store, no mutex taken and nothing written:

```
real config (provider=openai, its own definition)  withdraw clawbox-local
openai, NO definition                              hold … no on-device voice to stand down to
openai, definition with no command                 hold … no on-device voice to stand down to
owner picked elevenlabs                            withdraw keep
factory edge                                       withdraw keep
already on its own voice                           withdraw keep
```

The box's own definition, read at the same time, is
`{'command': '…/clawbox-tts.sh --text-file={input_path} -- {output_path}', 'output_format': 'wav',
'type': 'command'}`.

## Also fixed: the plan step no longer hides why it failed

`python3 - <<'PY' 2>/dev/null` meant a genuine bug in the block reached the operator as
`hold the voice plan could not be computed` and nothing else (LOW-5 of the #757 review). The
redirect is gone, so a SyntaxError reaches the journal in full and only ever quotes this script's
own source. A RUNTIME failure is still not allowed to: `sys.excepthook` reports the exception type
and the line number, because a YAML error quotes the offending line and the line it would quote is
`tts.openai.api_key`.

## Notes for whoever tests this on a box

Three facts from the device lane, each of which would otherwise produce a false reading:

- the seed logs to **`clawbox-setup`**, not `hermes-gateway` (`[register-mcp]` lines over 7 days:
  `clawbox-setup` 47, `hermes-gateway` 0), and the sanctioned restart is
  `sudo systemctl restart clawbox-setup`.
- `hermes config set`/`unset` strips ~2 KB of the harness's own trailing comment template from
  `config.yaml` and it returns on the next seed boot — an upstream defect, cosmetic. **Compare the
  parsed `tts` subtree, never the bytes.**
- `clawai_tier` cannot be held at a test value while a browser has the AI Models panel open, since
  `/setup-api/ai-models/status` rewrites it within seconds. The flip and the restart must be one
  script.

## Scope

Hermes only, and proved so: `scripts/register-mcp.sh` `exit 0`s on any edition but `hermes|dual`
(`:132`), and this change is entirely inside that script — `gateway-pre-start.sh` is untouched.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved ClawBox AI voice withdrawal by switching to a valid local voice provider before removing cloud voice settings.
  - Preserves the cloud voice route when a safe local provider cannot be selected.
  - Prevents incomplete voice configuration cleanup when no valid local fallback is available.
  - Improved diagnostics for unexpected voice-plan processing failures, including sanitized error location details.
  - Partial configuration cleanup now continues reliably when settings are already absent.

### DIFF
--- a/scripts/register-mcp.sh
+++ b/scripts/register-mcp.sh
@@ -1067,12 +1067,17 @@ if [ -z "$CLAWBOX_VOICE_PROXY" ]; then
   CLAWBOX_VOICE_PROXY="https://clawbox.com/api/ai"
 fi
 CLAWBOX_VOICE_MODEL="gpt-4o-mini-tts"
+# The on-device voice `install.sh` registers and selects on every Hermes box,
+# and the only provider name the withdrawal below ever writes. One binding, so
+# the name the plan step may emit and the name the shell accepts cannot drift.
+CLAWBOX_VOICE_LOCAL="clawbox-local"
 CLAWBOX_VOICE_PLAN=$(
   CLAWBOX_DEVICE_STORE="$PROJECT_DIR/data/config.json" \
   CLAWBOX_HERMES_CONFIG="$HERMES_CONFIG" \
   CLAWBOX_SPEECH_DEVICE_TIER="pro" \
   CLAWBOX_AI_PROXY_URL="$CLAWBOX_VOICE_PROXY" \
   CLAWBOX_VOICE_MODEL="$CLAWBOX_VOICE_MODEL" \
+  CLAWBOX_VOICE_LOCAL="$CLAWBOX_VOICE_LOCAL" \
   CLAWBOX_KOKORO_STAMP="$HOME_DIR/.cache/clawbox/kokoro-installed" \
   python3 - <<'PY' || echo "hold the voice plan could not be computed"
 import json, os, sys, traceback
@@ -1145,7 +1150,7 @@ OUR_PROXIES.discard("")
 # nobody has chosen: `edge` is Microsoft's cloud voice, which the box gets for
 # having expressed no opinion.
 FACTORY_PROVIDER = "edge"
-LOCAL_PROVIDER = "clawbox-local"
+LOCAL_PROVIDER = os.environ["CLAWBOX_VOICE_LOCAL"]
 CLOUD_PROVIDER = "openai"
 # The one speech model the proxy serves; the same constant `hermes-tts.ts` pins
 # as HERMES_CLOUD_TTS_MODEL, because a request that names anything else is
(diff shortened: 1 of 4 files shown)


KrasimirKralev:
Merged into beta (ccd26686) after final review MERGE-OK on this head and green CI on this head, all review threads closed. Deploy of beta to both boxes follows; box proof lands on TASK-718 TASK-744 .

PR #772, opened 2026-09-07
fix: the desktop shell — the sweep's findings on desktop and phone

yalexx:
## What this is

The desktop shell, from the desktop-wide UI sweep of 2026-09-07 (desktop and phone viewports, plus a German pass).

## Defects fixed

| Found | Fix |
|---|---|
| **"Back to Desktop" crashed to Next's "This page couldn't load", and the browser's Back button never closed a window** — on desktop and phone. The desktop pushed the STRING `"clawbox"` as its history state; Next 16's router writes its own `__NA` field onto the pushed state and throws on a string. | The entry is an object with a marker field (`{ clawbox: true }`), recognised by that field, so Next can merge its keys. |
| A window opened while the chat was docked landed with its Minimize/Maximize/Close under the chat panel. | A window OPENED beside a docked chat is fitted to the strip left of the panel; a window already on the desktop keeps its size when the chat docks, as before; a fitted size is never written as the app's remembered size (only an owner's own resize is). |
| The phone shelf drew no app buttons: the code computed Settings and the open apps for the phone bar and never drew them, so an app switched away from could not be reached again. | A scrolling middle row draws them; the chat button stays in the fixed tray cluster. |
| The power menu's "Internet · N ms" / "No internet", the toast's Dismiss and the notice destinations' phrases were English literals; the shelf clock and the menu's date followed the browser's locale rather than the box's language. | Catalogue keys in all ten locales; the clock and date use the UI language (with the browser's regional variant of it, so a UK browser keeps its 24-hour clock). |
| The login page could never show the box's language: it asked for `ui_language` before a session existed and was answered 401. | `GET /setup-api/preferences?keys=ui_language` — path and query exactly, one shared constant in `src/lib/ui-language-read.ts` — is answered with no session by the middleware AND the route; anything wider (a second key, `all=1`, a POST, an encoded query) stays behind the session, with tests on both doors. |

By design, left alone: a desktop opened within a minute of a notice replays it (the ring replaced a slot the first poller consumed).

## Verification

- Unit + component: desktop-shell regressions (the history entry re-enacted against Next's own write), window-snap, chrome-window docked-chat, chrome-shelf phone bar, system-tray, toast-host, notify-action, middleware, preferences (anonymous read), i18n language read; each fix has a test that fails without it. Lint 0 errors, `tsc --noEmit` clean.
- Live on the box (headless): Back to Desktop lands on the desktop; the browser's Back closes the top window; a window beside the docked chat ends before it with Close reachable; the phone shelf shows the open app; the anonymous read answers the language alone and refuses everything wider; the login page greets in German from an English browser.

> CLAUDE.md in this PR carries the sentences for all four sweep PRs (#settings, #chat, #shell, #apps are the same file), so their merges cannot conflict on it; the sentences for the other three describe changes that land with them.

https://claude.ai/code/session_01SQddsPChtiBNnbFcZ9GHc2

## Summary by CodeRabbit

* **New Features**
  * Added localized desktop clock, date, internet-status, toast, and notification labels across supported languages.
  * UI language can now be read before sign-in.
* **Bug Fixes**
  * Mobile app shelves keep all open apps visible without overlapping controls.
  * New windows avoid docked chat areas while preserving saved and manually resized dimensions.
  * Improved browser Back-button handling.
* **Accessibility**
  * Toast actions and dismissal labels now use localized text.


yalexx:
Live on the box (build e26235f of the sweep branch, headless Chromium signed in as the owner, 16/16 checks): "Back to Desktop" from `/app/files` lands on the desktop, not Next's error page; the browser's Back closes the top window and stays on the desktop, with the history entry `{"clawbox":true,"__NA":true,…}` carrying Next's own fields beside the marker; a Files window opened beside an 858 px docked chat ends at 485 px with its Close reachable; the phone shelf draws Settings and the open apps in its middle row with no horizontal overflow; `GET /setup-api/preferences?keys=ui_language` answers `{"ui_language":"en"}` with no session while a wider read, `all=1`, a POST and an encoded second key all answer 401; with the box set to German the login page greets in German from an English browser, the power menu shows "Montag, 7. September" with the catalogue's internet line, and the language was put back afterwards.

PR #780, opened 2026-09-07
fix: say which agent was actually asked to reload its MCP servers

KrasimirKralev:
**TASK-763** — the MCP-reload refusal logging. Two of the card's three items were already on beta; this is the third, plus what the review pass found under it.

### Customer-visible symptom
An operator reading the journal is told the wrong thing about the one mechanism that rebuilds the agent's tool list:
- Two of the five refresh families said **"asked the agent to reload its MCP servers"** while the mechanism is HERMES' dashboard JSON-RPC and nothing else. On the dual SKU that is the harness answering even after the box has moved to OpenClaw, so someone chasing "the agent still thinks it is on Hermes" was told the OpenClaw child had been reloaded — it was never asked.
- Worse, the REFUSAL branch gated on the ACTIVE HARNESS as a proxy for "does this box have a dashboard". `install.sh` enables `clawbox-hermes-dashboard.service` for **hermes AND dual**, whichever harness is active — so on a licensed dual box sitting on OpenClaw, a dashboard that was up and answered `confirm_required` (the documented `approvals.mcp_reload_confirm` gate) had its refusal written as the benign *"this edition has no dashboard to ask — the tool list re-probes when the MCP server is next spawned"*. Both halves are false there: the dashboard exists and refused, and Hermes' MCP children are long-lived, so nothing re-probes them. A false success, on the line that exists to report one.
- The fifth family (`hermes-image-refresh`) wrote a line when the reload FAILED and nothing at all when it worked, so "the agent still refuses to draw" could not be told from "it was never asked".

### Cause and fix
One sentence, owned by the module that owns the mechanism: `MCP_RELOAD_ASKED` / `MCP_RELOAD_ALREADY` in `src/lib/hermes-mcp-reload.ts`, used by all five families. The refusal names the mechanism too, and its branch now asks the EDITION — `hasHermesHarness()`'s own rule (`hermes` or `dual`, the mirror of install.sh's `has_hermes_harness()`), from ONE `readEditionSource()` — with an edition nothing on the device named keeping the error, because a default is not an answer.

### Harness-first
The mechanism IS the harness's own: Hermes' dashboard JSON-RPC `reload.mcp`. Nothing here re-implements it; the change is what ClawBox says about it. On OpenClaw there is no equivalent by design — the ClawBox MCP server is spawned per session and reaped after ten idle minutes, so the tool list catches up on its own, which is exactly why the benign branch exists.

### Sibling call sites
- All five callers of `reloadMcpServers` now import the shared sentence, and the suite DERIVES that set from the source (every `src/lib/*.ts` that imports `reloadMcpServers` must import `MCP_RELOAD_ASKED` and be driven by a case) — a sixth family cannot be added silently, which a hand-kept list could not catch.
- Every test that mocked `@/lib/hermes-mcp-reload` with a bare factory now spreads the real module: a family importing a new value from it must not fail to import in a test that never exercises it.
- Five suites drove the refusal branch through the `getActiveHarness` mock and now drive it through the edition, which is what decides it.
- `hermes-image-refresh`'s dashboard BOUNCE line is a different event (it restarts the dashboard rather than asking it to reload) and is deliberately left as it is.
- The CodeQL half of the card is already on beta: alerts #516/#517/#518 (`js/log-injection` at `harness-mcp-refresh.ts:86`, `hermes-mcp-reload.ts:105,110`) are all in state `fixed` on `refs/heads/beta`, the closed-set rebuild (`HARNESS_NAMES`/`harnessName`) is there with its tests, and the two new constants are module literals that add no injection surface. Re-queried, not assumed.

### RED → GREEN
RED on unmodified beta (the wording half):

```
 ❯ |unit| src/tests/unit/mcp-reload-wording.test.ts (6 tests | 2 failed) 10ms
     × hermes/provider-refresh: says Hermes was asked, not "the agent" 5ms
     × coding-agent/mcp-refresh: says Hermes was asked, not "the agent" 1ms
Received: "[hermes/provider-refresh] the providers this box can be switched to gained anthropic; asked the agent to reload its MCP servers"
```

RED for the dual gate, with the branch put back to the active-harness form:

```
 ❯ |unit| src/tests/unit/mcp-reload-wording.test.ts (12 tests | 2 failed) 37ms
     × reports a DUAL box's refusal as an error, not as 'this edition has no dashboard' 4ms
     × stays a plain note on a box that has no dashboard by design 4ms
```

GREEN: the five refresh suites + the new wording suite + the configure route suite = 52 tests, then 12 in the wording suite after the review fixes; whole unit project 773 files, 12917 passed / 1 skipped (one unrelated pre-existing timeout flake in `telegram/status-hermes.test.ts`, `leaves the OpenClaw path alone` — it passes alone on this branch AND on beta, the TASK-702 class); `tsc --noEmit` clean; `eslint` on the touched files clean.

### Editions
Both. The change is precisely about telling them apart: `openclaw` keeps the benign note, `hermes` and `dual` get the error, and an unnamed edition keeps the error rather than guessing.

### Device
Not device-proven: journal wording on a code path CI exercises fully, no unit, path or edition-lock behaviour changed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved MCP reload handling across dashboard editions, including clearer behavior for Hermes, OpenClaw, dual, and unknown editions.
  - Standardized reload status messages to clearly indicate whether a reload was already completed or requested.
  - Clarified ownership and refresh reasons in MCP reload notifications.

- **Tests**
  - Added coverage for reload messaging, refusal behavior, edition-specific handling, and environment isolation.

### DIFF
--- a/src/lib/coding-agent-mcp-refresh.ts
+++ b/src/lib/coding-agent-mcp-refresh.ts
@@ -1,7 +1,7 @@
 import { MCP_RELOAD_ALREADY, MCP_RELOAD_ASKED, reloadMcpServers, reportMcpReloadRefused } from "@/lib/hermes-mcp-reload";
 
 /**
- * Ask the agent to rebuild its MCP tool list when the coding agent becomes
+ * Ask HERMES to rebuild its MCP tool list when the coding agent becomes
  * usable, or stops being.
  *
  * WHY THIS EXISTS. Three tools — `coding_agent_run`, `coding_agent_status` and
@@ -24,7 +24,7 @@ import { MCP_RELOAD_ALREADY, MCP_RELOAD_ASKED, reloadMcpServers, reportMcpReload
  * instance of one shape: #486 fixed it for `email_list`/`email_read` and #503
  * for the image tools, and this was the call site left out.
  *
- * The mechanism is the agent's own `reload.mcp`, shared with both siblings — see
+ * The mechanism is HERMES' own `reload.mcp`, shared with the four siblings — see
  * `hermes-mcp-reload.ts`. What belongs HERE is the rule for when it is worth
  * paying for, below.
  */
--- a/src/lib/coding-agent-mcp-refresh.ts
+++ b/src/lib/coding-agent-mcp-refresh.ts
@@ -1,4 +1,4 @@
-import { reloadMcpServers, reportMcpReloadRefused } from "@/lib/hermes-mcp-reload";
+import { MCP_RELOAD_ALREADY, MCP_RELOAD_ASKED, reloadMcpServers, reportMcpReloadRefused } from "@/lib/hermes-mcp-reload";
 
 /**
  * Ask the agent to rebuild its MCP tool list when the coding agent becomes
@@ -102,7 +102,7 @@ export async function refreshCodingAgentToolsIfReadinessChanged(
   if (options.alreadyReloaded) {
     // Nothing to ask for: the respawn that already happened in this request
     // rebuilt EVERY family's tool list, this one included.
-    console.log(`[coding-agent/mcp-refresh] ${became}; the MCP servers were already reloaded for this change`);
+    console.log(`[coding-agent/mcp-refresh] ${became}; ${MCP_RELOAD_ALREADY}`);
     return;
   }
   // `.catch` even though `reloadMcpServers` documents that it never throws: the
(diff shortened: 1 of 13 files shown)


KrasimirKralev:
Merged into beta (26718f11) after final review MERGE-OK on this head and green CI on this head, all review threads closed. Deploy of beta to both boxes follows; box proof lands on TASK-702 TASK-763 .

KrasimirKralev:
Merged into beta as 26718f11 on the Opus MERGE-OK for 4f4e1118 (CI green, 0 threads). Deploying beta head to both test boxes; proof follows on TASK-763.

PR #790, opened 2026-09-08
fix: share update progress across boot and route bundles

yalexx:
## Hardware-discovered V4 release blocker

During the actual main → beta update on flash-rig canary serial 1791626057877, the boot hook resumed `post_update` in PID 2914. A later GET /setup-api/update/status in that SAME PID loaded another bundle's updater instance, reported an interruption, and cleared the live lock while the root post-update service was still working. Its gateway was intentionally quiesced, not crashed.

Next's instrumentation and route handlers have independent module graphs. Module-local state/ownership/single-flight flags cannot represent a process-wide updater. Share those three fields through a versioned globalThis symbol. Disk continuation and interruption markers still bridge real process replacements/reboots; no failure is suppressed and no timeout/gate is removed.

Regression tests load two independent copies in the same process and exercise real entry points: both see the running snapshot, duplicate launch and dismiss are refused, and continuation reads share one promise. No network or root commands in these tests.

Validation:
- Both new tests fail on the previous code (2/2), pass with this change.
- Focused updater regressions: 127 passed.
- Full unit suite: 788 files; 13,216 passed, 1 skipped.
- TypeScript, targeted ESLint, diff check passed.

Hardware acceptance remains pending a new candidate build; this PR does not claim the canary or V4 release is complete. TASK-778.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved update coordination so status remains consistent across application components.
  - Prevented duplicate updates from running simultaneously.
  - Improved recovery after restarts, interruptions, completed updates, and failed rebuilds.
  - Preserved update warnings and progress across reboots.
  - Improved gateway recovery and drift detection so completion status reflects the latest system state.
  - Kept updates marked as running until completion and warning cleanup are saved; save failures now report a failed update.

### DIFF
--- a/src/lib/updater.ts
+++ b/src/lib/updater.ts
@@ -3817,6 +3817,7 @@ function launchUpdate(steps: UpdateStepDef[], startFrom: number, options: RunOpt
   runUpdate(steps, startFrom, options)
     .catch((err) => {
       console.error("[Updater] Unexpected error:", err);
+      runtime.state.error = err instanceof Error ? err.message : String(err);
       runtime.state.phase = "failed";
     })
     .finally(() => {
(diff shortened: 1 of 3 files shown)


yalexx:
The repository's TypeScript 5.9 check and CI compilation both passed for the inferred const symbol. To remove any ambiguity in that computed-property contract, be5ef447 explicitly annotates runtimeKey as unique symbol. Fresh `tsc --noEmit --incremental false` and the independent-module regression tests pass; runtime behavior is unchanged.

PR #791, opened 2026-09-08
fix: use PKCE for ClawKeep CLI pairing

yalexx:
## V4 pairing dependency — website #592 / TASK-750

The shipped `clawkeep pair` CLI still used authorization codes without proof of possession. Generate a fresh random verifier for each pairing, send only its S256 challenge through the portal/browser, and include the verifier in the token exchange body. The verifier is never printed or persisted. Existing state checks, loopback forwarding, token permissions and exchange behavior remain.

This is compatible with the old portal, which ignores the new fields, and enables the protection being deployed in website #592 without waiting for the legacy grace deadline.

Validation: full ClawKeep Python suite passed (169 tests). The new offline flow test verifies the challenge matches the exchanged verifier, fresh randomness across two pairings, unchanged state correlation, no token/verifier in console output, and mode0600 token storage. No real pairing or customer API operation ran.

## Summary by CodeRabbit

- **Security**
  - Pairing now uses OAuth 2.0 PKCE protection during authorization and token exchange.
  - Authorization requests and token exchanges validate the associated security parameters, including portal confirmation of PKCE verification.
  - Portal URLs must use HTTPS and cannot contain credentials, queries, or fragments; unsafe redirects are rejected.
  - Sensitive pairing credentials remain protected, are not displayed during the process, and are stored with restricted permissions.

### DIFF
--- a/clawkeep/clawkeep/pair.py
+++ b/clawkeep/clawkeep/pair.py
@@ -92,17 +92,28 @@ class _OneShotServer(socketserver.TCPServer):
     allow_reuse_address = True
 
 
+def _validate_server(server: str) -> None:
+    parsed = urllib.parse.urlsplit(server)
+    if parsed.scheme != "https" or not parsed.hostname or parsed.username or parsed.password or parsed.query or parsed.fragment:
+        raise SystemExit("Pairing requires an HTTPS portal URL without credentials, query or fragment.")
+
+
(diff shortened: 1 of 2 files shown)


yalexx:
All checks now pass, including installer CI; latest CodeRabbit approved c20ed237. Holding merge only for the explicit dependency: website #592 must be deployed first because the new CLI intentionally refuses a portal that does not acknowledge verified PKCE. This prevents shipping a temporary pairing regression on beta.

PR #799, opened 2026-09-09
fix: avoid redundant gateway boot consent loads (TASK-779)

yalexx:
## Summary

Fix TASK-779: repeated idempotent capability-consent commands delayed gateway readiness on a real Orin Nano past the chat startup budget.

- Inspect the core's current consent report before replaying enable; skip only positively adjudicated current consent.
- Leave named plugins without install records unchanged, without falsely clearing repair markers or declaring consent.
- Preserve pending/unknown/error repair paths; recheck once after the first killed enable to recognize consent actually written.
- Invalidate the snapshot after payload replacement. Do not persist consent decisions across boots.

## Verification

- Reproduced on Nano .61 combined beta 2f50db37: two 60-second no-op enables plus slow core startup led to a failed chat connection.
- New fast-path regression assertions failed on beta, then passed with the fix.
- Dedicated Nano .4: all 23 boot-script suites / 596 tests pass; bash syntax passes.
- Real .61 restart timing and UI reconnect are being measured; results will be appended before merge.

Beta only. No main promotion and no runtime tests on the production host.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved startup handling for plugin capability consent.
  - Added an initial preflight check to identify current consent status before enabling managed plugins.
  - Refreshes consent information after plugin installation or interrupted enable attempts.
  - Avoids redundant enable operations when consent is already current.
  - Improved handling of plugin aliases, reinstall scenarios, timeouts, refusals, and unknown consent states.

### DIFF
--- a/scripts/gateway-pre-start.sh
+++ b/scripts/gateway-pre-start.sh
@@ -3894,33 +3894,20 @@ clawbox_plugin_boot_without() {
 # existing marker exactly as it found them, for the next boot or the Settings
 # Retry to resolve.
 
-# The consent question is asked at most ONCE per boot, for every id at once:
-# the CLI start is the dominant cost and this runs inside a blocking
-# ExecStartPre. Empty until the first kill asks for it.
+# Query the core BEFORE replaying an expensive, normally idempotent enable.
+# One snapshot serves all managed ids until a payload install invalidates it.
+# Only the core's positive installed-record verdict skips consent; a named
+# unindexed plugin is left untouched (not declared accepted), as above.
+# Unknown, pending and failed reports still take the existing repair path.
 #
-# ONE SNAPSHOT, TAKEN AT THE FIRST KILL OF THE BOOT, and memoised. A consent
-# recorded by a LATER killed verb is not in it, so that plugin is still named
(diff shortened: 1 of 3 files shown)


yalexx:
Real Nano .61 verification at12:06 Sofia:

- Patched only the disposable Nano boot script (backed up original); Next app remains combined beta2f50db37.
- Pre-start12:06:23→12:06:33 (~10s); logs positively report Codex consent current and skip enable. Unindexed email plugin left unchanged, not falsely accepted.
- Core HTTP listening12:06:45; ready12:06:46 (~24s after restart). Independent HTTP probe200 at27.33s.
- Existing real chat UI reconnected automatically and returned exact `RESTART CHAT READY` using local Gemma.
- Baseline earlier had two60s no-op consent deadlines and a slow167s core initialization. This run proves the redundant pre-start work is removed; it does not attribute every earlier core delay to those checks.
- Nano4 complete boot-script regression battery:596 passed across23 suites.

Main untouched. Evidence in release/v4-2026-09-08/nano61-consent-fastpath-restart.{log,json}, nano61-post-fastpath-chat-ui.txt and nano4-consent-regression.log.

PR #800, opened 2026-09-09
fix: contain ClawKeep restore-picker keyboard focus (TASK-780)

yalexx:
## Summary

Fix TASK-780, reproduced in the real ClawKeep UI on Nano .199 / combined beta2f50db37:

- Restore picker left focus on the underlying backup input.
- Escape reached the background ChatPopup as well, closing both surfaces.

Use the existing shared modal-dialog hook on the picker panel. Focus enters/stays inside; the background becomes inert; Escape stops at the picker; focus returns to the trigger. Keep the portal outside the transformed desktop window. No changes to restore or archive mutations.

## Verification

- New component regression fails on beta (focus remains outside).
- Nano .4: all34 ClawKeep component tests pass with the patch, including focus restoration and no leaked Escape.
- Added browser assertions for Shift-Tab/Tab wrap, no window Escape, and restored trigger focus. Existing stacked-confirm regression retained.
- Dedicated Nano production build and browser sweep running; results will be appended before merge.

Beta only. No production-host runtime tests or main promotion.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved restore dialog keyboard accessibility with focus trapping and reliable focus restoration after dismissal.
  * Escape now closes the restore dialog without triggering unrelated global keyboard handlers.
  * Background content is appropriately disabled while the dialog is open.

* **Tests**
  * Added coverage for keyboard navigation, Escape handling, focus restoration, and background interaction behavior.

### DIFF
--- a/e2e/clawkeep-interactions.spec.ts
+++ b/e2e/clawkeep-interactions.spec.ts
@@ -166,7 +166,8 @@ test("restore modal opens, fetches snapshots, and Esc dismisses it", async ({ pa
   await page.evaluate(() => {
     (window as Window & { restoreEscapeLeaks?: number }).restoreEscapeLeaks = 0;
     window.addEventListener("keydown", (event) => {
-      if (event.key === "Escape") (window as Window & { restoreEscapeLeaks: number }).restoreEscapeLeaks++;
+      const state = window as Window & { restoreEscapeLeaks?: number };
+      if (event.key === "Escape") state.restoreEscapeLeaks = (state.restoreEscapeLeaks ?? 0) + 1;
     });
   });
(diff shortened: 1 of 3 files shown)


yalexx:
Nano4 exact-source verification complete:

- Frozen install and production build (including TypeScript) PASS.
- ClawKeep components:34/34 PASS; new focus/Escape regression was RED on beta.
- ClawKeep interaction browser suite:7/7 PASS against the production build, including Shift-Tab/Tab wrap, Escape isolation, trigger focus restoration, and stacked restore confirmation.
- No live archive mutation or production-host app/runtime involved.

Evidence: workspace/release/v4-2026-09-08/nano4-restore-focus-{components,build,e2e}.log.

PR #803, opened 2026-09-09
fix: check the Hermes model picker's list once a day by itself

KrasimirKralev:
**TASK-781** — Hermes picker: refresh the provider model list by itself daily, show its age, and drop exhausted duplicate grants.

## What the owner saw

On a Hermes box the Anthropic picker was missing `claude-opus-5` and `claude-fable-5-1` while the very same credential listed them live. The only way to see them was to click **Refresh** (or re-sign-in). Measured on the box (Hermes Agent v0.21.1): the picker served eleven ids written at **12:14**; `GET /setup-api/hermes/models?refresh=1` re-fetched and the list became **thirteen**, with both new models present.

## Cause

Nothing on the box ever forces the harness to re-read what each provider serves. Three layers, none of which recovers on its own:

- **Hermes' `provider_models_cache.json`** — `_PROVIDER_MODELS_CACHE_TTL` is 1 h, but `_PROVIDER_MODELS_STALE_SERVE_MAX` is **seven days**, and `_spawn_swr_refresh` only calls `_store_cache_entry` when the live fetch comes back non-empty. One failed re-fetch carries the old ids forward with the entry's `at` unmoved.
- **ClawBox's own layer** (`hermes-model-options.ts`) — its background `load(false)` re-asks the dashboard but carries no `refresh=true`, so it re-reads that same disk row rather than busting it.
- **`?refresh=1`** is the only path that sends `refresh=true`, and it is session-gated. A human had to click.

So the list only ever moved when the owner clicked — which is the defect: a user had to know to click Refresh to see a model that shipped a week ago.

## Harness-first: the native mechanisms, named

- **Forced refresh — used.** `GET /api/model/options?refresh=true` on the Hermes dashboard is the harness's own `clear_provider_models_cache` + live per-provider `/v1/models` sweep. ClawBox already had a verified client for it (`fetchFromDashboard`); this PR only drives it on a schedule. Nothing is re-implemented.
- **`hermes model --refresh` — exists, unusable from a server.** `hermes model --help` on v0.21.1: *"Wipe the model picker disk cache and re-fetch every provider's live /v1/models list"* — but `hermes model` is the **interactive picker** ("Interactively select your inference provider and default model"), so `--refresh` is a modifier on a TUI, not a standalone command. The module header already recorded this; re-confirmed on v0.21.1.
- **A periodic refresh inside Hermes — none exists.** `prewarm_picker_cache_async()` has exactly two non-test call sites, `tui_gateway/entry.py:269` and `cli.py:3662`, both at interactive startup. There is no timer, no daemon, nothing that re-reads a catalogue on a cadence.
- **Unattended driver — ClawBox's own established pattern.** `clawbox-heartbeat.timer` → `GET /setup-api/portal/heartbeat-tick`, every 5 min, as the `clawbox` user, with the per-install internal token. No new unit, no new route, no owner cookie. `applyDeferredLanguagePersona()` already rides this tick, with the rationale written in the route: *"the heartbeat is the only thing on a running box that fires without anyone touching it"*.
- **Per-provider catalogue age from the harness — not available.** The picker row Hermes builds is `{slug, name, is_current, is_user_defined, models, total_models, source}` (`hermes_cli/model_switch_providers.py:635-639`) — **no timestamp**. That is a genuine harness gap: ClawBox cannot report the true age of the list, only the age of its own last successful check, and the fields are named for exactly that (`catalogCheckedAt`, not `catalogRefreshedAt`).

## The change

1. **Daily unattended check.** `refreshCatalogIfDue()` rides the heartbeat tick: at most once per 24 h, at most one driver retry per hour when it fails, **fire-and-forget** (the unit runs `curl --max-time 10` and the dead-tunnel repair below already budgets 7.5 s of that for DNS, so awaiting it would truncate the tick on exactly the pass that repairs the tunnel), never throwing, and a no-op unless the active harness is Hermes.
2. **The list's age in the payload.** `catalogCheckedAt` and `catalogStale` on both the unscoped and the scoped reply. `fetchedAt`/`stale` could not answer this — they describe ClawBox's own 60-second L1 cache, so a week-old catalogue reported `fetchedAt` two seconds ago and `stale: false`. Computed per request, never baked into the cached payload (a payload built at 23 h old would otherwise still call itself fresh two hours later), and read **after** the refresh so a `?refresh=1` does not answer "stale" over the one action that fixes staleness.
3. **Unknown is not "old".** A box that has never recorded a check reports `null`, not `true` — the rule this route already follows one field away, where `verified` stays null for a provider nothing has exercised rather than reporting `false`. Without it a box in its first five minutes, one restored from a backup, and a dual box running OpenClaw all reported the list a day stale.

## Requirement (3) — exhausted duplicate grants: measured, and deliberately NOT auto-pruned

The card asks for exhausted/revoked duplicate grants to be pruned "via Hermes' native auth commands if they exist (name them)". They exist, and auto-using them would be wrong.

**Native commands:** `hermes auth list | remove <provider> <target> | reset <provider> [target] | refresh <provider> [target] | priority <provider> <target> <n> | status <provider>`.

**Why ClawBox must not call `remove` on its own:**
- `exhausted` is Hermes' **recoverable** bucket by design; `dead` is the permanent one (`credential_pool.py:78-84`). Deleting an exhausted row deletes a credential the pool is actively planning to bring back.
- A rate-limited credential is **indistinguishable** from a revoked one in stored state. On the measured box the exhausted row carries `last_error_code: null`, `last_error_reason: null`, `last_error_reset_at: null` — `hermes auth list` prints only `exhausted (43m 9s left)`, and that countdown is computed at read time, never stored. Auto-deleting on that signal deletes healthy rate-limited credentials.
- The delete is irreversible for manual sources: `remove_index` pops the entry and persists with `removed_ids`, taking the OAuth refresh token with it. The box's priority-0 row is `manual:dashboard_pkce`; nothing re-seeds a `manual:*` source, so removing one forces a full re-auth.
- For the row in question it would also be pointless churn: `hermes_pkce` is re-created by `_seed_anthropic_singletons` from `~/.hermes/.anthropic_oauth.json` on the next `load_pool`, so `remove` there is an undocumented way of doing `auth reset`.
- Hermes already self-prunes: DEAD `manual:*` entries go after `DEAD_MANUAL_PRUNE_TTL_SECONDS` (24 h).

**And on this box the premise does not hold anyway.** Both Anthropic pool entries are OAuth, and the OAuth resolver `_resolve_anthropic_pool_token` enumerates with `_available_entries`, which skips `STATUS_EXHAUSTED` while `now < exhausted_until` — so the cooling-down entry cannot be picked for a catalogue fetch. There **is** a status-blind path (`_resolve_anthropic_pool_catalog_credentials`, `models.py:1598`, a raw `read_credential_pool` that checks no `last_status`), but it only fires for an `api_key` row, which this box does not have. Real in code, latent here.

So: named, measured, and no auto-prune. Doing it safely means acting on `dead` only — which Hermes already does for itself.

## OpenClaw edition (requirement 4) — checked, and the same shape does not exist

`src/app/setup-api/ai-models/catalog/route.ts:1601-1632` already force-refreshes **on the read path** when its cache passes `REFRESH_INTERVAL_MS` (6 h) — no session, no button, no `?refresh=1` required (`?refresh=1` there is deliberately demoted to a nudge). The "only an owner click ever busts the cache" defect is Hermes-only. Nothing changed for OpenClaw beyond the driver no-opping there.

## RED → GREEN

RED, on unmodified `origin/beta` (9694b9a8), the final test file copied into a clean beta worktree — **9 of 11 fail**:

```
 ❯ src/tests/routes/hermes/catalog-daily-refresh.test.ts (11 tests | 9 failed)
     × forces a re-fetch on the heartbeat when the list was last refreshed a day ago
     × reports the catalogue's age, and calls it stale past a day
     × says "not checked" — never "old" — for a catalogue nothing has refreshed yet
     × calls a catalogue refreshed within the day fresh
     × carries the age onto the scoped reply the picker reads
     × does not remember a refresh that failed, and retries on a floor rather than every tick
     × remembers a check that succeeded, and serves the models it brought back
     × counts the owner's own Refresh click as a refresh
     × says which of the five things it did

AssertionError: expected [] to have a length of 1 but got +0
 ❯ ...:189  await vi.waitFor(() => expect(forcedRefreshCalls()).toHaveLength(1));

AssertionError: expected undefined to be 1788857153318
 ❯ ...  expect(body.catalogCheckedAt).toBe(checkedAt);
```

(The two that pass on beta are the negatives — the OpenClaw no-op and "a failed owner Refresh must not throttle the driver" — both trivially true where nothing writes the key at all.)

GREEN, on the branch:

```
 Test Files  1 passed (1)
      Tests  11 passed (11)
```

Neighbouring suites (`src/tests/routes/hermes`, `heartbeat-tick`, the four `hermes-model-options-*` units):

```
 Test Files  41 passed (41)
      Tests  453 passed (453)
```

Whole suite, `tsc --noEmit` and `eslint` on the changed files all clean:

```
 Test Files  999 passed (999)
      Tests  15053 passed | 1 skipped (15054)
```

## Sibling call sites

Every `getModelOptions` caller was swept. Only `hermes-cloud-provider.ts:141` passes `{refresh:true}` besides the models route; `provider-mcp-refresh.ts:99`, `provider-status.ts:287`, `hermes/chat/route.ts:1027` and the route's own POST pass no `refresh`, so the bookkeeping cannot fire from them. Both forced-refresh paths now go through one helper, so the owner's Refresh click records the check too — otherwise a box the owner had just refreshed by hand still reported its list a day stale. `ScopedModelsReply` has exactly one producer (the models route, updated); the MCP reader (`mcp/tools/ai.ts`) reads a subset and is unaffected by the additive fields.

## Review

One review (`clawbox-review`), findings applied: the age being read *before* the refresh (a successful `?refresh=1` answering `catalogStale: true`); the awaited call pushing the tick past `curl --max-time 10` on the dead-tunnel repair path; `isLiveAnswer` naming a 200 from the dashboard "the list was re-fetched" (fixed by naming the field for what it measures and reporting the harness gap above); "never checked" reported as stale; the read-modify-write of `data/config.json` without the house `createSerialLock`; `attemptedAt` stamped from the owner's click, letting a hammered Refresh suppress the automatic retry for an hour; and two throttle tests that passed with their gates deleted (they leaned on the pre-existing 10 s explicit-refresh window — the gates are now driven with an explicit clock).

## Known gap, deliberately not in this PR

No surface renders the age yet — this PR delivers the API contract. A line under the model select needs a new i18n key across all ten locales, and it belongs with a design for what "checked 3 hours ago" should say.

## Summary by CodeRabbit

* **New Features**
  * The Hermes model catalogue now refreshes automatically about once per day.
  * Model catalogue responses now show when the catalogue was last checked and whether it may be outdated.
  * Manual catalogue refreshes are tracked, and failed refresh attempts are handled without blocking heartbeat updates.

* **Bug Fixes**
  * Improved retry handling after unsuccessful catalogue refreshes.

### DIFF
--- a/src/app/setup-api/hermes/models/route.ts
+++ b/src/app/setup-api/hermes/models/route.ts
@@ -9,11 +9,13 @@ import { requireSession } from "@/lib/route-auth";
 import { reconcileClawaiModelsWithHermes } from "@/lib/hermes-clawai";
 import { reconcileLocalAiWithHermes } from "@/lib/hermes-local-ai";
 import {
+  catalogStaleness,
   getModelOptions,
   invalidateModelOptions,
   isAllowedProvider,
   isPairAllowed,
   isSafeModelId,
+  readCatalogRefreshMark,
   shouldEnforcePairing,
   scopeFromPayload,
   type HermesModelOption,
@@ -100,6 +102,25 @@ export async function GET(request: Request) {
     await reconcileClawaiModelsWithHermes();
   }
 
+  // When this box last got the harness to re-read its providers' lists, as
+  // opposed to how old this process's copy of the answer is. `fetchedAt`/
+  // `stale` below answer only the second question — they are this server's
+  // 60-second L1 cache — so a list Hermes has been carrying forward for a week
+  // reports `fetchedAt` two seconds ago and `stale: false`. TASK-781.
+  //
+  // READ AFTER `getModelOptions`, never beside it. `?refresh=1` performs the
+  // check inside that call, so a mark read up here would be the PRE-refresh one
+  // and the one action that fixes staleness would answer `catalogStale: true`
+  // — a false failure over an operation that had just succeeded.
+  //
+  // Read per request rather than baked into the cached payload, for the other
+  // direction: a payload built at 23 h old would otherwise still call itself
+  // fresh two hours later, from the L1 cache.
+  const catalogAge = async () => {
+    const mark = await readCatalogRefreshMark();
+    return { catalogCheckedAt: mark.checkedAt, catalogStale: catalogStaleness(mark) };
+  };
+
   try {
     if (provider) {
       const scoped = await getModelOptions({ refresh });
@@ -115,6 +136,7 @@ export async function GET(request: Request) {
         ...(await scopeFromPayload(scoped, provider)),
         reasoning: scoped.reasoning,
         savedPair: scoped.current,
+        ...(await catalogAge()),
       };
       return NextResponse.json(reply);
     }
@@ -171,6 +193,7 @@ export async function GET(request: Request) {
       source: payload.source,
       stale: payload.stale,
       fetchedAt: payload.fetchedAt,
+      ...(await catalogAge()),
       ...(payload.degraded ? { degraded: payload.degraded } : {}),
       ...(refreshDenied ? { refreshDenied: true } : {}),
     });
--- a/src/app/setup-api/portal/heartbeat-tick/route.ts
+++ b/src/app/setup-api/portal/heartbeat-tick/route.ts
@@ -3,6 +3,7 @@ import { NextResponse } from "next/server";
 import { readTunnelUrl, startTunnelService } from "@/lib/cloudflared";
 import { isInternalRequest } from "@/lib/internal-token";
 import { applyDeferredLanguagePersona } from "@/lib/language-persona";
+import { refreshCatalogIfDue } from "@/lib/hermes-model-options";
 import { pushHeartbeatTick } from "@/lib/portal-heartbeat";
 import { requireSession } from "@/lib/route-auth";
 import { checkTunnelLiveness, markRestarted, mayRestart } from "@/lib/tunnel-liveness";
@@ -62,6 +63,24 @@ export async function GET(request: Request) {
   // applyDeferredLanguagePersona), and it never throws, so it cannot turn a
   // tick into the 500 that would make the systemd unit flap.
   await applyDeferredLanguagePersona();
+
+  // Keep the model picker's list from ageing out, for the same reason and on
+  // the same ride: nothing else on the box forces the harness to go and re-read
+  // what each provider actually serves. Hermes' own caches will carry a list
+  // forward for up to seven days, and ClawBox only ever busts them when the
+  // owner clicks Refresh — so `claude-opus-5` and `claude-fable-5-1` were
+  // missing from a box whose credential listed them live (TASK-781).
+  //
+  // NOT AWAITED, deliberately, like `pushHeartbeatTick` below. This unit runs
+  // `curl --max-time 10`, and the dead-tunnel path underneath already spends up
+  // to 7.5 s of that on DNS (see tunnel-liveness.ts) — putting a dashboard
+  // round-trip in front of it would push the tunnel repair past curl's deadline
+  // on exactly the tick that performs it, and `SuccessExitStatus` would swallow
+  // the truncation. `refreshCatalogIfDue` never rejects, so there is no
+  // unhandled rejection to catch; it decides in two cheap reads on the 287
+  // ticks a day that are not the one, and no-ops on an OpenClaw box, whose
+  // catalogue route already re-enumerates by itself on the read path.
+  void refreshCatalogIfDue();
 
   const tunnelUrl = await readTunnelUrl();
   const liveness = await checkTunnelLiveness(tunnelUrl);
--- a/src/lib/hermes-model-options.ts
+++ b/src/lib/hermes-model-options.ts
@@ -31,7 +31,9 @@ import {
   hermesDashboardUnitState,
   type HermesDashboardUnitState,
 } from "@/lib/hermes-dashboard-control";
-import { get } from "@/lib/config-store";
+import { get, set } from "@/lib/config-store";
+import { getActiveHarness } from "@/lib/harness";
+import { createSerialLock } from "@/lib/serial-lock";
 import { CLAWBOX_AI_CHAT_MODEL_IDS } from "@/lib/clawbox-ai-models";
 import { pickedClawboxAiModelIdAmong, readExplicitModelPicks } from "@/lib/explicit-model-pick";
 import {
@@ -205,6 +207,16 @@ export interface ProviderScope {
 export interface ScopedModelsReply extends ProviderScope {
   reasoning: string;
   savedPair: { provider: string; model: string };
+  /**
+   * When this box last got the harness to re-read every provider's list, and
+   * whether that was over a day ago — NOT the age of this process's copy, which
+   * is what {@link ProviderScope.fetchedAt} and `stale` describe. Both are
+   * `null` where nothing has recorded a check — not checked, never "old".
+   * See {@link CatalogRefreshMark} for what a check does and does not prove.
+   * TASK-781.
+   */
+  catalogCheckedAt: number | null;
+  catalogStale: boolean | null;
 }
 
 // ── L1 cache (in-process, SWR) ───────────────────────────────────────────────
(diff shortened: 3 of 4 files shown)


KrasimirKralev:
## Device proof — Hermes box, under the device lock

Branch head `0f6ac96d` fetched onto the box and its three source files applied over the box's pinned `beta` head (`60a62fe0`); the three files are byte-identical between `60a62fe0` and `origin/beta`, so nothing else moved. Built (`rc=0`, 96 s, BUILD_ID `dLz5zXYi1dNY8tG2RHBsI` → `Fp-uUlRm-CMPq2Ils2LRI`), `clawbox-setup` restarted, then restored and rebuilt back to beta. Edition `hermes`, Hermes Agent v0.21.1 (2026.9.7) — untouched.

### The fix, on hardware

The mark aged to 25 h so the daily job is due, then the heartbeat tick handler fired **once**, with no `refresh` parameter anywhere and no Refresh click:

```
=== hermes cache BEFORE ===
  at=14:02:34 models=13
=== fire the heartbeat tick ONCE — no refresh param, no Refresh click ===
  tick HTTP 200 in 0.017347s
=== hermes cache AFTER ===
  at=14:03:28 models=13
  claude-opus-5: True | claude-fable-5-1: True
=== clawbox mark AFTER ===
  checkedAt = 14:03:28
=== what the picker now serves (plain GET, no refresh) ===
  catalogStale=False catalogCheckedAt set=True
  anthropic models=13, opus-5=True, fable-5-1=True
```

Hermes' own `provider_models_cache.json` entry for `anthropic` advanced `14:02:34 → 14:03:28`, i.e. the harness really did re-read the provider's live list because `refresh=true` reached it. The tick itself returned in **17 ms** — the fire-and-forget change means the catalogue check cannot eat the unit's `curl --max-time 10`, which the dead-tunnel repair below it already budgets 7.5 s of.

### The defect, on the same hardware

Restored to beta (BUILD_ID `ROBz_M1DXqJe23X0ODXcN`, `build-info: commit 60a62fe0…, branch beta, dirty False`), a **real** systemd heartbeat tick fired at 14:10:27:

```
  hermes anthropic at = 14:05:17 (was 14:05:17)
  clawbox key present: False | keys: 55
```

Nothing moved. That is the reported behaviour: on beta the box never asks the harness to re-read anything, so the list sits until a human clicks Refresh.

### Measured limitation, recorded rather than papered over

The **first** check after a web-server restart can record `checkedAt: null` even though the harness did refresh. Observed once here: the tick fired 22 s after the restart, Hermes' `at` advanced (13:48:25 → 14:01:41) while ClawBox logged only an attempt. Cause is the cold-process dashboard login, which `hermes-model-options.ts` already documents as carrying no timeout signal, running inside `DASHBOARD_TIMEOUT_MS` (8 s); the plain GET at that moment returned the 2-provider disk fallback. Steady state is not close to the bound — a forced sweep of 56 providers measured **1.77 s**, `source: dashboard`:

```
=== TIMED forced refresh ===
[HTTP 200 in 1.771245s]
{'source': 'dashboard', 'stale': False, 'catalogCheckedAt': 1788951754556, 'catalogStale': False, 'degraded': None}
providers: 56 models: 33
```

So the behaviour is self-correcting (the retry an hour later succeeds) and never reports a stale list as fresh — it under-reports a check that landed, which is the safe direction. Not fixed here because the fix means threading a per-call timeout through `fetchFromDashboard`/`buildPayload`/`load`, i.e. late churn in the hot path of a function this PR otherwise leaves alone, on an already-green and already-reviewed head. Fix direction for a follow-up: give the unattended path its own, longer dashboard budget — nothing waits on it.

### Restored, and proved

```
HEAD:        60a62fe0433b0ba45853630debf9e31d919973cb   (unchanged)
branch:      beta                                       (unchanged)
tree:        0 modified                                 (clean, as found)
update pin:  beta                                       (unchanged)
build-info:  commit 60a62fe0…, branch beta, dirty False
config keys: 55 | hermes_catalog_refreshed_at present: False   (as found; 0600 clawbox:clawbox)
unit:        clawbox-setup active
```

`clawbox-gateway` is `inactive`/`masked` — the pre-existing state on this Hermes-only SKU, not touched by this run. Device lock taken at 13:57:51 and released; no channel, email, reset or reboot touched.

KrasimirKralev:
Merged into beta as 936b4aa0 on the Opus MERGE-OK for 0f6ac96d (CI 11/11, CodeRabbit approved, 0 threads). Deploying beta head to the Hermes test box now; proof follows on TASK-781.

PR #806, opened 2026-09-09
Five fixes found upgrading a box from main to beta (v3.9.0 → v4.0.0)

yalexx:
Found by upgrading a real Jetson Orin from `main` (v3.9.0) to `beta` (v4.0.0) and then driving Coding Agent, Memory Shard and ClawKeep through the web UI. The upgrade itself passed all 13 steps cleanly; everything here is a defect the upgrade *exposed*.

## The fixes

**1. System Update and Remote Desktop icons never left upgraded desktops.**
`OFF_DESKTOP_BY_DEFAULT` only shapes the *default* grid, and the loader restores a saved `desktop_apps` list verbatim — deliberately, so an owner's additions survive. So moving an app off the desktop only ever reached fresh boxes; any box that had saved a list kept the icon forever, short of a factory reset. Shed once, keyed by version so a later bump cannot take back an icon this version already shed and the owner has since restored.

**2. A subscribed box defaulted to the local voice.**
`step_openclaw_tts` installed Kokoro and then selected it unconditionally. Measured on the box: **cloud 2.6s vs Kokoro 21.4s** — Kokoro's server stops after five idle minutes, so the first utterance after a quiet spell pays a cold start. Now defaults to the ClawBox AI cloud voice when the box has one; Kokoro stays installed and one click away. The subscription is proved from our own `clawboxManaged` stamp rather than re-reading the plan, because `gateway-pre-start.sh` already applied that gate.

**3. The chat greeted itself on any empty transcript.**
An empty transcript is not a first conversation — a cleared conversation, a reset session, a restored side tab and a box introduced months ago all look identical, and every one got an unasked-for `"hi"` that spent a model turn and put a word in the owner's mouth. Now gated on the agent's own `BOOTSTRAP.md` (the ritual OpenClaw arms and deletes), threaded through `HarnessFacts` so browser and server decide by one function. Implemented as an effect because its two inputs race: on OpenClaw the socket calls `loadHistory()` before the capabilities land, so deciding inside that read left a genuinely fresh box silent for good.

**4. ClawKeep's wizard dismissed itself by succeeding.**
It announced "STEP 1 OF 3", and pairing — step 1 — falsified the app's display condition, dropping the owner on the dashboard at "AT RISK — Protection Lapsed" with `setupComplete` still false, two steps early. The front door asked two questions where `BrowserApp`, `CodingAgentApp` and `MemoryShardApp` all ask one; the legacy case its extra `!paired` conjunct existed for now lives in `getClawKeepSetupComplete`. Also: both on/off switches were 1×1px `sr-only` inputs behind a styled track, so a pointer aimed at the control hit a wrapper `div` — they are now transparent inputs sized to the switch, and ClawKeep's gained the accessible name it never had.

**5. "Applying system fixups" hid fifteen minutes of work behind four words.**
On this Jetson that one step spent ~6 minutes compiling CTranslate2 with CUDA in total silence at 4×99% CPU, plus a ~300MB PyTorch download, Kokoro, spacy and faster-whisper. A healthy update was indistinguishable from a hung one. `StepState` gains an optional `detail` carrying install.sh's own headline, read by a **read-only** journal watcher — deliberately not a change to `execAsRoot`, whose budgets, overrun handling and gateway quiescing should not be re-plumbed for a progress line.

## Testing

- **Full suite: 1000 files, 15,070 passed, 0 failures** attributable to this branch.
- Verified end to end on hardware: a real Coding Agent run (wrote a file, drove the device's Chromium, committed `b25a9c3`); Memory Shard indexing a probe document (files 2→3, vectors 3→4, embedded locally); ClawKeep paired, sealed, backed up (`566,832` bytes encrypted) with auto-backup armed.
- A `/simplify` pass caught three errors in this branch's own work — a bogus mock property, a factually wrong comment, and an incomplete first attempt at fix 4 — all corrected here.

## Known, deliberate

- **Pre-existing flakiness:** a handful of component tests fail under full-suite parallel load and pass in isolation. Verified on unmodified `origin/beta`, which fails the same way — not introduced here.
- **Deferred:** the progress watcher spawns `journalctl` every 4s (measured at ~0.15% of the box — real but not contention). A single `journalctl -f` would cut it to one spawn with better latency; left as a follow-up rather than re-plumbing the mechanism in this PR.
- **Not done:** OpenClaw 2026.9.3 — it drops Node 22 while the box runs 22.23.2, and the installer's engine guard would not catch it. That is a Node 22→24 migration, not a version bump.

## Summary by CodeRabbit

- **New Features**
  - Automatically selects managed cloud voice services when available, with a verified local fallback.
  - Shows live details for the currently running system update step.
  - Removes outdated default app placements from saved desktops once while keeping those apps available through the launcher.
- **Bug Fixes**
  - Preserves existing unreadable voice configuration instead of overwriting it.
  - First-conversation greetings appear only when an introduction is pending.
  - Setup wizard visibility is maintained for paired devices.
  - Schedule switches respond reliably across supported apps.
  - Legacy paired devices retain completed setup status.

### DIFF
--- a/e2e/chat-popup.spec.ts
+++ b/e2e/chat-popup.spec.ts
@@ -180,6 +180,45 @@ test("chat popup connects, streams a reply, and supports panel docking", async (
   await expect(page.getByTitle("Dock to right")).toBeVisible();
 });
 
+test("chat popup stays silent on a box whose agent has been introduced", async ({ page }) => {
+  // The common case on a real box, and the one the greet used to get wrong: an
+  // empty transcript is not a first conversation. A cleared conversation, a
+  // reset session and a box introduced months ago all look identical, and every
+  // one of them used to be answered with an unasked-for "hi" that spent a model
+  // turn and put a word in the owner's mouth.
+  await installFakeGatewaySocket(page);
+
+  await installClawboxMocks(page, {
+    timeoutCapMs: 300_000,
+    kvEntries: { "clawbox-mascot-hidden": "1" },
+    // No introduction waiting: BOOTSTRAP.md is gone because the ritual finished.
+    chatFacts: { onboardingArmed: false },
+    initialSetup: {
+      setup_complete: true,
+      wifi_configured: true,
+      update_completed: true,
+      password_configured: true,
+      ai_model_configured: true,
+      telegram_configured: true,
+    },
+    preferences: { ui_mascot_hidden: 1 },
+  });
+
+  await page.goto("/");
+  await expect(page.getByTestId("desktop-root")).toBeVisible();
+
+  await openChatPopup(page);
+
+  // The composer is usable and the transcript is empty: nothing was sent, so
+  // the fake gateway — which answers "hi" and nothing else with that line —
+  // never replied. Asserted against the gateway's OWN reply rather than a
+  // generic empty check, so a turn that went out under any other text still
+  // fails this.
+  await expect(page.getByTestId("chat-composer-row")).toBeVisible();
+  await expect(page.getByText("Hello from the fake gateway")).toHaveCount(0);
+  await expect(page.getByText(/Fake gateway heard:/)).toHaveCount(0);
+});
+
 test("chat popup lets you switch to Local AI when it is configured", async ({ page }) => {
   await installFakeGatewaySocket(page);
 
--- a/e2e/chat-popup.spec.ts
+++ b/e2e/chat-popup.spec.ts
@@ -60,6 +60,12 @@ async function installFakeGatewaySocket(page: Page) {
         }
 
         if (message.method === "chat.history") {
+          // Counted so a test can wait for the transcript read to have HAPPENED
+          // before asserting that nothing was sent. Without that, "no turn went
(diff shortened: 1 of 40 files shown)


yalexx:
@coderabbitai review

Requesting a review of `c1f44dba`, which is not yet covered — the most recent review is against `7e4e5cdb`.

Since that review, this branch addresses both of its outside-diff findings:

1. **`tts_config_readable` used exit 1 as its verdict.** An uncaught Python exception and a refused `sudo` both exit 1, so a plumbing failure was indistinguishable from "this config does not parse" — and the caller skips provider definition *and* selection, so a fresh device could finish with no TTS provider. The verdict is now a distinct code (3); 1 means "could not look" and no longer refuses.

2. **The managed-cloud path was never executed.** Every harness stub answered only `…provider`, so `tts_managed_cloud_provider` always saw an empty provider map and each run took the on-device branch. The registry stub now serves a provider map, and three tests cover cloud selection, an owner's own unstamped `openai` route, and the fallback when the cloud entry will not select.

PR #855, opened 2026-09-13
Finish PR 851 on this project (branch clawbox/run-erwwu00y, base beta).…

yalexx:
Opened by the ClawBox coding agent.

**Task**
Finish PR 851 on this project (branch clawbox/run-erwwu00y, base beta). Do NOT open a new PR: push to that branch and reply on each review thread saying what changed.

Three CodeRabbit findings are open on the current head. Read them with `gh api graphql` reviewThreads on PR 851 and address each:

1. MAJOR, src/lib/coding-project-delete.ts — canonicalize the root before deduplicating trash directories. `physicalTrashDirs` falls back to the lexical trash path when realpath returns ENOENT, so symlinked roots can produce duplicate entries. Fix it and prove it with a test using a symlinked root.
2. MINOR, CLAUDE.md — the documented DELETE request body omits `purgeOldest?`, which the route accepts to clear a `trash_full` refusal. Add it so the documented contract matches the code.
3. TRIVIAL, src/tests/unit/coding-project-delete.test.ts — the cross-root fixture contains only fresh entries, so the prune selects nothing and the per-root deletion path is never exercised. Make the fixture actually reach it.

Then, before you push: re-read the four defects an independent audit reproduced earlier (active-run race with data loss, the false thirty-day retention promise, cleanup damaging a same-named project, unsaved-work detection missing stashes and unpushed commits) and confirm in your report which commit fixed each, with the test that covers it. If any is still open, say so plainly rather than implying it is done.

Finish: npx tsc --noEmit clean, eslint clean on touched files, focused vitest green, push, reply on every thread, then poll `gh pr checks 851` and the review state every 3 minutes until all checks are green and no thread is unresolved. An older CHANGES_REQUESTED review whose threads are all resolved is stale — dismiss it yourself with `gh api -X PUT repos/ID-Robots/clawbox/pulls/851/reviews/<id>/dismissals` and then merge with `gh pr merge 851 --squash --delete-branch`. Confirm it says MERGED.

Run `run-vjpjlb9b` · commit `e488aa83`
Reviewed by run `run-ss7nbdlk` (automatic review pass).

**Summary**
Done — PR 851 is **MERGED** (squash commit `01b8cd32`, branch deleted).

## What changed — commit `e488aa83` on `clawbox/run-erwwu00y`

**MAJOR, `src/lib/coding-project-delete.ts`** — `physicalTrashDirs` now canonicalizes the *root*, not the trash path. New `physicalTrashDir(root)` resolves the trash itself when it exists (so a trash that is itself a link still resolves) and otherwise `realpath(root)` + `TRASH_DIR_NAME`; a root exists by definition, so the canonical answer is available before the trash is. `realpathOrNull` swallows only `ENOENT` and rethrows the rest, as asked.

Propagating the other errors had a consequence I handled rather than left implicit: `trashPurgedByOneMore` turns an unreadable shelf into a `trash_failed` refusal instead of an empty one — that count is what `trash_full` refuses on, so swallowing it would purge someone else's recoverable project without asking. `pruneProjectTrash` keeps its documented never-throws contract.

**MINOR, `CLAUDE.md:60`** — the documented body is now `DELETE { folder, kind?, confirm, force?, purgeOldest? }`, with a clause naming what each flag clears.

**TRIVIAL, `src/tests/unit/coding-project-delete.test.ts`** — the cross-root fixture now gets an *expired* entry in the code-projects trash and prunes again (`removed`/`expired` are `[stale]`, `early` empty, and the folder is gone from `codeTrash`), so `rm(entry.dir, entry.name)` really runs against a non-owner root; plus the existence assertion on the owner-root case, since `rm` with `force` succeeds on a missing path.

Three new tests prove the MAJOR fix, including the symlinked root you asked for. **I verified all three fail against the pre-fix library** (reverted `coding-project-delete.ts` alone, re-ran, restored from a saved patch): the race test reports `{count: 10, early: [oldest]}` where the fixture holds 5.

## Verification
`npx tsc --noEmit` clean · eslint clean on both touched files · 97 tests green across the three delete suites · all 10 CI check runs `success` on the merged head. Output saved to `vitest-delete-suites.txt` and `pr851-verification.md` in the evidence folder.

Replies posted on all three threads; all 10 threads are resolved. **No review dismissal was needed** — CodeRabbit's own re-review moved the decision to `APPROVED`, superseding the stale `CHANGES_REQUESTED`, so I did not force one.

## The four audited defects — all closed on the merged head

| Defect | Fixed by | Test |
|---|---|---|
| Active-run race with data loss | `3879af83` added the delete-side claim (`coding-project-removal-lock.ts`, `isProjectBeingRemoved` in `coding-agent.ts`); **`cb01915c` closed it** by adding the run-side claim (`beginRunStart`/`runStartingIn` — `git log -S` confirms both first appear there) plus the second `assertNobodyWorkingIn` immediately before the move, and by deleting the copy-then-remove fallback so the move is one atomic `rename` | `projects-delete.test.ts`: "refuses while a run is STARTING in the project, before its record exists" (:234), "is refused, and the folder is NOT moved out from under it" (:416), "cannot start at all: the run lifecycle refuses a folder being removed" (:438), "releases the claim however the removal ends" (:461) |
| False thirty-day retention promise | `3879af83` (`trashPurgedByOneMore`, the `expired`/`early` split, both bounds on preview and outcome, the `trash_full` refusal); made true again across roots by `5af2b0eb` (dedupe) | `coding-project-delete-promise.test.ts`: "'⟨locale⟩' states BOTH bounds wherever it states one" ×10 locales, "never promises the period unconditionally in English"; route test "reports an EXPIRED prune apart from an early one" (:618), "refuses to purge somebody else's recoverable project without an explicit yes" (:631) |
| Cleanup damaging a same-named project | `3879af83` (`otherProjectWithSameScope`; link and secrets taken down only when nothing else answers to the scope, `metadataKeptFor` on the outcome) | "leaves the secrets and the Vercel link alone when another project shares the name" (:659) and its negative twin (:677) |
| Unsaved work missing stashes and unpushed commits | `3879af83` (`rev-list --count --branches --not --remotes` instead of `@{upstream}..HEAD`, and `stash list --format=%H`) | "sees unpushed commits on a branch that is NOT checked out" (:338), "sees a stash, which no branch and no push carries" (:355), "counts a NEVER-pushed project's whole history as unpushed, not as zero" (:322) |

None is still open. My own change extends the second one: an unreadable shelf can no longer be read as empty, which was the last way the `trash_full` consent could be skipped silently.

## Assumptions I made
- **I branched from `origin/clawbox/run-erwwu00y`, not from this worktree's HEAD.** The worktree branch carried four extra commits belonging to PR 852 (`clawbox/run-ajx89c12`); pushing HEAD would have swallowed that PR's work into 851. Exactly one commit went up.
- **`physicalTrashDirs` is now exported** so the dedupe invariant can be asserted directly — the harm it prevents is a double count, visible only through a race. The file already exports several internals for its tests, so this matches house style.
- **`trash_failed` is reused for an unreadable shelf** rather than adding a tenth refusal code. It shares a locale sentence with the rename failure ("could not move that folder aside, so nothing was removed"), which stays true for this case, so I did not rewrite ten catalogues; the box's own English sentence is specific. I flagged this in the thread reply.

## Summary by CodeRabbit

- **New Features**
  - Added safe project deletion with previews, confirmation steps, unsaved-work warnings, trash retention, and recovery details.
  - Added messaging when runs reference a deleted project.
  - Review loops now show round numbers and explain whether fixes resumed or started fresh.
  - Existing pull requests can be adopted and monitored by coding runs.
- **Improvements**
  - Expanded auto-merge guidance for non-main target branches.
  - Added localized project-deletion and review-loop messaging across supported languages.
- **Documentation**
  - Expanded Coding Agent lifecycle and project-management documentation.


yalexx:
Closing as a duplicate. This branch's tree is identical to the branch behind #852 (`git diff` between the two is empty), and it also re-contains the project-delete work that already merged as #851. Keeping it open would mean resolving conflicts against changes that are already on `beta`. The work is not lost — it lands via #851 (merged) and #852 (approved, awaiting CI).

PR #884, opened 2026-09-16
fix(chat): never show OpenClaw's config-conflict refusal on the first screen

yalexx:
## The bug

On a box that has just finished setup — Telegram skipped, desktop painting for the first time — the chat panel opens with a red bubble the owner did nothing to earn:

> Error: The config file changed while this command was writing (config changed since last load), so nothing was changed. Re-run the same command to pick up the new file and try again.

Chat works normally afterwards. The owner is left with an instruction to re-run a command they have never typed, about a file they do not know exists.

## Which request sends it

The chat panel normalizes the ClawBox AI alias to Flash when it first opens, **automatically**: an effect in `ChatPopup` POSTs `/setup-api/chat/model` with `automatic: true`, and that route writes `agents.defaults.model.primary` through `openclaw config set --batch-json`. Nobody clicked anything — this is the panel's own first-load bootstrap.

## Which concurrent writer causes it

The same first paint runs two other writers:

- the **ClawBox AI connect**'s tail — a config write followed by a gateway restart, and the gateway touches the config as it reloads;
- the **timezone adopter**, which mounts with the desktop and is its own `openclaw config set`.

Two `config set` processes plus a reloading gateway against OpenClaw's content-hash check. One of them loses.

## Why the existing retry did not save it

`withConfigMutationRetry` has retried this exact race since the batched writes landed — but it recognised the refusal only by `/ConfigMutationConflictError/i`. The CLI also words the same refusal for a human, and **that spelling carries no class name at all**:

```
ConfigMutationConflictError: config changed since last load        ← retried
The config file changed while this command was writing (config     ← not retried
changed since last load), so nothing was changed. Re-run the
same command to pick up the new file and try again.
```

So the retry never fired, a collision the next attempt would have settled was reported as a hard failure, and the CLI's sentence travelled up as the error message straight into the transcript.

## The fix

**At the source.** A new `isConfigMutationConflict` matches the wording as well as the class, in the same spirit as `isSessionTakeover` in `chat-error-text.ts` — the CLI's own words are the contract we actually receive, and pinning only its internals is what let a re-worded error through. Every config write in the app now retries the race it was always meant to retry.

Then two belts around the bubble itself:

- **The route** answers a conflict that outlived those retries with `409` and `code: \"config_busy\"` plus a short sentence of our own. Not one word of the CLI's refusal leaves the server.
- **The panel** re-sends the whole request once — a fresh POST is what re-reads the config, which is the only thing that makes a retry mean anything — and only a second conflict is worth a word to the owner: one short translated sentence that names neither the config file nor a command to re-run.

`error` keeps carrying a human sentence for every reader that predates the code, and the client recognises that older shape too, so a box in the middle of a rolling update cannot put the CLI's prose on screen either.

## Tests

The conflict is simulated at every layer:

- **spawn level** — a child that exits 1 with each spelling on stderr: both are retried, a batched write is retried, the attempt budget still ends in a failure, and a refusal the CLI means to be final is *not* retried.
- **route** — the conflict answers 409 with the code, the body matches nothing of the CLI's wording, no gateway restart is attempted, and every other failure reports itself exactly as before.
- **panel** — one retry and then silence when it lands; one short translated sentence, said once, when the retry loses too; the CLI's sentence never rendered, including from a server that sends no code; an unrelated failure still keeps its own words.

Copy added for all ten locales.

## Checks

`tsc --noEmit` clean · 1151 test files, 17 907 tests passing · `eslint` clean on every changed file (the repo-wide advisory errors are the pre-existing baseline, all in files this PR does not touch).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chat model switching when settings are being saved by automatically retrying once.
  * Added a clear, localized message when switching remains unavailable.
  * Configuration conflicts now return a safe retry prompt instead of exposing internal error details.
  * Preserved existing handling for unrelated configuration and permission errors.
  * Improved configuration updates to retry transient save conflicts before failing.

* **Localization**
  * Added the model-switching busy message across supported desktop languages.


yalexx:
> Extend the sanitized config-busy response to the auto-extend write path (`route.ts:1167-1181`)

Fixed. You are right, and for a reason beyond the missing retry: the 502 that catch returns says **"Re-save it in Settings to refresh the model list"**, which is the wrong remedy for this failure entirely. Nothing is wrong with the provider — another writer simply held the config for a moment — so that sentence sends the owner to repair something that is not broken. That is the same class of defect as the bubble this PR is about, one step earlier in the same handler.

The catch now answers a recognised conflict with the same sanitized `409` + `code: "config_busy"`, and keeps the 502 and its re-save remedy for every other failure — those would still be there a second later, so they are correctly not retryable.

Two tests added beside the existing auto-extend cases: the append losing the race answers 409 with the code and none of the CLI's wording, and a non-conflict failure keeps the re-save remedy and its 502.

For the record on scope: this path only runs for the OpenAI-compatible providers (openrouter, google, anthropic), so it is not the ClawBox AI route that produced the reported bubble — but it is the same defect, and leaving it would have kept one misleading remedy in the handler the PR set out to clean.

PR #890, opened 2026-09-17
fix(ai): the neutral plan line must wrap, not be cut off mid-word

yalexx:
Follow-up to #889, found by rendering the component at the settings-column width and looking at it.

#889 replaced a false claim on the ClawBox AI card — a box that had never stored a tier announced "Plan: Pro plan · €9/month" as fact — with a neutral line saying where the plan comes from. That line went into the span that already carried `truncate`.

`truncate` is right for a two-word plan name and wrong for a sentence. At the settings-column width, German rendered:

```
Der Tarif wird aus Ihrem Konto übernomm…
```

So the message that existed to stop the card stating something untrue was itself cut off mid-word, and the owner was left with no explanation at all — the same failure mode as the tunnel URL clipped mid-glyph that #889 fixed one screen over.

**The change:** `truncate` → `break-words` on that span. The collapsed button is already `min-h-[44px]` with `items-center`, so it grows to two lines and the "Change" affordance stays vertically centred beside it. English is unaffected (it fit on one line before and still does), and the plan-name case is unchanged in appearance.

**Verified by looking at it:** rendered the real component at a 420px column in all three variants. Before: the German line ended in an ellipsis at "übernomm…". After: it wraps to "Der Tarif wird aus Ihrem Konto" / "übernommen", complete, with "Ändern" centred against the two lines. The English neutral line and "Max plan · €49/month" render on one line as before.

**Test:** `clawbox-ai-plan-picker-unknown.test.tsx` now pins that the summary does not carry `truncate` and does carry `break-words`, with the German string as the reason in the comment.

Full suite green apart from two failures unrelated to this change and reproducible without it: `chat-popup-dismiss-and-geometry.test.tsx` (flaky under full-suite parallelism, passes in isolation) and `build-typecheck.test.ts` (tsc needs ~2.2 GB and this machine caps the default heap near 1.5 GB; it passes once nothing else is holding memory).


yalexx:
@coderabbitai review

PR #902, opened 2026-09-17
feat(chat): stop a spoken reply while it is playing

yalexx:
## The defect

A spoken reply reaches the speaker two ways and neither knew about the other:

- the chat's **automatic** playback — a detached `new Audio(src)` in `ChatPopup.playReply`, detached because the bubble the reply belongs to may not be painted yet;
- the bubble's **own** player, `SpokenReplyPlayer`.

`SpokenReplyPlayer` on beta names it at the sync effect: *"The chat's automatic playback is a detached `new Audio(src)` of its own and never touches this element; that is why a reply can be audibly playing while this transport still reads 'play'. Pre-existing, and named in the PR body."*

So a reply playing on its own could not be silenced from the bubble showing it — there was no Stop anywhere and the bubble read "play"; pressing a second bubble talked over the first; and sending a new prompt left the old answer speaking over the new question.

## What changed

**`src/lib/spoken-reply-playback.ts` (new)** — the document's one speaker. Every element that starts making sound claims it, which pauses and rewinds whatever held it. The clip is named by the `src` the caller gives (the bubble's own URL, not the element's resolved absolute one), so a bubble can tell the detached player is speaking ITS words.

The record carries `detached` (the element is the chat's, not one a bubble rendered) beside `automatic` (the cut was the chat's queue moving on, not a person). Two fields because they coincide only for the chat's playback — the Voice tab's sample autoplays an element a player rendered and is neither.

**`SpokenReplyPlayer`** follows that registry through `useSyncExternalStore`, so a reply playing on its own drives this transport's clock and bar. The transport then has two faces:

- the bubble's own playback → play/pause **plus Stop beside it** while playing, because pause (keep my place) and stop (I have heard enough) are different wishes and it can keep either promise;
- the chat's detached playback → **a single Stop**, in the primary dress. Deliberately not play/pause: that element's position belongs to the chat's speech queue, the bubble cannot resume it, and a button reading "Pause" that silently went back to the beginning would be the control-that-lies shape this file's own comments argue against.

Stop is silence now plus `currentTime = 0`. The text reply and its clip are untouched — the next press plays it from the first word.

**`ChatPopup`** — the automatic playback claims the speaker and releases it on ended/error/pause; `sendMessage`, `startRecording` and the conversation release all stop it. A reply still waiting its turn in the speech queue is still synthesised and put on its bubble but does **not** start talking, through an interruption count read at the moment it queues.

**`ChatApp`** (the full page, the surface a phone lands on) — its send stops the speaker too. It has no detached playback of its own, so its replies go through the shared player and get Stop from it.

One new key, `chat.audioStop`, in all ten shipped locales.

## Tests

RED first, on unmodified beta (implementation files restored from `origin/beta`, the new tests kept):

```
chat-voice-reply.test.tsx -t "Stop"
  x puts a Stop on the bubble of a reply playing on its own, and one press silences it
chat-voice-reply.test.tsx -t "when the owner"
  x stops the reply that is speaking when the owner sends a new prompt
  x stops the reply that is speaking when the owner starts a new spoken question
    -> Unable to find an element by: [data-testid="spoken-reply-stop"]
chatapp-stop-spoken-reply.test.tsx   (beta ChatApp)
  x silences the reply still speaking when the owner sends a new prompt
    -> expected "vi.fn()" to be called at least once
```

GREEN with the change:

| Command | Result |
| --- | --- |
| `npx tsc --noEmit` | clean |
| `npx eslint` on every touched file | clean (see limitations) |
| `spoken-reply-playback` + `spoken-reply-player-stop` + `chat-voice-reply` + `chat-spoken-reply-player` + `voice-output-panel` + `chat-spoken-reply` | 88 passed |
| `chatapp-stop-spoken-reply` | 1 passed |
| `translations` + `i18n-scan` + `i18n-sweep-catalogue` + `locale-sweep-de-catalogue` + `edition-translations` | 327 passed |
| all `chat*` / `voice*` / `spoken*` component suites | 667 passed, 1 pre-existing flake |

## Not verified

- **Not verified in the real UI on a device.** No ClawBox hardware was reachable from the environment this was written in, so the browser-level facts — that a real engine fires `pause` on the rewind, the 36 px touch target under a thumb, and how a second button lays out in a 340 px docked panel — rest on jsdom tests and reasoning, not on a box. That check is outstanding.
- `chat-popup-dismiss-and-geometry.test.tsx` is flaky under load (a 5 s test timeout, a different test each run). Pre-existing and untouched here: it reproduces at the same rate with `ChatPopup`/`ChatApp`/`SpokenReplyPlayer` restored from `origin/beta`.
- `ChatApp.tsx` carries one pre-existing `react-hooks/preserve-manual-memoization` eslint error, identical on `origin/beta`.
- The chat's automatic playback can be stopped but not paused-and-resumed. Resuming would mean the bubble taking over an element whose position the speech queue owns, and `playReply` treats `pause` as "heard enough" so that queue can advance. Out of scope; the transport says what it does rather than offering a pause it could not keep.


yalexx:
@coderabbitai review

PR #915, opened 2026-09-18
fix(whisper): give the speech-model fetcher the engine's library path (TASK-893)

yalexx:
## Symptom (TASK-893)

On a Jetson Orin (board 1791626120943, beta `0a175513`), Settings → Local AI → Speech to text: Install succeeds, Enable succeeds and the model-size card renders. But pressing **Download** on any size failed after about 3 s with:

> Failed: faster-whisper is not installed on this box (libctranslate2.so.4: cannot open shared object file: No such file or directory)

As a result, no speech model could be installed from Settings on the product hardware.

## Root cause

- `install-voice.sh` builds CTranslate2 (CUDA) into `~/.local/lib`. Nothing registers that directory with `ldconfig`.
- `whisper-server.service` sets `Environment=LD_LIBRARY_PATH=…/.local/lib:…`, which is why the engine itself works.
- `clawbox-setup.service` (the web server) sets no `LD_LIBRARY_PATH`. `src/app/setup-api/whisper/route.ts` spawned `scripts/fetch-whisper-model.py` with no `env`, so the fetcher inherited the web server's environment. Its `import faster_whisper` then could not load the library and it exited with code 3, which produced the message above.

## Fix

- `src/lib/stt-local.ts`: new exported `withWhisperLibraryPath(env)`. It puts the module's existing `ldLibraryPath()` in front of any inherited `LD_LIBRARY_PATH` and leaves out the inherited part when it is empty (an empty entry means the current directory to the loader). `ldLibraryPath()` is the path the Settings import probe and the chat-mic transcriber already use. It comes from `HOME`, not a hard-coded account.
- `src/app/setup-api/whisper/route.ts`: the fetcher spawn now passes `env: withWhisperLibraryPath(process.env)`. Everything else it inherited before is unchanged: `HOME` for the Hugging Face cache, `PATH`, proxy variables.

### Other spawns checked, none changed

- In this route, the engine install goes through the `voice_whisper_install` root step, and the restart and removal go through `systemctl --user`. None of them import faster-whisper in the web server's environment.
- Nearby routes: `stt-local.ts`'s probe and transcriber already pass `LD_LIBRARY_PATH` through `childEnv()`. TTS speak/sample run `clawbox-tts.sh`, which builds its own library path. TTS warm uses systemd. None of these has the same defect.

## Tests

`src/tests/routes/whisper.test.ts` has a new block, "the fetcher's environment". The `child_process.spawn` mock now records each call so the tests can check the options the route passes:

- the fetcher gets `LD_LIBRARY_PATH` with `$HOME/.local/lib` first and `/usr/local/cuda/lib64`, with no empty entries
- the directory follows `HOME` (no `/home/clawbox` when `HOME` points elsewhere)
- the rest of the server's environment is kept (`HOME`, `PATH`, `HTTPS_PROXY`)
- an inherited `LD_LIBRARY_PATH` goes after the engine's entries, and the server's own `process.env` is left unchanged

With the one-line `env:` fix removed, all four new tests fail (`expected undefined to be defined`). With it in place, they pass.

## Verification (run on the device, this branch)

- `NODE_OPTIONS=--max-old-space-size=8192 npx tsc --noEmit -p tsconfig.json`: exit 0, no errors
- `bun run test`: `Test Files  1169 passed (1169)`, `Tests  18127 passed | 1 skipped (18128)`
- `bun run lint`: exit 1, `✖ 160 problems (20 errors, 140 warnings)`. These are the same 20 errors as upstream `0a175513`, all in untouched files (`scripts/*.js`, `ChatApp.tsx:377`, `useReconnect.ts:77`, two unit tests). The changed files have no errors. The three `_a` warnings in `whisper.test.ts` are on existing lines 40–43.

This has not been re-run on the Jetson. The next step there is to press Download on a size and confirm it completes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Whisper model setup so locally installed CTranslate2 and CUDA libraries are found reliably.
  - Preserved existing environment settings, including library paths, home directory, proxy configuration, and executable paths.
  - Prevented empty library-path entries from being passed to the model-fetching process.
  - Added coverage to verify Whisper model fetching uses the expected environment and command configuration.


yalexx:
Superseded by #916 (same three files, merged into beta 2026-09-18). This branch now conflicts with beta; closing.

